PR #21, opened 2022-10-05
Added support Windows (build Visual Studio 2022)

CHERTS:
Hi,

I've added support  Windows (build in Visual Studio 2022), everything builds and works just fine.


ggerganov (changes requested):
Hi @CHERTS - thank you very much for the contribution!

I want to keep the third-party dependencies at minimum, so before merging this the `SDL`, `pthreads4w` and `stdatomic` dependencies have to be removed.

- The `SDL` library is only needed for the `stream` example and the user should provide it themselves if they want to use it
- On Windows, there is a [native thread API](https://learn.microsoft.com/en-us/windows/win32/procthread/creating-threads?redirectedfrom=MSDN) that should be used instead of `pthreads`
- The `stdatomic` is [part of the C11 standard](https://en.cppreference.com/w/c/atomic) so it should be directly supported by the compiler - not sure why you had to add third-party implementation

CHERTS:
I'm sorry, but I don't see the point in removing the SDL header files from the VisualStudio project, without them it won't be able to compile
I also see no point in writing my own implementation of pthreads for windows when pthreads4w is available.
stdatomic - well, sorry, Microsoft didn't implement it for C, it's only for C++, I had to take the implementation from ffmpeg

Issue #15, opened 2022-10-03
Request to support aarch64


trholding:
Make errors out on a aarch64 server

```
make base.en
#gcc -pthread -O3 -c ggml.c
gcc -pthread -O3 -mcpu=cortex-a72 -mfloat-abi=hard -mfpu=neon-fp-armv8 -mfp16-format=ieee -mno-unaligned-access -c ggml.c
gcc: error: unrecognized command-line option ‘-mfloat-abi=hard’
gcc: error: unrecognized command-line option ‘-mfpu=neon-fp-armv8’
gcc: error: unrecognized command-line option ‘-mfp16-format=ieee’
gcc: error: unrecognized command-line option ‘-mno-unaligned-access’
make: *** [Makefile:7: ggml.o] Error 1

```
Perhaps this is enough as C flags? : -Ofast -g -mfpu=neon

ggerganov:
@trholding Yes, I think this will work. Obviously, the current Makefile is not very cross-platform - I mainly focus to be able to compile on M1 and Intel processors. With time, it will be improved to support various other platforms.

trholding:
@ggerganov Works awesome on aarch64! 

```
time make base.en 
bash ./download-ggml-model.sh base.en
Downloading ggml model base.en ...
Model base.en already exists. Skipping download.

===============================================
Running base.en on all samples in ./samples ...
===============================================

----------------------------------------------
[+] Running base.en on samples/jfk.wav ... (run 'ffplay samples/jfk.wav' to listen)
----------------------------------------------

whisper_model_load: loading model from 'models/ggml-base.en.bin'
whisper_model_load: n_vocab       = 51864
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 512
whisper_model_load: n_audio_head  = 8
whisper_model_load: n_audio_layer = 6
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 512
whisper_model_load: n_text_head   = 8
whisper_model_load: n_text_layer  = 6
whisper_model_load: n_mels        = 80
whisper_model_load: f16           = 1
whisper_model_load: type          = 2
whisper_model_load: mem_required  = 377.00 MB
whisper_model_load: adding 1607 extra tokens
whisper_model_load: ggml ctx size = 163.43 MB
whisper_model_load: memory size =    22.83 MB 
whisper_model_load: model size  =   140.54 MB

[...]

Issue #8, opened 2022-10-01
Correct parameter for cross compile for ARM Android ?


ekorudi:
What is correct parameter for cross compile for ARM Android ? I'm using Intel Ubuntu , android-ndk-r25b

```$ ../arm-compiler/bin/clang -pthread -O3 -mcpu=cortex-a72 -mfloat-abi=hard -mfpu=neon-fp-armv8  -mno-unaligned-access  -c ggml.c

ggml.c:232:16: warning: implicit declaration of function 'vfmaq_f32' is invalid in C99 [-Wimplicit-function-declaration]
        sum0 = vfmaq_f32(sum0, x0, y0);
               ^
ggml.c:232:14: error: assigning to 'float32x4_t' (vector of 4 'float32_t' values) from **incompatible type** 'int'
        sum0 = vfmaq_f32(sum0, x0, y0);
             ^ ~~~~~~~~~~~~~~~~~~~~~~~

./ggml.c:331:14: error: assigning to 'float16x8_t' (vector of 8 'float16_t' values) from **incompatible type** 'int'
        sum0 = vfmaq_f16(sum0, x0, y0);
             ^ ~~~~~~~~~~~~~~~~~~~~~~~             
```

ekorudi:
Hi, I found a way to compile with following Makefile using raspi-4 brach

`
gcc="../arm-compiler/bin/aarch64-linux-android28-clang"
gpp="../arm-compiler/bin/aarch64-linux-android28-clang++"

main: ggml.o main.o
	$(gpp) -pthread -o main ggml.o main.o -static-libstdc++
	adb push main /data/local/tmp/main
    echo "run with adb shell"

ggml.o: ggml.c ggml.h
	$(gcc) -pthread -O3 -c ggml.c -mcpu=cortex-a75 -mfpu=neon-fp-armv8 

main.o: main.cpp ggml.h
	$(gpp) -pthread -O3 -std=c++11 -c main.cpp -static-libstdc++

# clean up the directory
clean:
	rm -f *.o main
`

Top from other terminal show below :

`
Tasks: 609 total,   2 running, 607 sleeping,   0 stopped,   0 zombie
  Mem:      5.5G total,      5.1G used,      379M free,      7.0M buffers
 Swap:      3.0G total,      548M used,      2.5G free,      2.0G cached
800%cpu 395%user   0%nice   4%sys 394%idle   0%iow   0%irq   8%sirq   0%host
[...]

PR #23, opened 2022-10-05
Support Raspberry Pi 4 + Android

ggerganov:
ref #7
ref #8 
ref #15 

Should now build correctly on different platforms.
On Arm platforms without `__ARM_FEATURE_FP16_VECTOR_ARITHMETIC` we convert to 32-bit floats. There might be a more efficient way, but this is good for now.


nyadla-sys:
@ggerganov Used the following command to run a stream on a Raspberry Pi 4, but it didn't work as live streaming decoding.
Perhaps further improvements are needed.
./stream -m models/ggml-tiny.en.bin

PR #24, opened 2022-10-05
ADD VAD to save inference cycles

nyadla-sys:
Simple voice activity detection was included.

Thank you to: https://github.com/cirosilvano/easyvad.git

### DIFF
--- a/stream.cpp
+++ b/stream.cpp
@@ -17,6 +17,34 @@
 #include <string>
 #include <thread>
 #include <vector>
+#include <math.h>
+
+// vad settings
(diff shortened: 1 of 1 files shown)


randomeizer:
Looks like the original code is GPL, which is not compatible with the MIT license.

ggerganov:
@nyadla-sys Thanks for the contribution!

Unfortunately @randomeizer is correct - adding GPL code to this project is not possible because this will make the project also GPL and I don't want that.

You can try to re-implement this idea in an original way and re-submit it again.
If you do so, it would be nice to add a command line option to toggle it on and off.

Issue #5, opened 2022-09-30
Windows build


ggerganov:
Would be nice if someone can help and provide build instructions for Windows.

I think the only thing that might need an update is the [pthread dependency in ggml.c](https://github.com/ggerganov/whisper.cpp/blob/310f4883d1d4bbba56a69f560870566ed1e4ed76/ggml.c#L12).
The rest of the code should build successfully.

Probably a `.bat` script to download the models would also be nice since no Bash on Windows.

kevin01881:
Would love this!! I want to run whisper on my ancient i7-5500U and right now with OpenAI's version on Windows 10, I can only transcribe 100 minutes in 24 hours. ☹️

trholding:
Here is my recommendation, for x86 windows and multiple targets at once in a single binary format with zero to minimal changes: https://github.com/jart/cosmopolitan 

For a demo of what is possible with this: https://redbean.dev/

Info here:

https://justine.lol/cosmopolitan/ 
https://justine.lol/ape.html

I have used it for small internal projects. And works fine. Basically this is how it works:

```Here's how simple it is to get started:

gcc -g -O -static -fno-pie -no-pie -mno-red-zone -nostdlib -nostdinc -o hello.com hello.c \
  -Wl,--oformat=binary -Wl,--gc-sections -Wl,-z,max-page-size=0x1000 -fuse-ld=bfd -gdwarf-4 \
  -Wl,-T,[ape.lds](https://justine.lol/cosmopolitan/ape.lds) -include [cosmopolitan.h](https://justine.lol/cosmopolitan/cosmopolitan.h) [crt.o](https://justine.lol/cosmopolitan/crt.o) [ape.o](https://justine.lol/cosmopolitan/ape.o) [cosmopolitan.a](https://justine.lol/cosmopolitan/cosmopolitan.a)
In the above one-liner, we've basically reconfigured the stock compiler on Linux so it outputs binaries that'll run on MacOS, Windows, FreeBSD, OpenBSD, and NetBSD too. They also boot from the BIOS. Please note this is intended for people who don't care about desktop GUIs, and just want stdio and sockets without devops toil.```

Doesn't work if you use libc functions that are not implemented, most are. If you get issues, this is your safe way to mitigate missing headers etc: https://github.com/fabriziobertocci/cosmo-include

Here are more resources to get you started:

https://github.com/shmup/awesome-cosmopolitan

Wish you good fun! I'm sure you'll love it. 

[...]

PR #31, opened 2022-10-09
Add MinGW support

lkwq007:
Hi, I just tried building with `mingw-w64` and it works quite well. https://github.com/ggerganov/whisper.cpp/issues/5

### DIFF
--- a/ggml.c
+++ b/ggml.c
@@ -1,6 +1,11 @@
 #include "ggml.h"
 
+#if defined(_MSC_VER) || defined(__MINGW32__)
+#include <malloc.h> // using malloc.h with MSC/MINGW
+#else
 #include <alloca.h>
+#endif
+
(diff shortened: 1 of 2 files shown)


whisper.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include <algorithm>
 #include <cassert>
+#define _USE_MATH_DEFINES
```
lkwq007, line 7: For `error: 'M_PI' was not declared in this scope`

ggml.c
```diff
@@ -1,6 +1,11 @@
 #include "ggml.h"
 
+#if defined(_MSC_VER) || defined(__MINGW32__)
```
lkwq007, line 3: `ggml.c:3:10: fatal error: alloca.h: No such file or directory`

PR #36, opened 2022-10-10
Fix SRT timestamp format from mm:ss.sss to hh:mm:ss.sss

Topping1:
Fixed SRT timestamp format from mm:ss.sss to hh:mm:ss.sss. See https://docs.fileformat.com/video/srt/ for reference.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -14,13 +14,16 @@
 //  500 -> 00:05.000
 // 6000 -> 01:00.000
 std::string to_timestamp(int64_t t) {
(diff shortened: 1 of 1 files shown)


ggerganov (approved):

patarapolw:
> Fixed SRT timestamp format from mm:ss.sss to hh:mm:ss.sss. See https://docs.fileformat.com/video/srt/ for reference.

hh:mm:ss,sss that is , (comma) rather than . (dot) as well.

Issue #25, opened 2022-10-06
Token decoding issue - some characters are missing


yujinqiu:
```
./main -m models/ggml-medium.bin -l zh -f ~/Movies/samplecn16k.wav
```
output
```
[00:00.000 --> 00:16.000]  元����,其实就是����世界,而且要用����世界这个��来定��元����的话,要比元����本身更加����。到这里就出现问题了。那它为什么不叫����世界呢?最��单的原因就是,����世界这个说法大家已经听��了,而元������得更为新��,又包��成为了一个新的概念。
[00:16.000 --> 00:44.000]  现在的元����技��,����没有我们想象中那么先进。按照目前世界第一元����公司,Roblox公司对于元����的定��来看,它起��要具��8个要素,分别是身份、社交、成进、����、多元、��地、经��、文明。身份就是一个����身份,��现实中的角色无关,这个比��好理解。社交也就是社交系��。成进就是感知����的升��,要做到和现实世界的体��完全相同。����就��������,不会有卡��,多元就多元化,
```

with OpenAI `whisper` cli
```
whisper --language zh ~/Movies/samplecn16k.wav
```

```
[00:00.000 --> 00:01.760] 元宇宙其实就虚拟世界
[00:01.760 --> 00:04.400] 而且要用虚拟世界这个词来定义元宇宙的话
[00:04.400 --> 00:06.400] 要比元宇宙本身更加准确
[00:06.400 --> 00:07.680] 但这里就出现问题了
[00:07.680 --> 00:09.360] 那它为什么不叫虚拟世界呢?
[00:09.360 --> 00:10.720] 最简单的原因就是
[00:10.720 --> 00:12.880] 虚拟世界这个说法大家已经听腻了
[00:12.880 --> 00:14.320] 而元宇宙显得更为吸引
[00:14.320 --> 00:16.200] 又包装成为了一个新的概念
[00:16.200 --> 00:17.440] 现在的元宇宙技术
[00:17.440 --> 00:19.160] 原有没有我们想象中那么先进
[00:19.160 --> 00:21.320] 按照目前世界第一元宇宙公司
[00:21.320 --> 00:23.480] 罗布洛克斯公司对于元宇宙的定义来看
[00:23.480 --> 00:25.080] 它起码要具备8个要素
[00:25.080 --> 00:30.680] 分别是身份、社交、成敬、延迟、多元、随地、经济、文明
[00:30.680 --> 00:32.280] 身份就是一个虚拟身份
[...]

PR #58, opened 2022-10-17
fix decode missing token issue

r0y6a3n0:
@ggerganov , thanks for the great project!
I fixed the #25 issue.  
Just fix one line in convert-pt-to-ggml.py. 
I think the root cause is :
`bytearray([byte_decoder[c] for c in key])`
This have already convert token to utf-8 so you don't need encode and decode again.

before change code:
截圖 2022-10-17 下午8 55 30
after change code:
截圖 2022-10-17 下午8 53 46


ggerganov (approved):
Awesome! Thanks so much for the help!

ggerganov:
I will merge this and upload the new model files later tonight

PR #121, opened 2022-11-02
Cross compilation

jafri:
Makes the makefile more cross compilation compatible

Was able to compile to all 4 major platforms on my M1 Mac 

Using zig 0.9.1 on M1 Mac (upgrading is blocked by https://github.com/ziglang/zig/issues/8531)

MacOs x86_64 (no Accelerate atm)
```
UNAME_S="Darwin" \
UNAME_P="arm" \
UNAME_M="x86_64" \
CC="zig cc -target x86_64-macos" \
CXX="zig c++ -target x86_64-macos" \
arch -x86_64 make main
```

MacOs Aarch64 (Accelerate)
```
UNAME_S="Darwin" \
UNAME_P="arm" \
UNAME_M="arm64" \
CC="zig cc" \
CXX="zig c++" \
make main
```

WIndows
```
UNAME_S="Windows" \
UNAME_P="pc" \
UNAME_M="x86_64" \
CC="zig cc -target x86_64-windows" \
CXX="zig c++ -target x86_64-windows" \
AR="zig ar" \
RANLIB="zig ranlib" \
make main
```

Linux
```
UNAME_S="Linux" \
UNAME_P="unknown" \
UNAME_M="x86_64" \
CC="zig cc -target x86_64-linux -D_POSIX_C_SOURCE=199309L" \
CXX="zig c++ -target x86_64-linux" \
AR="zig ar" \
RANLIB="zig ranlib" \
make main
```


Makefile
```diff
@@ -1,6 +1,21 @@
...
+else
 UNAME_P := $(shell uname -p)
+endif
+
+ifdef UNAME_M
+UNAME_M := $(UNAME_M)
+else
 UNAME_M := $(shell uname -m)
+endif
+
```
ggerganov, line 18: Use `indef UNAME_X ..` instead of `ifdef .. else .. `

Makefile
```diff
@@ -78,13 +93,13 @@ main: examples/main/main.cpp ggml.o whisper.o
 	./main -h
 
 ggml.o: ggml.c ggml.h
-	$(CC)  $(CFLAGS)   -c ggml.c
+	$(CC)  $(CFLAGS) -c ggml.c -o ggml.o
```
ggerganov, line 96:
```suggestion
	$(CC)  $(CFLAGS)   -c ggml.c -o ggml.o
```

I really want to keep this whitespace 😄 
It makes the commands aligned vertically

ggerganov (changes requested):

jafri:
Fixed @ggerganov, one consideration here is that we should add OPENBLAS support to makefile too

ggerganov:
Yes, you can give it a shot if interested. Maybe check if some env variable is set and if yes, link to blas. Not sure what is the best way

jafri:
Doesn't BLAS work on all x86 platforms?

ggerganov:
AFAIK it does but it does not come out of the box. You have to install a third party library and link it.

jafri:
Can't you statically link libopenblas.a? https://github.com/xianyi/OpenBLAS/blob/develop/USAGE.md#linking-with-the-library

PR #128, opened 2022-11-08
Fix the Windows pthread_create shim (Assertion failed: rc == 0)

boolemancer:
The current implementation doesn't actually set the out parameter, and it returns 0 on failure instead of on success.

This should fix people hitting the `Assertion failed: rc == 0` failure.

I also made a change to fix the compiler flags to enable AVX2 in the MSVC build.

Should address at least some of the problems mentioned in #5.


ggerganov (approved):
Thanks for the fixes!

Issue #66, opened 2022-10-18
Not working on MacOS (ARM)


undefdev:
Hi, I've been trying to get this to work a few times, but it always fails with an illegal hardware instruction error.

E.g. for `./main -m models/ggml-small.bin -f samples/jfk.wav` I get the following output:
```
whisper_model_load: loading model from 'models/ggml-small.bin'
whisper_model_load: n_vocab       = 51865
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 768
whisper_model_load: n_audio_head  = 12
whisper_model_load: n_audio_layer = 12
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 768
whisper_model_load: n_text_head   = 12
whisper_model_load: n_text_layer  = 12
whisper_model_load: n_mels        = 80
whisper_model_load: f16           = 1
whisper_model_load: type          = 3
whisper_model_load: mem_required  = 1048.00 MB
whisper_model_load: adding 1608 extra tokens
whisper_model_load: ggml ctx size = 533.05 MB
fish: Job 1, './main -m models/ggml-small.b...' terminated by signal SIGILL (Illegal instruction)
```

I've tried other models as well, but the result is always the same.

ggerganov:
That is interesting. What device are you using?

What is the output of:

```bash
make clean
make
```

undefdev:
I'm using a MacBook Pro with M1 Max.

output of `make clean`:
```
rm -f *.o main libwhisper.a
```
output of `make`:
```
cc  -O3 -std=c11   -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -pthread -mavx -mavx2 -mfma -mf16c   -c ggml.c
c++ -O3 -std=c++11 -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -pthread -c whisper.cpp
c++ -O3 -std=c++11 -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -pthread main.cpp whisper.o ggml.o -o main 
./main -h

usage: ./main [options] file0.wav file1.wav ...

options:
  -h,       --help           show this help message and exit
  -s SEED,  --seed SEED      RNG seed (default: -1)
  -t N,     --threads N      number of threads to use during computation (default: 4)
  -o N,     --offset N       offset in milliseconds (default: 0)
  -v,       --verbose        verbose output
            --translate      translate from source language to english
[...]

PR #131, opened 2022-11-09
Check for AVX and AVX2 on Darwin

chidiwilliams:
Related to https://github.com/ggerganov/whisper.cpp/issues/66, this checks for the presence of AVX1.0 and AVX2 before adding the -mavx and -mavx2 flags


ggerganov (approved):
Much appreciated!

PR #135, opened 2022-11-09
Support for stdin pipe stream

### LINKED ISSUE #65: [Feature] recognize data coming via pipe stream
Hi,

it would be great to have a simple app that takes data from pipe and runs recognition on it ... similar to [stream.cpp](https://github.com/ggerganov/whisper.cpp/blob/master/stream.cpp), but instead taking data from audio device, taking it from pipe ...

Could also be an addition to the main-example, so that you can use it like this:

`cat samples/jfk.wav | ./main -m models/ggml-medium.bin -f -`

[Here](https://github.com/alphacep/vosk-api/blob/master/python/example/test_ffmpeg.py) something similar is done with vosk and python. (ffmpeg-pre-processing could be something people can do on their own before filling the pipe and not part of the app ...)

Kind regards,
abelbabel

ocordeiro:
Fix #65

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -454,9 +454,30 @@ int main(int argc, char ** argv) {
         std::vector<float> pcmf32;
         {
             drwav wav;
-            if (!drwav_init_file(&wav, fname_inp.c_str(), NULL)) {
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
Awesome!

Issue #137, opened 2022-11-10
Faster streaming support


ameenba:
Have you tried building the spectrogram and encoder output in smaller chunks and appending? I think the spectrogram should generate fairly easily with minimal noise depending on the size of the chunk, and the encoder output can also be appended with sufficiently large chunks.

So the encoder instead takes in Bx80xN as its input and outputs Bx(N/2)x(embedding size), if you wanted to send in 1s of audio into the tiny.en model, for example, 1x80x100->1x50x384. This should result in much faster processing for short clips (when the audio clip is <30s), and allows real time streaming without much wasted computation (like having to calculate a full x1500 encoding for each chunk of audio).

Some noise may be introduced at various size of chunks (spectrogram chunk size can be independent from encoder chunk size), and some overlap of spectrogram input/encoder output may help further to reduce that noise. This allows us for better scheduled deployment where the decoder, encoder, and spectrogram can run on different threads at the same time to produce our transcription.

Choosing when to decode will be another challenge as you don't want to decode if a full word is not complete in the encoding, but there are definitely solutions around that as well.

ggerganov:
Oh my god! I just tested that and it seems to work o.O
I reduced the audio context by half and the performance doubled. `jfk.wav` transcribed correctly!

Very interesting...
I have to go now, but I think this is a big breakthrough. Need to double check.

ggerganov:
Here is diff if someone wants to play with it:

```diff
git diff
diff --git a/whisper.cpp b/whisper.cpp
index 7078863..df47bff 100644
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -1053,6 +1053,7 @@ static bool whisper_model_load(const std::string & fname, whisper_context & wctx
             return false;
         }
     }
+    model.e_pe->ne[1] /= 2;
 
     fin.close();
 
@@ -1076,7 +1077,7 @@ static bool whisper_encode(
     const auto & mel_inp = wctx.mel;
     const auto & hparams = model.hparams;
 
-    const int n_ctx   = hparams.n_audio_ctx;
+    const int n_ctx   = hparams.n_audio_ctx/2;
     const int n_state = hparams.n_audio_state;
     const int n_head  = hparams.n_audio_head;
     const int n_layer = hparams.n_audio_layer;
@@ -1474,7 +1475,7 @@ static bool whisper_decode(
     const int n_layer = hparams.n_text_layer;
 
     const int N = n_tokens;
[...]

PR #141, opened 2022-11-11
Real-time streaming

ggerganov:
[WIP in progress]

With the idea in #137 it is possible to reduce the time in the encoder multiple times.
This is beneficial for the `stream` example, because it already processes the audio in short chunks.
The decoding quality seems to drop, but I think not significantly.

With the current parameters, I am able to run the following commands in real-time on MacBook M1 Pro:

```
# real-time transcription with a step of 1.5 seconds using "medium.en"
./stream -m ./models/ggml-medium.en.bin -t 8 --step 1500 --length 7500 -ac 512

# real-time translation with a step of 2.5 seconds using "large"
./stream -m ./models/ggml-large.bin -t 8 --step 2500 --length 7500 --language bg --translate  -ac 512
```

This was not possible before.

Next thing to try is to run the `tiny` model in streaming mode in the browser using WASM with a step of 1 or 2 seconds.
I think there is some chance it could actually work.

### DIFF
--- a/examples/stream/stream.cpp
+++ b/examples/stream/stream.cpp
@@ -40,6 +40,7 @@ struct whisper_params {
     int32_t step_ms    = 3000;
     int32_t length_ms  = 10000;
     int32_t capture_id = -1;
+    int32_t audio_ctx  = 0;
 
     bool speed_up             = false;
     bool verbose              = false;
@@ -69,6 +70,8 @@ bool whisper_params_parse(int argc, char ** argv, whisper_params & params) {
             params.length_ms = std::stoi(argv[++i]);
         } else if (arg == "-c" || arg == "--capture") {
             params.capture_id = std::stoi(argv[++i]);
+        } else if (arg == "-ac" || arg == "--audio_ctx") {
+            params.audio_ctx = std::stoi(argv[++i]);
         } else if (arg == "-su" || arg == "--speed-up") {
             params.speed_up = true;
         } else if (arg == "-v" || arg == "--verbose") {
@@ -116,6 +119,7 @@ void whisper_print_usage(int argc, char ** argv, const whisper_params & params) 
     fprintf(stderr, "            --step N         audio step size in milliseconds (default: %d)\n", params.step_ms);
     fprintf(stderr, "            --length N       audio length in milliseconds (default: %d)\n", params.length_ms);
     fprintf(stderr, "  -c ID,    --capture ID     capture device ID (default: -1)\n");
+    fprintf(stderr, "  -ac N,    --audio_ctx N    audio context size (default: %d, 0 - all)\n", params.audio_ctx);
     fprintf(stderr, "  -su,      --speed-up       speed up audio by factor of 2 (faster processing, reduced accuracy, default: %s)\n", params.speed_up ? "true" : "false");
     fprintf(stderr, "  -v,       --verbose        verbose output\n");
     fprintf(stderr, "            --translate      translate from source language to english\n");
@@ -322,7 +326,6 @@ int main(int argc, char ** argv) {
         {
             whisper_full_params wparams = whisper_full_default_params(WHISPER_SAMPLING_GREEDY);
 
-            wparams.max_tokens           = 32;
             wparams.print_progress       = false;
             wparams.print_special_tokens = params.print_special_tokens;
             wparams.print_realtime       = false;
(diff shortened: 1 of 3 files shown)


meakbiyik:
The performance gain here is absurd. Is there anything I can pitch in here to finalize this PR @ggerganov? I could not exactly get the issue with stitch encoder outputs together 😅

ggerganov:
The "stitching" is basically instead of running 10 seconds of audio through the encoder at one pass, run for example 5 x 2 second chunks and combine the results in the cross-attention layer to get effectively what we would have gotten with 10 seconds directly. This would allow to process audio more often and be more real-time.

The PR missies an option to enable/disable the encoder truncation - I currently hardcoded the values. It's not difficult to finalise, but I want to see how I will use it in the streaming examples - probably will get a better idea for the API.

ggerganov:
@meakbiyik 
This is now on master.
Simply add `-ac 512` to the `./stream` arguments and you will enable the 3x faster Encoder

meakbiyik:
Wow, this is great - thanks a lot @ggerganov!

A quick follow-up question: would you recommend 2x speedup or reducing audio context size? Or can I mix them up, what was your experience? I do not quite understand why reducing the audio context should also reduce transcription accuracy, so I cannot be sure 😅 

Also, interestingly I have noted that lowering step size improves much better transcription, so much so that using low step size + base model is better than 2x step size + small model. Is there anything going on behind the scenes that can explain this phenomenon? Does the option `-kc / keep context` play any role here?

ggerganov:
The 2x speed-up does not seem very useful yet in my experience, so I don't recommend using it.
The smaller audio context intuitively is worse compared to the full context because you are analysing less data - less data means worse results.

The step size observation is strange - as long as your hardware is capable to process the data in real time, then the bigger model should be always better, regardless of step size. Regarding the `-kc` flag - I don't use it for `stream` because errors occur more often when doing real-time stream and the `-kc` flag can actually help propagate the errors in the future transcription.

meakbiyik:
interesting, but why is there less data, particularly if the `--length` parameter is set less than the context? What I assumed was that `--length` amount of data is used (if available), and the rest is padded with zeros, therefore if we reduce the audio_context so that `--length` fits there snugly, there should be no issues. I feel like I totally misunderstood some of these parameters 😅 

On step size, I observed that the transcription is "refined" every time the model reruns on the data it has already seen, and more refinements are better, which makes sense if the model has access to the current context of size `--length`.

`-kc` part makes sense. I actually plan the create a PR to inject arbitrary contexts as you recommended in some previous PR, but let's see what happens 😄

ggerganov:
Yeah, actually you have a good point - for a fixed `--length`, if the context is bigger, then it shouldn't affect the quality. For example `-ac 512` corresponds to a little more than `10s` context, so for `--length 10000` or less you should be getting the same quality. Your understanding is correct.

> On step size, I observed that the transcription is "refined" every time the model reruns on the data it has already seen, and more refinements are better, which makes sense if the model has access to the current context of size --length.

Yes, correct. For example, if you have `--length` of `10s`, then regardless if the step is `1s`, `2s`, `3s`, etc.. the final pass when it processes the full `10s` chunk will give the same result. Actually, I now realise that you must not use `-kc` if the `--step` is smaller than `--length` because it will be using the "partial" transcription as context for the next step and it will definitely get messy.

The `-kc` option has to be reworked as you suggest to be able to provide the context from the previous `--length` pass for each step of the current `--length` pass. Feel free to give it a shot and don't hesitate to ask if you have any questions.

meakbiyik:
Perfect, thanks a lot, all of this makes full sense! Will try to do that -kc thing quite soon.

Buuut I got one final follow-up just to understand it better: what happens if length>audio_context? Does the model it trim from then end? Or is there a downsampling going on?

ggerganov:
Currently, it will trim from the end:

https://github.com/ggerganov/whisper.cpp/blob/f2df9bd7689475e73da6480212c1a0e6aa348979/whisper.cpp#L1103-L1104

meakbiyik:
A-ha, lovely. Thanks a lot again!

xyx361100238:
According to #137 , I set -ac = 750, the result have lots of noise word “[buzzer] / [static] / [AUDIO OUT]”, how can I remove it?
BTW,it's works well use src set  audio_ctx= 0.
image

ggerganov:
Currently, the only way is to manually replace these strings yourself (for example, using regex).
Btw, `-ac 768` is better than `-ac 750` - you want the number to be multiple of 64 for better performance.

xyx361100238:
Yes Yes！Much better set  -ac 768 ：
image
add i will replace strings too. Thanks again！

PR #150, opened 2022-11-16
fix: free ggml_context

### LINKED ISSUE #149: ggml_context is not freed during whisper_model_load
This results in 2 ggml contexts not being freed even after calling `whisper_free`. As a result, if you call whisper_init repeatedly, eventually, you run out of ggml contexts. 

It's quite a simple fix. Happy to open a PR

greeshmay:
Fixes #149 

These ggml_contexts are not freed during whisper_free. As a result, the number of contexts can keep increasing if you initialize whisper consecutively. Ultimately they reach the limit of 64, and thus cannot create any new contexts

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -930,7 +930,6 @@ static bool whisper_model_load(const std::string & fname, whisper_context & wctx
                 model.tensors["decoder.blocks." + std::to_string(i) + ".cross_attn.out.bias"]   = layer.cross_attn_ln_1_b;
             }
         }
-        ggml_free(ctx);
     }
 
     // create the ggml memory context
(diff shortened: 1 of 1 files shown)


ggerganov:
Thanks for spotting this bug!
The 2 `ggml_free(ctx)` calls that you added actually have to be moved to `whisper_free()`.

The reason is that we don't want to be releasing them during initialisation.
I can see it can work this way since the memory buffers are allocated separately, but generally it will be undefined behaviour using the proposed change.

greeshmay:
I did try to free in whisper_free, but it was causing a crash for me, which was unexpected; and I couldn't figure out a way to get past the crash

ggerganov:
@greeshmay Strange - just to double-check, does a28ec7c crash for you?

greeshmay:
Weird, no. I guess I did something wrong when testing it with whisper_free. This works for me

Issue #90, opened 2022-10-25
[Feature Request] Support openai/whisper "--initial_prompt" flag


ninjalu:
I would love to add initial prompts to increase the accuracy of the transcription based on this:
https://github.com/openai/whisper/discussions/66
How do I go about implement it here?

Many thanks!

ggerganov:
The `prompt_past` member of the context contains the tokens that have been decoded so far during the transcription:

https://github.com/ggerganov/whisper.cpp/blob/a6c786d5dccf5f54a92da890f70715b4c9831172/whisper.cpp#L2388-L2392

Currently, it is empty when we start processing an audio and grows up to `whisper_n_text_ctx(ctx)/2`.

So to achieve the `--initial_prompt` functionality, you need to tokenize the input text and pass it to the `prompt_past` in the context. For example, the passing can be done by extending the C-interface with a `whisper_set_prompt()` function similar to `whisper_set_mel()`:

https://github.com/ggerganov/whisper.cpp/blob/a6c786d5dccf5f54a92da890f70715b4c9831172/whisper.h#L86-L96

And also extending the `whisper_full_params` struct to accept a `const char * prompt` string when `no_context == false`:

https://github.com/ggerganov/whisper.cpp/blob/a6c786d5dccf5f54a92da890f70715b4c9831172/whisper.h#L168

Regarding the tokenization of the text - you can try to directly use the GPT tokenization function that I implemented here:

https://github.com/ggerganov/ggml/blob/624e4f531370d254b4e06268506f704524f57dc9/examples/utils.h#L53-L65

It's probably not 100% correct, but I found that it generally works.

PR #163, opened 2022-11-21
Prompt previous tokens for streaming

meakbiyik:
ref #90 

@ggerganov a quick and dirty fix on using previous length's tokens as prompt. Beware that this is the first time I wrote cpp code, but it works - the prompt contains the correct tokens.

I first tried to tokenize a given string but that was quite problematic and beyond my nonexistent cpp capabilities. I decided to just use the tokens provided my the model, and I used a vector pointer instead of vector itself because it gave weird errors. Not sure if it is the good practice or not.

### DIFF
--- a/whisper.h
+++ b/whisper.h
@@ -3,6 +3,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <vector>
 
 #ifdef WHISPER_SHARED
 #    ifdef _WIN32
@@ -208,6 +209,7 @@ extern "C" {
         bool speed_up;  // speed-up the audio by 2x using Phase Vocoder
         int  audio_ctx; // overwrite the audio context size (0 = use default)
 
+        const std::vector<whisper_token> * prompt_tokens;
         const char * language;
 
         struct {
--- a/whisper.h
+++ b/whisper.h
@@ -3,7 +3,6 @@
 
 #include <stdint.h>
 #include <stdbool.h>
-#include <vector>
 
 #ifdef WHISPER_SHARED
 #    ifdef _WIN32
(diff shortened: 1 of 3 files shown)


whisper.h
```diff
@@ -208,6 +209,7 @@ extern "C" {
         bool speed_up;  // speed-up the audio by 2x using Phase Vocoder
         int  audio_ctx; // overwrite the audio context size (0 = use default)
 
+        const std::vector<whisper_token> * prompt_tokens;
```
ggerganov, line 212: The idea is correct, but the problem is we cannot afford to use C++ constructs like `std::vector` in the C-style API. If we use it, then it will break compatibility with other programming languages that want to use the library.

Instead of `std::vector<whisper_token>` you will have to pass `whisper_token * prompt_tokens` and `int prompt_n_tokens`. I know it's ugly, but it's portable.

meakbiyik: Tried to do it, no idea if this is what you mean. I hate it, I wish we could have the same performance with Go 😆

ggerganov:
Perfect! I'll merge and add some final touches

ggerganov:
Hmm, are you sure? The idea now is to aways set `no_context = true` in the `stream` example:

https://github.com/ggerganov/whisper.cpp/blob/dad109c3f188e8ad3dfc494de1c1b1c56b4aebd9/examples/stream/stream.cpp#L339

So the `prompt_past` is always cleared:

https://github.com/ggerganov/whisper.cpp/blob/dad109c3f188e8ad3dfc494de1c1b1c56b4aebd9/whisper.cpp#L2587-L2591

Then, if you optionally provide `-kc` to `stream`, the text from the previous segment will be passed.
If you don't provide `-kc`, then you fallback to the original behaviour where you never have context when using `stream`.

Let me know if you agree - I might be missing something, but I think the current version is OK and the `-kc` is now actually useful. Before the change, it was working incorrectly.

meakbiyik:
A-ha, indeed, I misunderstood the negative flag - and it seems the -kc in main is different from -kc in stream, so it looks good. Thanks again!

Issue #192, opened 2022-11-27
Cancelling a transcript


szeidner:
Is there a way to cancel a transcript that is in progress programmatically? Calling `whisper_free` results in the following error:

```
ggml_new_tensor_impl: not enough space in the context's memory pool
Assertion failed: (false), function ggml_new_tensor_impl, file ggml.c, line 1426.
```

I wondered if there was some way built-in to notify the main loop to stop running. For the time being, I added a `cancel` boolean flag to `whisper_ctx` that the main loop checks and ends if it's enabled. This works, but I just wondered if there was possibly a built in way to do this. Thank you!

PR #194, opened 2022-11-27
Add mechanism for aborting the whisper_full() computation

ggerganov:
ref #192

This is a bit more cumbersome compared to adding a mutex + bool flag in the `whisper_context`, but I want to avoid adding `<mutex>` or `<atomic>` to `whisper.cpp` for now.

The idea is to provide a callback that is called on every inference iteration before the encoder starts. If the callback returns `false`, the processing will be aborted. A usage example is provided in `main`:

https://github.com/ggerganov/whisper.cpp/blob/afadabef84faa3f28fc2b60fa04bf33d47366a41/examples/main/main.cpp#L610-L621

@szeidner
Would this approach work for you?

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -607,6 +607,19 @@ int main(int argc, char ** argv) {
                 wparams.new_segment_callback_user_data = &user_data;
             }
 
+            // example for abort mechanism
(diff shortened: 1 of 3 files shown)


szeidner:
@ggerganov Yes this would work for my purposes. Thank you for the quick response here!

PR #245, opened 2022-12-09
fix potential bug reading model data into a small size optimized string + wrapper

berthubert:
fix potential bug reading model data into a small size optimized string which could lead to memory corruption. In an SSO string, you can't write data to &str[0] and expect it to work well. 

Also added a small wrapper function to more safely read model data without having to get the sizeof right. I tested this on tiny, base and large models, there was no change in behaviour.

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -429,6 +429,12 @@ struct whisper_context {
     int32_t exp_n_audio_ctx; // 0 - use default
 };
 
+template<typename T>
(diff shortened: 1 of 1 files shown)


ggerganov:
Thanks for the contribution!
Very minor nitpick before we merge: `readSafe` should be `read_safe` in order to follow the `snake_case` naming convention in that source file.

berthubert:
I think it should be good now, my git(hub) is a bit rusty.

PR #287, opened 2022-12-18
Initial import of golang bindings

djthorpe:
I've added some bindings for golang here. They are pretty basic but I've tested on Darwin x64_86 and Debian arm64. There is one example (go-whisper) for the most basic case of using the bindings.

Linked to issue #269


ggerganov:
This looks very well done - thank you for the contribution!

Regarding the need for the `WHISPER_NO_AVX2=ON` CMake option - maybe if you use `make libwhisper.a` instead of CMake it would be more portable. The [Makefile](https://github.com/ggerganov/whisper.cpp/blob/1d716d6e34f3f4ba57bd9706a9258a0bdb008153/Makefile#L57-L84) has some logic to determine if AVX / AVX2 is available or not and add the necessary flags when possible.

I'll let this sit for a day before merging.
Also, hoping someone can help with adding a CI for this in the [build.yml](https://github.com/ggerganov/whisper.cpp/blob/master/.github/workflows/build.yml). Would make the bindings more resilient to future changes in the API. If not, we can merge and figure it out later.

djthorpe:
Thanks Georgi. On the AVX2 option, I'll take a look at using make and I can add the github actions for tests and building as well, but probably not going to get to them this week. I can send you some more pull requests for those changes later, and then begin to build out the full functionality and live streaming option in go examples. Thanks for your comments and continued efforts on this.

jaybinks:
@ggerganov any chance we can get this merged?
Its a bit hard for me to do any testing against this code before its merged, because of the way gomodules work with the module URL (which wont exist until you merge it)

jaybinks:
Awesome stuff !!
@david I think we need to look at a few little tidyups to simplify usage of
the go library.

It seems we can use environment variables to pass C and LD options to the
cgo implementation.

This would also allow someone to point to whisper.h outside of their go
application tree (keeping things near)

Have you ever done this before?

djthorpe:
Thanks. Yep, I'll fix that up and improve the documentation for it in the next few days:

- [ ] Add a github action to make/test on PR merge
- [ ] Use "make libwhisper.a" instead of cmake
- [ ] Use LD_LIBRARY_PATH and INCLUDE_PATH to point to the compiled whisper library and include files

Issue #278, opened 2022-12-15
Support for beam search decoding


noe:
First, thanks a lot for this wonderful project and all the effort you are putting into it.

I would like to know if adding support for beam search decoding is on your roadmap.

Issue #270, opened 2022-12-13
Improving timestamps for words


akatkov7:
Hi,

Wanted to start off with a big thanks for porting whisper to C++. It has been very useful for integration on iOS.

I wanted to open this issue to see if you had any thoughts or suggestions on how to address an issue we're seeing in one of our sample audio files.

Roughly 56 seconds into the audio file, one of the people says 'What?' after a longish pause (6-7 seconds) since the previous word.

Running in through whisper.cpp:
```
[49.9 --> 50.47] | What| (Confidence: 0.83800673)
```

Running it through openai/whisper:
```
>>> stab_segments[-3]
{'id': 26, 'seek': 2840, 'start': 49.4, 'end': 50.4, 'text': ' What?', 'tokens': [708, 30], 'temperature': 0.0, 'avg_logprob': -0.36360436898690685, 'compression_ratio': 1.6474820143884892, 'no_speech_prob': 0.00023565757146570832}
```

Running it through jianfch/stable-ts:
```
>>> results['segments'][-2]['whole_word_timestamps']
[{'word': ' What?', 'timestamp': 56.8799991607666}]
```

So it looks like stable-ts has made some changes that properly detects the timing of "What?". Looks like stable-ts has some silence detection that is likely aiding it in this scenario.

So my question is is there anything I can do on the input side to help improve these scenarios? Is the only solution adjusting the core logic in the library itself? Are some of the improvements in stable-ts scheduled to be added to this repo as well?

I have attached the sample audio file. (GitHub didn't like the wav directly so I zipped it up).

[bad_caption_timing.wav.zip](https://github.com/ggerganov/whisper.cpp/files/10220652/bad_caption_timing.wav.zip)

Thanks again.

ggerganov:
I'm not sure what is the best way to fix this. There is however an interesting effect that I have observed and is easily reproduced with your audio file. To demonstrate here are 4 different results with `whisper.cpp` for different starting offsets (0 ms, 100 ms, 200 ms, 300 ms):

<details>
  <summary>offset = 0 ms (default)</summary>

```java
$  ./main -m ./models/ggml-base.en.bin -f ./bad_caption_timing.wav
whisper_model_load: loading model from './models/ggml-base.en.bin'
whisper_model_load: n_vocab       = 51864
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 512
whisper_model_load: n_audio_head  = 8
[...]

Issue #255, opened 2022-12-10
Can the accuracy of the timestamp be improved?


czkoko:
The timestamp of whisper is not very accurate. 
The following is the comparison between Microsoft Cognitive Services Speech and whisper.
```
1                                    
00:00:00,120 --> 00:00:01,379 (Microsoft)    
[00:00:00.000 --> 00:00:02.000] (whisper)
2
00:00:02,120 --> 00:00:06,320 (Microsoft)  
[00:00:02.000 --> 00:00:07.500] (whisper)
```

misutoneko:
Yes, this would be much appreciated. I suppose you are using the large model?
I've found the smaller models to be more inaccurate.

ggerganov:
The timestamp precision is a limitation of the model. You would need some sort of pre/post-processing to improve the timestamps. But at the moment it is not clear what is the best approach.

pneyrinck:
Apparently, this work has been done to improve time stamps. https://github.com/jianfch/stable-ts

Issue #172, opened 2022-11-23
Model gets stuck in some words


CarlitosDev:
Last Whisper.cpp version On Mac M1
Model `ggml-medium.en.bin`
Additional parameters: `-t 8 -ml 1`
Mono audio file

It seems that the model gets _stuck_ in some words and misses the actual conversation.

![Screenshot 2022-11-23 at 11 54 09](https://user-images.githubusercontent.com/4416831/203542828-3ae49903-e997-4325-a440-08885dc0e0b4.png)
![Screenshot 2022-11-23 at 11 57 14](https://user-images.githubusercontent.com/4416831/203542864-71111370-92d0-4438-bcab-6f4dbbf29f94.png)

ggerganov:
I believe this is a known limitation of the model - see this discussion for more info:

https://github.com/openai/whisper/discussions/29

There are various strategies that can be added to reduce the occurrence of this behaviour (i.e. beam search decoding, temperature fallbacks, VAD, etc.). Some of these are already available in the original implementation from OpenAI, so you can try running it and see if this resolves your issue.

szeidner:
I've run into this issue as well, but see a difference between the output of Whisper (python) vs Whisper.cpp. While there are some repeated words in the python version of Whisper, there are pretty long sections where a phrase is repeated (up to 8 minutes or so) with Whisper.cpp. I wonder if there is anything that can be done to improve the behavior. Do you think maybe this difference is due to using beam search decoding or something similar in the original implementation? If so, I wonder how difficult it would be to implement that in c++?

I've attached the output from both versions of whisper for comparison. I ran it on this [podcast episode](http://www.merlinmann.com/roderick/ep-483-the-power-girls.html) with the tiny model used for both runs.

[whisper.python.txt](https://github.com/ggerganov/whisper.cpp/files/10195881/whisper.python.txt)
[whisper.cpp.txt](https://github.com/ggerganov/whisper.cpp/files/10195882/whisper.cpp.txt)

ggerganov:
@szeidner 
Yes, it's likely due to the inferior decoding strategy in `whisper.cpp`.
I've made some improvements lately - you might give it another try, but probably your case is still going to fail.
I think we need the `temperature` feature from the OpenAI decoding method to fix this.
Implementation is not very difficult, but I keep prioritising other stuff.

geimist:
I also keep having this problem, which is why I keep having to discard tasks, unfortunately. A workaround would be great. 👍

[...]

PR #291, opened 2022-12-18
Improve decoding

ggerganov:
ref #278 #133 #172 #255 #270 

The goal of this PR is to reach OpenAI decoding parity and potentially go beyond

There are several ideas for improving the decoding strategy that will be explored.
There is some chance that these ideas will improve segment and token timestamp precision, but no guarantees.

---

### Implemented decoding strategies

- Average log probability threshold support
  Decoded sequences can be discarded based on the average logprob of the tokens. When the avg logprob is below the threshold, it means that the model wasn't very confident in the transcription and we need to apply some fallback strategy to generate a better sequence
- Entropy-based threshold support
  This is similar to OpenAI's compression ration threshold logic used to determine if a sequence is too repetitive. However, in `whisper.cpp` instead of using `zlib` compression, we use a basic entropy metric `H = -sum(p*log(p))` of the last 32 tokens in the sequence to determine if the decoding has degraded in endless repetition. Low entropy means more repetition. This approach has to be further tested - probably the entropy threshold might need some adjustments
- Temperature support
  By default, the decoding starts with `T = 0`, deterministically sampling the best token each time based on the computed logits. Upon failure, we increase the temperature and we start sampling the tokens from a discrete probability distribution obtained by scaling the logits with `1/T`
- The `Greedy` decoding strategy
  Uses `--best-of` number of independent decoders for `T > 0`. Each decoder keeps a separate decoding sequence. At temperature `T > 0.5` we clear any previous context. The rationale is that sometimes the context can confuse the decoder and drive it into a failure case
- The `BeamSearch` decoding strategy
  At `T = 0` we start with `--beam-size` independent decoders. Each one generates the top `--beam-size` sequences from it's current state. From all generated candidate sequences, we pick the top `--beam-size` based on the logprob sum of their tokens and reassign them to the decoders. Upon failure, we increase the temperature and fallback to the `Greedy` strategy. The `BeamSearch` decoder is `--beam-size` times more computationally heavy than the `Greedy` decoder
  I think it is worth exploring a strategy which initially uses 1 beam at `T = 0` and only activates `--beam-size` decoders upon failure. This would significantly speed-up the processing and I hope it will keep the transcription quality high. Will probably add a flag for that

---

### Development notes

- `best_of` is used only by the Greedy decoder at `temperature > 0`
- `beam_size` is used by the BeamSearch decoder
- Both `best_of` and `beam_size` require to maintain a separate KV memory for each decoder stream. Need changes both in `whisper.h` interface + `whisper_context` / `whisper_model` to support that. Introduce `whisper_decoder`
- `compression_ratio` [heuristic](https://github.com/openai/whisper/blob/0b1ba3d46ebf7fe6f953acfd8cad62a4f851b49f/whisper/utils.py#L25-L28) might be out-of-scope - I cannot implement `zlib.compress` from scratch. Maybe use something simpler, like n-gram entropy?
  <details>
    <summary>ChatGPT brainstorming</summary>

    image
    image

  </details>
- Clear past prompt only for `temperature >= 0.5`:

  image
- [patience](https://github.com/openai/whisper/blob/0b1ba3d46ebf7fe6f953acfd8cad62a4f851b49f/whisper/decoding.py#L283) controls the max number of sequences to obtain from the beam search
- BeamSearch algorithm is explained here: [https://arxiv.org/pdf/2204.05424.pdf](https://arxiv.org/pdf/2204.05424.pdf)
- For each decoded sequence, maintain `avg_logprob` of the tokens in order to implement `logprob_threshold` fallback:
  https://github.com/openai/whisper/blob/0b1ba3d46ebf7fe6f953acfd8cad62a4f851b49f/whisper/transcribe.py#L119-L120
- Sequence ranking with and without `length_penalty`:
  https://github.com/openai/whisper/blob/0b1ba3d46ebf7fe6f953acfd8cad62a4f851b49f/whisper/decoding.py#L169-L192

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ build/
 build-em/
 build-debug/
 build-release/
+build-static/
 build-sanitize-addr/
 build-sanitize-thread/
 
@@ -18,6 +19,7 @@ build-sanitize-thread/
 /bench
 
 sync.sh
+libwhisper.a
 libwhisper.so
 compile_commands.json
 
--- a/README.md
+++ b/README.md
@@ -212,17 +212,7 @@ make large
 ## Limitations
 
 - Inference only
-- No GPU support
-- Very basic greedy sampling scheme - always pick up the token with highest probability.
-  This should be similar to the [GreedyDecoder](https://github.com/openai/whisper/blob/main/whisper/decoding.py#L249-L274)
-  from the original python implementation, so in order to make a fair comparison between the 2 implementations, make sure
-  to run the python code with the following parameters:
-
-  ```
-  whisper --best_of None --beam_size None ...
-  ```
-
-  In the future, `whisper.cpp` will support more sampling strategies.
+- No GPU support (yet)
 
 ## Another example
 
--- a/examples/command/command.cpp
+++ b/examples/command/command.cpp
@@ -671,56 +671,81 @@ int process_command_list(struct whisper_context * ctx, audio_async &audio, const
                 break;
             }
 
-            const auto * probs = whisper_get_probs(ctx);
-            std::vector<std::pair<float, int>> probs_id;
-
-            double psum = 0.0;
-            for (int i = 0; i < (int) allowed_commands.size(); ++i) {
-                probs_id.emplace_back(probs[allowed_tokens[i][0]], i);
-                for (int j = 1; j < (int) allowed_tokens[i].size(); ++j) {
-                    probs_id.back().first += probs[allowed_tokens[i][j]];
-                }
-                probs_id.back().first /= allowed_tokens[i].size();
-                psum += probs_id.back().first;
-            }
-
-            // normalize
-            for (auto & p : probs_id) {
-                p.first /= psum;
-            }
-
-            // sort descending
+            // estimate command probability
+            // NOTE: not optimal
             {
-                using pair_type = decltype(probs_id)::value_type;
-                std::sort(probs_id.begin(), probs_id.end(), [](const pair_type & a, const pair_type & b) {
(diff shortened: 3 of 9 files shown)


ggerganov:
This should be pretty close to OpenAI's decoding implementation.
There might be a few bugs left, but I think it's pretty much ready to merge.

The few failure cases that I have are now correctly transcribed using either `--best-of 5` or `--beam-size 5`
By default, I am thinking about leaving beam-search off and have a temperature + `--best-of 5` fallback.

stevevaius2015:
Thank you for your immense work and this wonderful project

RndyP:
Getting a potential uninitialized variable tid on line 3124 of Whisper.cpp reported by VC++

RndyP:
First, thanks for all the hard work on this!

I am playing around with 1.1.0 as I write this.

Still have the issue that was closed in #172. The problem is worse now in that the "echos" may eat up CPU like crazy. My test case is to repeat the number "six" multiple times. (Sorry about the 666 humor) I send whisper an audio block of about 4 seconds with the word "six" repeated at least 3 times, and Whisper will now, instead or returning with a large number of sixes, will crunch for up to 1 minute and return various odd strings.

ggerganov:
@RndyP 
Thanks for the reports - these are very useful.
The repetition problem is due to the new entropy-besed threshold for discarding repetitive sequences. I will fine-tune the parameters to try to avoid the case that you observe and also add parameters to enable/disable and control this functionality for the `main` example.

dennislysenko:
Thanks for the fixes 👍

PR #308, opened 2022-12-22
Whisper.swiftui

Digipom:
I created a SwiftUI project in the same spirit as the Android project. This way we could have both an example of how to integrate using Objective-C, and an example using SwiftUI which can also be used to target the Mac using the same codebase.


ggerganov:
Wow! What an excellent contribution!
I really like that it runs both on macOS and iOS without any modification.

PR #319, opened 2022-12-22
whisper: Use emplace_back in place of push_back

asmaloney:
This avoids potential construction of temporaries.

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -1863,7 +1863,7 @@ static whisper_token_data whisper_sample_best(
     probs_id.reserve(n_logits);
 
     for (int i = 0; i < n_logits; i++) {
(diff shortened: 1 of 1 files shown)


ggerganov:
@asmaloney Thanks for the contributions!

asmaloney:
My pleasure. Thank you for the project!

I'm having fun playing with it. I probably won't have any deep contributions since I don't understand the details, but I'm happy to do some little things. I think it's an important project because it lets people embed this tech into projects that otherwise don't have easy access to it (because... Python).

I can imagine lots of accessibility scenarios where it could be used to help people who have problems with fine motor skills. And with your other language bindings it really opens it up! (e.g. I use _go_ a lot.)

Have a great xmas weekend! 🎄

PR #336, opened 2022-12-27
Build ES6 modules

matteeyah:


bindings/javascript/CMakeLists.txt
```diff
@@ -30,8 +30,9 @@ if (WHISPER_WASM_SINGLE_FILE)
 endif()
 
 set_target_properties(${TARGET} PROPERTIES LINK_FLAGS " \
-    --bind \
+    -lembind \
```
matteeyah, line 33: `--bind` seems to be deprecated in favor of `-lembind`.

https://emscripten.org/docs/tools_reference/emcc.html#emcc-bind

Screenshot 2022-12-27 at 01 04 51

matteeyah:
I just realized `EXPORT_ES6` is not compatible with builds that target `node`.

matteeyah:
Actually, this is supported since https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#3127---112922

matteeyah:
But the outputted JavaScript still seems to be using `require`s all over, so it doesn't really output ES6 compatible code.

PR #340, opened 2022-12-27
Similar to Whisper PR#228, this adds -ocsv, aka --output-csv, writing CSV file containing millisecond timestamps

NielsMayer:
For efficiency in storage, parsing and accuracy from representing timestamps in milliseconds (rather than floating-point numbers), this PR, like https://github.com/openai/whisper/pull/228 adds feature to output CSV file containing lines formatted like:
` <startTime-in-integer-milliseconds>, <endTime-in-integer-milliseconds>, "<transcript-line-including-commas>"`

The CSV file is created when option `-ocsv`, or `--output-csv` is provided to the 'main' command-line.

Similar to the output file naming conventions for `--output_vtt` or `--output_srt`, the resulting CSV file for an input file named `<mediaBaseName>.wav` is named `<mediaBaseName>.wav.csv`

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -69,6 +69,7 @@ struct whisper_params {
     bool output_vtt     = false;
     bool output_srt     = false;
     bool output_wts     = false;
+    bool output_csv     = false;
     bool print_special  = false;
     bool print_colors   = false;
     bool print_progress = false;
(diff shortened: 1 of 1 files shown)


NielsMayer:
One of the big advantages of whisper.cpp is that when you're processing a 31 HOUR video file, it doesn't just die with "KILLED" and no other debug info, like whisper does on the exact same 16k .wav file (alongside medium model, 16 threads, 4.45Ghz 8 core AMD 4750g CPU): 

image

I also like how whisper.cpp correctly transcribes a C++ header file as "child.h" and not, say "child dot h" . (original source video: https://www.youtube.com/watch?v=8jLOx1hD3_o ) :-)

PR #343, opened 2022-12-29
Attempt to improve threading in ggml

ggerganov:
ref #200 

Instead of creating and joining new threads for each graph compute, we create a thread pool with active threads and make them wait on a condition variable instead of joining.

Surprisingly, this does not seem to improve the performance.
It is actually degraded. Not sure what is going on.

### DIFF
--- a/ggml.c
+++ b/ggml.c
@@ -304,6 +304,91 @@ int64_t ggml_cycles_per_ms(void) {
 #endif
 
 //
+// thread data
+//
+// synchronization is done via busy loops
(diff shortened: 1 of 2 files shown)


heshpdx:
Lot's of things have changed in six months, including that the segfault that happened when running more tha 18 threads. Should we try this again? I attempted to rebase to mainline and tried to resolve the conflicts, but my runs livelock... so I clearly missed something. If you can rebase this branch and get it functional, I can help test.

cook-jeremy:
Would also like to see this rebased, as I'm having threading issues when running this as an XPC Service.

PR #351, opened 2022-12-31
cmake : Update to more modern version

asmaloney:
- update from 3.0 (from 2014) to 3.19 (from 2020)
- move some global setting onto the targets (through a cmake include)

Right now it's not possible to include whisper.cpp using `add_subdirectory()` because it modifies the global cmake state (C++ version, install paths, etc.).

This PR starts to move it towards a proper target-based approach.

I created a cmake include file to more easily set the defaults on the many examples, though it could just as well be copied into each file if you like.

I don't have a way to test the EMSCRIPTEN changes, but they mirror all the other examples.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,15 +1,14 @@
-cmake_minimum_required (VERSION 3.0)
+cmake_minimum_required (VERSION 3.19)
 
 project(whisper.cpp VERSION 1.0.4)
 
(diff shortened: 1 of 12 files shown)


djthorpe:
Debian Bullseye (the latest version) has CMake 3.18 by default, and this also affects Raspberry Pi's. So I recommend bumping the CMake version down to 3.18

PR #353, opened 2023-01-01
Add loader class to allow loading from buffer and others

prsyahmi:
Hi, I added a class to allow flexible loading of whisper model from any source.

Please note that I made a breaking change (renaming `whisper_init` to `whisper_init_from_file`) and didn't test go, wasm, emscripten and iOS/apple related.


ggerganov:
Thanks for this - on first look, it seems well done!
I will review the changes in more details in the following days.

ggerganov:
Thanks for this and sorry for force-pushing in your repo - not sure if there is a better way to resolve merge conflicts when the PR is from another remote.

Issue #332, opened 2022-12-24
Suggestion: Add .clang-format file


asmaloney:
It would be useful when making larger changes (like [this](https://github.com/ggerganov/whisper.cpp/pull/331)) if it could be properly formatted using clang-format.

Do you happen to have a `.clang-format` file you can drop into the repo?

ggerganov:
I don't have a `.clang-format` - it would be great to create one that matches the existing style if possible

asmaloney:
In my experience it will be tricky to match your style 100%.

If you're willing to compromise on a few things I believe getting code formatting automated and out of the way is a win for any project.

I'll work on it to see how close I can get to the existing code then you can assess it and tweak it if you like.

ggerganov:
I think I've seen a tool that can analyse a codebase and try to generate `.clang-format` that is very close to the existing style. I might be wrong though

asmaloney:
I've tried a couple with not-so-great results, but maybe I haven't found the right one yet.

The tricky thing is that [clang-format options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) are complicated and amended frequently. C++ formatting is a mess and has caused so many wasted hours (and bugs) - I like go's approach with `gofmt`! 😄

PR #359, opened 2023-01-02
Add .clang-format

asmaloney:
Here is a first attempt to clang-format as close to the existing format as possible ([#332](https://github.com/ggerganov/whisper.cpp/issues/332)).

I formatted the whisper and ggml files to show what it does. If there's anything here you absolutely won't accept, please let me know and I'll see if I can fix it or find a workaround.

Notes:
- The header files were not consistent with indenting the "extern" block. I chose the no-indent style, but it's easy to change.
- Some table-like declarations in whisper.cpp are almost impossible to format the way they were, so the formatter was turned off for them using `// clang-format off`.
- The unusual formatting in the case statement in `whisper_full_default_params` got mangled. We can either change the way this is commented or turn off formatting for that section as well.

As noted, some formatting can be turned off using `// clang-format off` but in general I think this should be discouraged.


ggerganov:
On second thought, I think `.clang-format` is not a good option.
In this project, I use a unique code formatting where sequential lines are vertically aligned and I really like it this way.
I don't think there is a way for `.clang-format` to achieve this.

asmaloney:
If you mean these:
```cpp
struct ggml_tensor * ggml_scale(
        struct ggml_context * ctx,
        struct ggml_tensor  * a,
        struct ggml_tensor  * b);

struct ggml_tensor * ggml_scale(struct ggml_context * ctx, struct ggml_tensor * a, struct ggml_tensor * b);
```

Indeed at the time I could not figure out how to force that properly. 🤔 

It's just that without auto-formatting it makes it more difficult for others to contribute & you will probably have to waste time paying attention to & fixing the formatting for any PRs (especially since the code already isn't self-consistent).

PR #362, opened 2023-01-02
go : go.sum should be included in git

asmaloney:
https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file


ggerganov:
cc @djthorpe

PR #365, opened 2023-01-03
Convert POWER9 fp16 SIMD code to macros

fitzsim:
I changed GGML_F16_VEC_LOAD and GGML_F16_VEC_STORE to pass the index as a separate argument.  I haven't tested those changes on the other architectures.  Will they be tested automatically before the merge?


fitzsim:
I'll submit a new pull request for this later; I want to try to eliminate the need for the load/store argument changes.

Issue #300, opened 2022-12-21
Very slow - any way to speed up?


luke-jr:
Per #10,

>Noting that the processing time is considerably shorter than the length of speech,

Yet even using 64 threads, it's taking days to process minutes of audio on my POWER9.

Has something changed since #10, or is there something I am doing wrong?

RndyP:
What is your platform?

luke-jr:
Yeah, I'm not surprised it isn't optimised for PPC64, but I wouldn't expect it to be magnitudes slower either. Real-time to days is a huge difference. :/

ggerganov:
For example on my `Ryzen 9 5950X` if I remove the `-mavx -mavx2 -mfma -mf16c` flags I observed about x50 slower computation of the `bench` tool. Removing those flags is similar to what you have on the PPC64 - no SIMD, no F16 support.

SIMD can make a huge difference

fitzsim:
ChatGPT is out-of-date regarding the Power ISA being proprietary.  It is open source now, just like RISC-V.  See [https://openpowerfoundation.org/](https://openpowerfoundation.org/).

luke-jr:
After #320, `./main -m models/ggml-base.en.bin -f samples/jfk.wav` takes 15.7 seconds.

Additional options | Time
--------------|--------
`-p 64` | 77s
`-t 64` | 28s
`-t 1` | 59.6s
`-t 16` | 5.8s
`-t 32` | 5.1s
`-m models/ggml-large.bin -t 32` | 111.1s

>ChatGPT is out-of-date regarding the Power ISA being proprietary.

In my experience, ChatGPT tends to be wrong quite often.

ggerganov:
@fitzsim @luke-jr 
I am planning to merge a refactored version of the SIMD routines in `ggml` which I think will make things easier to maintain in the future. The PR is pretty much ready in #324 

All instruction sets fit quite nicely in the proposed pattern, but I'm having a little trouble with the `ppc64le` stuff since I don't have a way to test it. So for the moment, I've special-cased it, which is not great.

If you are interested and have some free time, you can take a look at the implementation and see if you can fit it in the new pattern. Or at the very least - run a test and see that it still works after the changes.

Regarding the new performance: 5s for `jfk.wav` using `base` still seems quite a lot. Not sure why the performance is so bad

fitzsim:
[...]

PR #366, opened 2023-01-03
Enable POWER9 fp32 and fp16 SIMD code

fitzsim:
With the FP32 base model and this patch set, the jfk example takes about 3.2 seconds to transcribe.  This is another data point for #300, and it is about one second faster than the current FP16 SIMD code.

### DIFF
--- a/ggml.c
+++ b/ggml.c
@@ -386,14 +386,14 @@ static const size_t CACHE_LINE_SIZE_F32 = CACHE_LINE_SIZE/sizeof(float);
     #define GGML_F16_STEP 32
     #define GGML_F16_EPR  8
 
-    #define GGML_F16x8              float16x8_t
-    #define GGML_F16x8_ZERO         vdupq_n_f16(0.0f)
-    #define GGML_F16x8_SET1(x)      vdupq_n_f16(x)
-    #define GGML_F16x8_LOAD         vld1q_f16
-    #define GGML_F16x8_STORE        vst1q_f16
-    #define GGML_F16x8_FMA(a, b, c) vfmaq_f16(a, b, c)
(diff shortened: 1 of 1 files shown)


ggml.c
```diff
@@ -526,39 +526,31 @@ static const size_t CACHE_LINE_SIZE_F32 = CACHE_LINE_SIZE/sizeof(float);
...
+#define GGML_F32x4_REDUCE(sumf, sum)                            \
+  sum[0] = vec_add(sum[0], sum[1]);                             \
+  sum[2] = vec_add(sum[2], sum[3]);                             \
+  sum[4] = vec_add(sum[4], sum[5]);                             \
+  sum[6] = vec_add(sum[6], sum[7]);                             \
+  sum[0] = vec_add(sum[0], sum[2]);                             \
+  sum[4] = vec_add(sum[4], sum[6]);                             \
+  sum[0] = vec_add(sum[0], sum[4]);                             \
+  sumf = vec_extract(sum[0], 0) + vec_extract(sum[0], 1)        \
+    + vec_extract(sum[0], 2) + vec_extract(sum[0], 3);
```
ggerganov, line 553: Is there a reason to use this version instead of the `for`-based version?
The advantage of the latter is that it will work for `GGML_F32_ARR == 1, 2, 4, 8, 16`, while doing it like this it will only work for `GGML_F32_ARR == 8`

fitzsim:
I'll try reverting that and compare speed.  Also, I think I can get rid of the load/store argument changes, but it'll take some more work.  And the F32 implementation seems to still use some F16 operations, so I'll investigate that.  I'll make a new pull request for all this later.  For now I'll close this one.

PR #369, opened 2023-01-04
Reorganize POWER9 SIMD code

fitzsim:
I could not eliminate the separate index argument in the f16 load and store macros, so this patch set needs testing on other architectures.

The existing GGML_F32x4_REDUCE macro performs as well as the implementation in #366 so I used the existing one.

When I test with the F32 model, ggml_vec_dot_f16 and ggml_vec_mad_f16 are still being called.  Is that expected?

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ High-performance inference of [OpenAI's Whisper](https://github.com/openai/whisp
 - Plain C/C++ implementation without dependencies
 - Apple silicon first-class citizen - optimized via Arm Neon and Accelerate framework
 - AVX intrinsics support for x86 architectures
+- VSX intrinsics support for POWER architectures
 - Mixed F16 / F32 precision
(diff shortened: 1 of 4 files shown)


ggerganov:
> When I test with the F32 model, ggml_vec_dot_f16 and ggml_vec_mad_f16 are still being called. Is that expected?

Yes - good point. I have to update `whisper.cpp` to not use F16 when in F32 mode.
Currently it stores some of the intermediate tensors in F16 which leads to the `ggml_vec_dot_f16` and `ggml_vec_mad_f16` pathways.

ggerganov (approved):
Tested all platforms - everything is good!
Ready to merge if you are good?

fitzsim:
Yes, I don't have any changes left to push.  Thanks for reviewing and testing the patch set.

PR #379, opened 2023-01-06
go bindings updated so they can be used in third party packages.

djthorpe:
* go.sum should be included in the repository #362
* Removed LDFLAGS and CFLAGS so that bindings can be used in third partt packages
* Added some more README documentation
* Simplified go-model-download slightly
* Added `Whisper_n_audio_ctx` binding


djthorpe:
Three other changes, which you can edit as needed:

* In the main Makefile, for Darwin target, checks for processor FMA instructions before adding -mfma compiler flag. My Xeon processor (MacPro 2013) does not have this
* Updated the minimum CMake version to 3.18 - this is the default version for Debian bullseye, which is the latest Debian
* CMakeLists.txt added WHISPER_NO_FMA option for Darwin for my very old Mac Pro

PR #381, opened 2023-01-06
Support AVX2 in windows better

Mike-Bell:
I spent quite a while trying to figure out why AVX2 wasn't enabled when I built this on Windows in my build system. Adding `/arch:AVX2` to the C-compiler seems to fix it. I'm _not_ building through Visual Studio but via a more customized build system (leveraging MSBuild), so that's likely why this isn't required for other people building on Windows, but since ggml.c is a C file (not C++), it seems like this should be correct and necessary in some cicrumstances.


RndyP:
Thanks Mike! I had tried AVX2 and it worked, but was extremely slow. The JFK.wav sample took 17 seconds, where with AVX it takes 4.5 seconds. With AVX2 set under C/C++ Code Generation -> Enable Enhanced Instruction set -> Advanced Vector Extensions 2 (/arch:AVX2) time went from 4.5 to about 3.6 seconds. Nice improvement.

RndyP:
I did some resaearch into what's going on here. One of the most critical sections of code is in ggml_vec_dot_f16(). AVX2 has fused multiply add instructions and AVX does not. So with AVX2 we have:
image
Without (just AVX) we have:
image
 Note in the AVX2 case the disassembly shows the use of the fused intrinsics

ggerganov:
@RndyP 
So recently, I was revisiting the SIMD code in `ggml.c` and I came to the realisation that we don't use AVX2 at all because all intrinsics that we currently use are either AVX, FMA or F16C. This is based on lookup in the following site:

https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html

For example, the case that you refer to I think depends actually depends on wether FMA is supported or not:

https://github.com/ggerganov/whisper.cpp/blob/87dd4a30811ee07700ee6fee267508e8935b32fc/ggml.c#L483-L487

So it is strange that AVX2 flags make any difference at all on Windows. My understanding is that AVX2 should not make a difference, but I might be missing something.

RndyP:
I think what's happening is the Microsoft compiler is smart enough to convert the _mm256_add_ps(_mm256_mul_ps(b, c), a) to the fused assembly version, when the compiler option is set. The code above goes from 16 down to 12 vector  instructions. 4.5 to 3.6 seconds is a nice improvement.

PR #383, opened 2023-01-07
Command: always test the prompt

dnhkng:
This adds a new mode for the Command example.

Adding a custom prompt parameter with `-p `or `--prompt` activates this mode:

`./command --prompt "OK whisper, please"`

Command will only output valid commands, which must always be prepended by the given prompts and have a word count > 0. Otherwise, nothing is output.

This is designed to match the behaviours of devices like Alexa and Siri.

### DIFF
--- a/examples/command/command.cpp
+++ b/examples/command/command.cpp
@@ -11,6 +11,8 @@
 #include <SDL.h>
 #include <SDL_audio.h>
 
+#include <iostream>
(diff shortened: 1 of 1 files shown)


ggerganov:
Haha - very nice work! I tried it and it works perfect!
Will merge this and do some clean-up and refactoring in the `command.cpp` on `master`

PR #387, opened 2023-01-08
Add process-specific timings

### LINKED ISSUE #382: Should we use process time for benchmarks/timings?
Currently, the `bench` tool simply spits out the timings from `whisper_print_timings`. These timings are not process specific and as such are influenced by other processes on the system.

Should we use something like `CLOCK_PROCESS_CPUTIME_ID` instead of `CLOCK_MONOTONIC` in `ggml_time_*`? This would make the benchmarks in #89 outdated, but would make future benchmarks more realistic.

abitofevrything:
Adds functions that use `CLOCK_PROCESS_CPUTIME_ID` instead of `CLOCK_MONOTONIC` for timings, and are therefore not affected by other processes on the system.

One thing to check before we merge: since the time is per-process, the time reported is `n_threads` times more than realtime since each active thread adds to the timings - should we adjust for this?

Also updated bench-all to include information about the measured parameters, better formatting and better configuration detection.

Closes #382.

### DIFF
--- a/ggml.c
+++ b/ggml.c
@@ -288,46 +288,46 @@ void ggml_time_init(void) {
     timer_freq = frequency.QuadPart;
 }
 
-int64_t ggml_real_time_ms(void) {
+int64_t ggml_time_real_ms(void) {
     LARGE_INTEGER t;
     QueryPerformanceCounter(&t);
     return (t.QuadPart * 1000) / timer_freq;
 }
 
-int64_t ggml_real_time_us(void) {
+int64_t ggml_time_real_us(void) {
     LARGE_INTEGER t;
     QueryPerformanceCounter(&t);
     return (t.QuadPart * 1000000) / timer_freq;
 }
 
-int64_t ggml_process_time_ms(void) {
+int64_t ggml_time_proc_ms(void) {
     return (clock() * 1000) / CLOCKS_PER_SEC;
 }
 
-int64_t ggml_process_time_us(void) {
+int64_t ggml_time_proc_us(void) {
     return (clock() * 1000000) / CLOCKS_PER_SEC;
 }
 #else
 void ggml_time_init(void) {}
-int64_t ggml_real_time_ms(void) {
+int64_t ggml_time_real_ms(void) {
     struct timespec ts;
     clock_gettime(CLOCK_MONOTONIC, &ts);
     return (int64_t)ts.tv_sec*1000 + (int64_t)ts.tv_nsec/1000000;
 }
 
-int64_t ggml_real_time_us(void) {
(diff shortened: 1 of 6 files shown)


ggml.c
```diff
@@ -287,49 +287,51 @@ void ggml_time_init(void) {
...
-#ifdef GGML_PERF
-#define ggml_perf_time_ms()       ggml_time_ms()
-#define ggml_perf_time_us()       ggml_time_us()
-#define ggml_perf_cycles()        ggml_cycles()
-#define ggml_perf_cycles_per_ms() ggml_cycles_per_ms()
-#else
-#define ggml_perf_time_ms()       0
-#define ggml_perf_time_us()       0
-#define ggml_perf_cycles()        0
-#define ggml_perf_cycles_per_ms() 0
```
ggerganov, line 332: This is actually needed and should not be removed.

The idea is to use the `ggml_perf_...` macros in `ggml.c` in order to time the performance of critical sections like `ggml_compute_forward_...` and others. However, since these sections can be invoked many hundreds of times and the clock system calls are not infinitely cheap, we want to be able to disable these measurements.

This is what the `GGML_PERF` flag does - when we are developing and measuring the performance, we define `GGML_PERF` and use `ggml_graph_print()` to see timing results for various `ggml` operations. However, when we are ready, we want to disable `GGML_PERF` in order to make all clock calls noop.

The basic rule is that all clock calls inside the `ggml.c` source file should be done via the `ggml_perf_...` macros in order to be able to disable them.

ggerganov (changes requested):

ggml.c
abitofevrything: Makes sense. I've added it back, but removed the cycle and millisecond related macros as they aren't used anywhere (cycles having been replaced with process time).

ggerganov (approved):

ggerganov:
@abitofevrything
Can you verify that everything is good on Windows?
After that I will merge it

abitofevrything:
Gave it a test and it works fine on my Windows machine

abitofevrything:
Tested on my Windows machine and there seems to be some issues with using `clock()`. Apparently its implementation is OS-specific and on Windows it returns wall clock time instead of process time.

Apparently there is a `GetProcessTimes` to get process time, but I'll need to look into that later.

abitofevrything:
Well, I changed the implementation to use GetProcessTimes but the issue with the timing still persists: process time is sometimes counted as negative.

I'm pretty sure this is an underflow/overflow issue, or an issue with the wrong data type being used somewhere, but I don't have a setup to debug this. I'll see what I could do, but if anyone used to windows development has any idea what is happening here I'd be glad if you could let me know.

Issue #188, opened 2022-11-27
Using `--max-len` gives weird time codes


niksedk:
When running whisper.cpp with e.g. `--max-len 77` I get some weird time codes.
It does not happen when not using `--max-len`.

Examples:

````
[00:34:35.820 --> 00:34:36.820]   You built that with your brothers.

[00:-3:-29.-700 --> 00:41:15.820]   with and you were gone more than you were home.

[00:41:11.820 --> 00:-3:-29.-700]   They resent the hell out of you because you were the parent they were left

[00:41:08.820 --> 00:41:11.820]   Most of your children don't speak to their mother.

````

````
[01:03:24.820 --> 01:03:26.820]   Well, have the Grayson brothers signed it yet?

[00:18:36.250 --> 01:03:24.820]   was a solid deal, and you weren't here, so paperwork's already been filed.

[01:03:15.820 --> 00:18:36.250]   Listen, Gabrielle convinced the other partners that the Playground merger

````

ggerganov:
The `--max-len` option uses the [experimental approach for word-level timestamps](https://github.com/ggerganov/whisper.cpp/blob/164df0d447bd16a4f6cd0e08caef5b01a63354af/whisper.cpp#L3126) that I implemented and it probably has a bug. We might be able to debug it if you uncomment the following lines and send me the output around the timestamps that get messed up:

https://github.com/ggerganov/whisper.cpp/blob/164df0d447bd16a4f6cd0e08caef5b01a63354af/whisper.cpp#L3343-L3353

It should look something like this:

```java
...
whisper_exp_compute_token_level_timestamps:        [?]  0.999  0.000  0.000  3.010 10044 10057 ' can'
whisper_exp_compute_token_level_timestamps:        [?]  0.996  0.012  0.000  4.010 10060 10082 ' find'
whisper_exp_compute_token_level_timestamps:        [?]  1.000  0.998  0.000  4.010 10082 10102 ' hope'
whisper_exp_compute_token_level_timestamps:        [?]  0.994  0.082  0.000  3.010 10110 10124 ' and'
whisper_exp_compute_token_level_timestamps:        [?]  0.999  0.095  0.000 11.010 10124 10193 ' inspiration'
whisper_exp_compute_token_level_timestamps:  [_TT_722]  0.940  0.603  0.057  2.010 10196 10206 ' in'
[...]

PR #388, opened 2023-01-08
Do sample_to_timestamp calculation with 64 bit precision to avoid overflow

boolemancer:
As written, the `sample_to_timestamp` function will overflow around 22 and a half minutes and it will start returning negative timestamps.

22m23s = 1343s
1343s * 16000 samples/s = 21488000 samples
Passing 21488000 to `sample_to_timestamp` would result in 100*21488000 (2148800000), which overflows to -2146167296 before that result is divided again by the sample rate.

By doing the multiplication with 64 bit precision, you avoid the overflow and you can now process audio clips longer than 22 minutes.

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -3387,8 +3387,8 @@ static int timestamp_to_sample(int64_t t, int n_samples) {
     return std::max(0, std::min((int) n_samples - 1, (int) ((t*WHISPER_SAMPLE_RATE)/100)));
 }
 
-static int64_t sample_to_timestamp(int64_t i_sample) {
-    return (100*i_sample)/WHISPER_SAMPLE_RATE;
+static int64_t sample_to_timestamp(int i_sample) {
+    return (100ll*i_sample)/WHISPER_SAMPLE_RATE;
 }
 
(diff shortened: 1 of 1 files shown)


whisper.cpp
```diff
@@ -3387,7 +3387,7 @@ static int timestamp_to_sample(int64_t t, int n_samples) {
     return std::max(0, std::min((int) n_samples - 1, (int) ((t*WHISPER_SAMPLE_RATE)/100)));
 }
 
-static int64_t sample_to_timestamp(int i_sample) {
+static int64_t sample_to_timestamp(int64_t i_sample) {
     return (100*i_sample)/WHISPER_SAMPLE_RATE;
```
ggerganov, line 3391: Great find!
I think this actually solves #188 

Wouldn't it be better to make the cast via the `100ll` literal like this:

```suggestion
static int64_t sample_to_timestamp(int i_sample) {
    return (100ll*i_sample)/WHISPER_SAMPLE_RATE;
```

boolemancer: That's actually how I implemented it at first, but I didn't see many other uses of literal suffixes in the codebase, so I changed it.

Updated the PR.

PR #390, opened 2023-01-08
ggml: correct behaviour of ggml_vec_sum_f32

abitofevrything:
Assuming the new implementation with accelerate is correct, the implementation of ggml_vec_sum_f32 has been wrong since the initial release. This PR corrects the behaviour.


ggerganov:
Oops 🤦 
I introduced this bug yesterday with d51fc3ee0a0038cdf1522ca3d58b58299de41eb8

Thanks for spotting this!

ggerganov:
Ah, I see what you mean: for some reason we always had `*s += sum` and now we have `*s = sum`.
Hmm.. not sure which one is correct. Need to see how it is used.

PR #398, opened 2023-01-11
PPC64 big-endian support

fitzsim:


Makefile
```diff
@@ -115,6 +115,13 @@ endif
 ifeq ($(UNAME_M),amd64)
 	CFLAGS += -mavx -mavx2 -mfma -mf16c
 endif
+ifeq ($(UNAME_M),ppc64)
+	POWER9_M := $(shell grep "POWER9" /proc/cpuinfo)
+	ifneq (,$(findstring POWER9,$(POWER9_M)))
+		CFLAGS += -mpower9-vector -DGGML_BIG_ENDIAN
+		CXXFLAGS += -std=c++23 -DGGML_BIG_ENDIAN
+	endif
```
awilfox, line 123: It seems that this port will be big-endian even if run on non-Power9 systems (consider P8, P10, or any other generation), so I think this is more correct:

```suggestion
	ifneq (,$(findstring POWER9,$(POWER9_M)))
		CFLAGS += -mpower9-vector
	endif
	CFLAGS += -DGGML_BIG_ENDIAN
	CXXFLAGS += -std=c++23 -DGGML_BIG_ENDIAN
```

ggml.c
```diff
@@ -6772,6 +6781,31 @@ static void ggml_visit_parents(struct ggml_cgraph * cgraph, struct ggml_tensor *
...
     }
 
+#if defined(GGML_BIG_ENDIAN)
+    if (node->src1) {
+        switch (node->src1->type) {
+            case GGML_TYPE_I16:
+            case GGML_TYPE_F16: {
+                unsigned short * toswap = (unsigned short *)(node->src1->data);
+                for (int k = 0; k < (ggml_nbytes(node->src1) / sizeof(unsigned short)); k++) {
+                    toswap[k] = bswap_16(toswap[k]);
```
awilfox, line 6791: Any reason for using `std::byteswap` in `whisper.cpp` but not here?

whisper.cpp
```diff
@@ -439,6 +442,11 @@ struct whisper_context {
 template<typename T>
 static void read_safe(std::ifstream& fin, T& dest) {
     fin.read((char*)& dest, sizeof(T));
+#if defined(GGML_BIG_ENDIAN)
+    if constexpr (std::endian::native == std::endian::big) {
```
awilfox, line 446: I realise the reason for the guard is because `<bit>` is guarded by it, but this seems wrong; if that's defined, it will always be BE.

What you probably want is something like `#if __cplusplus >= 202110L` (the revision of C++2x that added `std::byteswap`).

awilfox (commented):
Few nits I found while reviewing the changes.

fitzsim:
This is ready to merge now.  @awilfox, all your comments on the draft are addressed, thanks.

Maybe `convert-pt-to-ggml.py` should hard-code struct-packing to use little-endian, so that when it is run on big-endian targets the resulting model will still be compatible with this code.  Calls like `struct.pack("i", 0x67676d6c)` would be changed to `struct.pack("<i", 0x67676d6c)` (untested).  That said, on PPC64 `import torch` doesn't even work, so the models can't currently be generated on that platform anyway.

PR #415, opened 2023-01-16
Use AssetManager for Android

Digipom:
Using AssetManager directly from C is more direct/idiomatic.


Digipom:
Nice work on adding the callback API! :) I left the InputStream code in place since it could also be useful for loading from other paths than the assets, however in many cases we can also just open a file descriptor and pass that to the C code, then native code can be used instead of using JNI to access the input stream.

ggerganov:
> Nice work on adding the callback API! :)

Credits to @prsyahmi for the implementation!

PR #443, opened 2023-01-24
addon: implement node addon call whisper through cpp

chenqianhe:
I tried to call Whisper in the node environment through addon. I hope I can contribute it to examples for reference.

The speed difference between calling in node environment and directly using cpp for reasoning is very small.

cpp(./main -m ggml-large.bin -l zh -f samples_jfk.wav):
```
whisper_init_from_file: loading model from 'ggml-large.bin'
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51865
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 1280
whisper_model_load: n_audio_head  = 20
whisper_model_load: n_audio_layer = 32
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 1280
whisper_model_load: n_text_head   = 20
whisper_model_load: n_text_layer  = 32
whisper_model_load: n_mels        = 80
whisper_model_load: f16           = 1
whisper_model_load: type          = 5
whisper_model_load: mem required  = 4641.00 MB (+   71.00 MB per decoder)
whisper_model_load: kv self size  =   70.00 MB
whisper_model_load: kv cross size =  234.38 MB
whisper_model_load: adding 1608 extra tokens
whisper_model_load: model ctx     = 2950.97 MB
whisper_model_load: model size    = 2950.66 MB

system_info: n_threads = 4 / 8 | AVX = 0 | AVX2 = 0 | AVX512 = 0 | FMA = 0 | NEON = 1 | ARM_FMA = 1 | F16C = 0 | FP16_VA = 1 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 0 | VSX = 0 | 

main: processing 'samples_jfk.wav' (176000 samples, 11.0 sec), 4 threads, 1 processors, lang = zh, task = transcribe, timestamps = 1 ...


[00:00:00.000 --> 00:00:11.000]  And so my fellow Americans, ask not what your country can do for you, ask what you can do for your country.


whisper_print_timings:     fallbacks =   0 p /   0 h
whisper_print_timings:     load time =  1884.19 ms
whisper_print_timings:      mel time =    16.54 ms
whisper_print_timings:   sample time =    11.85 ms /    27 runs (    0.44 ms per run)
whisper_print_timings:   encode time =  6082.55 ms /     1 runs ( 6082.55 ms per run)
whisper_print_timings:   decode time =  1759.77 ms /    27 runs (   65.18 ms per run)
[...]

PR #468, opened 2023-02-04
addon: add test ci and optimize compilation configuration

chenqianhe:
1. add test for #443 
2. avoid introducing test compilation when compiling addon
3. Improve and verify the Windows platform compilation addon


.github/workflows/examles.yml
```diff
@@ -0,0 +1,48 @@
+name: Examples Tests
```
mightymatth, line 1: file name typo. `examles.yml` to `examples.yml`

chenqianhe: thanks

examples/addon.node/__test__/whisper.spec.js
```diff
@@ -0,0 +1,15 @@
+const path = require('path');
+const { whisper } = require(path.join(__dirname, '../../../build/Release/whisper-addon'));
+
+const whisperParamsMock = {
+    language: 'en',
+    model: path.join(__dirname, '../../../models/ggml-base.en.bin'),
+    fname_inp: path.join(__dirname, './samples_jfk.wav'),
```
ggerganov, line 7: Maybe better to reuse the `jfk.wav` from the `samples` folder, instead of having a copy here:

```suggestion
    fname_inp: path.join(__dirname, '../../../samples/jfk.wav'),
```

ggerganov (changes requested):

PR #473, opened 2023-02-05
Suppress non-speech-related token outputs

shibukazu:
## Problem
- Speech unrelated outputs like "(笑)", "[Bell]" often emerge.

## Change
- Suppressed the output of non-speech-related tokens based on the OpenAI's implement. 
  (https://github.com/openai/whisper/blob/7858aa9c08d98f75575035ecd6481f462d66ca27/whisper/tokenizer.py#L224-L253)

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -3133,6 +3133,35 @@ static void whisper_process_logits(
         logits[vocab.token_translate]  = -INFINITY;
         logits[vocab.token_transcribe] = -INFINITY;
 
+
+        // suppress non-speech tokens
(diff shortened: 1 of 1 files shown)


RndyP:
I was suggesting this also. Good idea. However, shouldn't this be controlled by a bool in whisper_full_params ?

shibukazu:
Your point is right, I will add the parameter to be able to control suppression.

shibukazu:
I noticed that "non-speech-tokens suppression" make the decoding speed worse in the case input audio file doesn't have any voice (only silent audio).
I think this is because "non-speech-tokens suppression" makes it difficult to express silent state (e.g. [SILENT], [無音] )
Does anyone have good idea?

PR #474, opened 2023-02-05
Fixed Beam Search Strategy and exposed whisper_pcm_to_mel_phase_vocoder

sandrohanea:
This commit addresses multiple issues.

1. whisper_full_default_params is not returning the strategy defined in the parameter but the parameters are loaded for that one on line 2966.

2. Encountered some index out of range during beam search on beam_candidates[cur_c] when all previous decoders have the same sum_logprobs_all. Fixed it with an additional check.

3. whisper_pcm_to_mel_phase_vocoder is not exposed so it cannot be used to speed up auto-language detection in: https://github.com/sandrohanea/whisper.net/pull/1

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -2905,7 +2905,7 @@ const char * whisper_print_system_info(void) {
 
 struct whisper_full_params whisper_full_default_params(enum whisper_sampling_strategy strategy) {
     struct whisper_full_params result = {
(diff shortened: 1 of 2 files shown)


ggerganov:
Thanks for the fixes!
Pt. 2 needs to be fixes in a better way - I incorrectly assumed that the beam candidates are guaranteed to have different `sum_logprobs_all`. Will make a note to resolve this later

PR #482, opened 2023-02-08
Refactor examples code to reuse some bits

ggerganov:
We now have a `common` and `common-sdl` helper libs that provide common functionality across the examples.
This way we avoid a lot of copy-pasted code.


polvoazul:
This is a nice idea. Without refactoring its very hard to change stuff (add more params) and keep in sync with main repo, as merge conflicts are all over the place.

Issue #470, opened 2023-02-05
armv7l crashes, out of box Illegal instruction, with tweaking Killed


clach04:
I have a 32-bit arm7 CPU on an Odroid-XU4 with Armbian 21.08.6 Focal with Linux 5.4.160-odroidxu4

Out of box make crashes:

    ./main -m models/ggml-medium.bin  -otxt samples/jfk.wav
    ....
    Illegal instruction

Looking at the makefile, looks like it's using the wrong FPU lib for floating point math.

https://github.com/ggerganov/whisper.cpp/blob/ab1916fc598cc364b521a6d24752c4b092553e40/Makefile#L149

Introduced in https://github.com/ggerganov/whisper.cpp/commit/167324584b0927fa78d696743d29f0ff29bebfe9 via https://github.com/ggerganov/whisper.cpp/pull/23

Tweaked which but then crashes out with Killed instead:

    diff --git a/Makefile b/Makefile
    index 20915e3..f20bb5d 100644
    --- a/Makefile
    +++ b/Makefile
    @@ -147,8 +147,11 @@ ifneq ($(filter armv6%,$(UNAME_M)),)
            CFLAGS += -mfpu=neon-fp-armv8 -mfp16-format=ieee -mno-unaligned-access
     endif
     ifneq ($(filter armv7%,$(UNAME_M)),)
    +       # this label looks wrong - matches 71 which is ONLY 32-bit
            # Raspberry Pi 4
    -       CFLAGS += -mfpu=neon-fp-armv8 -mfp16-format=ieee -mno-unaligned-access -funsafe-math-optimizations
    +       #CFLAGS += -mfpu=neon-fp-armv8 -mfp16-format=ieee -mno-unaligned-access -funsafe-math-optimizations
    +       #  cc: note: valid arguments to ‘-mfpu=’ are: auto crypto-neon-fp-armv8 fp-armv8 fpv4-sp-d16 fpv5-d16 fpv5-sp-d16 neon neon-fp-armv8 neon-fp16 neon-vfpv3 neon-vfpv4 vfp vfp3 vfpv2 vfpv3 vfpv3-d16 vfpv3-d16-fp16 vfpv3-fp16 vfpv3xd vfpv3xd-fp16 vfpv4 vfpv4-d16; did you mean ‘neon-fp-armv8’?
    +       CFLAGS += -mfpu=neon -mfp16-format=ieee -mno-unaligned-access -funsafe-math-optimizations
     endif
[...]

PR #486, opened 2023-02-09
issue #470 - working 32-bit ARM

clach04:


clach04:
Fixes `Illegal instruction` at runtime.

NOTE not tested on Raspberry. Only on Armbian with OdroixXU4.

Tested with neon fpu flag. Also had success with:

  * `-mfpu=neon-fp16`
  * `-mfpu=neon-vfpv3 `

clach04:
> We need a way to detect 32-bit ARM in the Makefile and choose the respective flags

That's what this PR does, the original code did not do this. Can you clarify in case I'm missing something?

clach04:
> My Raspberry Pi 4 reports `armv7l`, but I can use `-mfpu=neon-fp-armv8` successfully. It has much better performance compared to the proposed `-mfpu=neon`.

Ahhh, sounds like you have a 64-bit board but a 32-bit Operating System? Can you confirm?


> Does `-mfpu=neon-fp-armv8` not work for you?

No it does not, I do NOT have a 64-bit Pi 4, I have an armv7 board with no v8 extensions (see https://forums.raspberrypi.com/viewtopic.php?t=282791 for Pi4 self reporting, it reports as v7 in a 32-bit OS but allows access to v8 instructions). My board is ONLY 32-bit.

New idea/proposal (assuming you have a really have pi4 board with 32-bit OS), in the 32-bit OS (v7) code block we add an additional conditional for detecting a raspberry pi board? Thoughts?

Slightly off topic, you might find moving to a 64-bit OS gains you some extra performance else where for your board, then the makefile would drop into the next "if" check.

ggerganov:
Yes, it looks like I am using a 32-bit OS on my 64-bit RPi4. Thanks for clearing this up.
It would be very useful to have a branch in the Makefile for my use case, since I don't see mysql upgrading the OS anytime soon. But if there is no clean way to do it, we can merge it like this and I will simply modify the Makefile locally when working on my RPi4

clach04:
What do you get for:

    cat /proc/cpuinfo |grep 'model name'

(`| head -1`) ?

If you get `ARMv7 Processor rev 4 (v7l)` (4 or above) that might be the (slightly messy) heuristic that could work 🤞 .

I don't have a Pi4 to check with 😿 This might also work for [Pi 3b+](https://thepihut.com/products/raspberry-pi-3-model-b-plus) boards too.

clach04:
Hmmm, that completely surprised me. Sadly, I think your idea of a private change is the safest :-( I was really hoping we could automate this nicely.

Alternative idea, mention pi 4 can use v8 extensions in a comment (e.g. the ones that worked for you) so its documented as a potential option?

Makefile
```diff
@@ -143,12 +143,12 @@ endif
...
-	# Raspberry Pi 1, 2, 3
-	CFLAGS += -mfpu=neon-fp-armv8 -mfp16-format=ieee -mno-unaligned-access
+	# 32-bit Raspberry Pi 1, 2, 3
+	CFLAGS += -mfpu=neon -mfp16-format=ieee -mno-unaligned-access
 endif
 ifneq ($(filter armv7%,$(UNAME_M)),)
-	# Raspberry Pi 4
-	CFLAGS += -mfpu=neon-fp-armv8 -mfp16-format=ieee -mno-unaligned-access -funsafe-math-optimizations
+	# 32-bit ARM, for example on Armbian or possibly raspbian
+	CFLAGS += -mfpu=neon -mfp16-format=ieee -mno-unaligned-access -funsafe-math-optimizations  # works with small
```
ggerganov, line 151:
```suggestion
	# 32-bit ARM, for example on Armbian or possibly raspbian
	CFLAGS += -mfpu=neon -mfp16-format=ieee -mno-unaligned-access -funsafe-math-optimizations
	
	# alternatively, on some RPi4, it can be better to use the following flags:
	# CFLAGS += -mfpu=neon-fp-armv8 -mfp16-format=ieee -mno-unaligned-access -funsafe-math-optimizations
```

Makefile
```diff
@@ -143,12 +143,15 @@ endif
...
+	# 32-bit Raspberry Pi 1, 2, 3
+	CFLAGS += -mfpu=neon -mfp16-format=ieee -mno-unaligned-access
 endif
 ifneq ($(filter armv7%,$(UNAME_M)),)
-	# Raspberry Pi 4
-	CFLAGS += -mfpu=neon-fp-armv8 -mfp16-format=ieee -mno-unaligned-access -funsafe-math-optimizations
+	# 32-bit ARM, for example on Armbian or possibly raspbian
+	CFLAGS += -mfpu=neon -mfp16-format=ieee -mno-unaligned-access -funsafe-math-optimizations
+	
+	# alternatively, on some RPi4, it can be better to use the following flags:
```
ggerganov, line 153:
```suggestion
	# 64-bit ARM, use these (TODO: auto-detect 64-bit)
```

Issue #475, opened 2023-02-05
Separation of transformation state and model context


sandrohanea:
Currently, in order to run multiple transformations in parallel, you'll need to instantiate multiple whisper_contexts because the state of the transformation (e.g mel spectogram, previous prompts, prtial results) is stored on the same context where the model, vocabulary, etc. are stored.

Is there something which I'm missing why this cannot be separated (to have the context and the state as different entities).

The context will be immutable in this case and can be used by multiple transformations without extra allocation.

What do you think @ggerganov ?

ggerganov:
Yes - you are correct. It seems I didn't design the `whisper_context` in the best possible way.

For now, you can achieve parallel processing using a single `whisper_context` by hacking `whisper.cpp` directly and doing something similar to what has been done in `whisper_full_parallel()`:

https://github.com/ggerganov/whisper.cpp/blob/2b85be14d8f3e8887623ed76b26b27dfccad7916/whisper.cpp#L4207

But it would be much better if the C API allowed to do this straight from the user code.
Will think about adding support for this in the future.

sandrohanea:
I didn't know about this whisper_full_parallel()

Thanks for pointing it out, will take a look in the meantime :)

Not sure if I can do the enhancement (it's been a while since I coded a little more in C++), but I will try this weekend (to separate the state and context).

PR #494, opened 2023-02-12
Separated context and state for easier parallelization.

sandrohanea:
https://github.com/ggerganov/whisper.cpp/issues/475

- Separated and tested the state of the transformation as a different struct.
- Fixed examples (stream, command, talk, main) => because these are not using the callbacks when parsing the results, they need to call `whisper_full_with_state` instead of `whisper_full` so, they can read the results from the state at the end of transformation.
- Fix rest of examples
- Fix Bindings + examples of bindings

Note:
In order for the bindings to be able to use same context with multiple different transformations (with a state for each transformation) in parallel, some adjustments are needed, but the default cases are already working.


whisper.cpp
```diff
@@ -540,37 +540,29 @@ struct whisper_decoder {
     bool has_ts;    // have we already sampled a non-beg timestamp token for the current segment?
 
     // new token probs, logits and logprobs after the last whisper_decode (1-dimensional array: [n_vocab])
-    std::vector<float> probs;
-    std::vector<float> logits;
-    std::vector<float> logprobs;
+    std::vector<float> probs{};
```
sandrohanea, line 543: Fought with some read access denied exception for almost 2h, only to find out that it's a bug in compiler (only for debug): https://github.com/microsoft/STL/issues/1934

The workaround is just to initialize the struct (otherwise, it's pretty hard to debug)

examples/bench.wasm/emscripten.cpp
```diff
@@ -10,20 +10,20 @@
 
 constexpr int N_THREAD = 8;
 
-// TODO: get rid of this vector of contexts - bad idea in the first place
```
sandrohanea, line 13: Fixed all these todos and keps only 1 context needed + a vector of states.

RndyP:
Been following your work here. Not sure if this is related, but I profile Whisper under Visual C++ profiling tool, and there is this section of code that is using a spinlock that is taking 10% of CPU time. Shouldn't this be using a mutex?
image

sandrohanea:
Hello @RndyP ,
Indeed, it looks like that code can be optimized, but it's not in the scope of this PR. This PR is meant to split the whisper_context into whisper_context + whisper_state so whisper_context can be reused and thread safe.

About that spinlock, I agree it would be nice to change it. Can you, maybe create another issue about your findings?

RndyP:
I reported this in issue #300, which has fallen down the list. I don't understand the code enough to suggest a fix myself.

ggerganov:
@sandrohanea 
Thanks for this nice work! I will need some time to think about this change before merging. I kind of want to keep the existing API the same if possible and add extra functions for your use case, but not sure if this is better compared to what you have proposed here.

Don't worry about keeping up-to-date - I will be able to do that if necessary.

sandrohanea:
Thanks a lot for taking the time, it totally makes sense.

Also, thanks again for creating the whole library in the first place, really titanic work to port all tensor operations and everything.

I was thinking also about keeping a "default state" in the context and use that if no different state is provided, this way it won't be a breaking change to existing functionality. 

On the other hand, it is really easy to use it wrong if context is thread safe "sometimes".

It's your call on this one. If you have any idea how this could be done better, please, let me know and I'll be happy to help.

sandrohanea:
Closing this as the alternative with opt in state was merged.

PR #500, opened 2023-02-14
Ruby Bindings

taf2:
I did some work to build a ruby binding with the following interface:

```ruby
require 'whisper'
params = Whisper::Params.new
whisper = Whisper::Context.new(File.join(File.dirname(__FILE__), 'models', 'ggml-base.en.bin'))

input = ARGV[0]
if input.nil?
  STDERR.puts "usage: t.rb your.wav"
  exit 1
end

whisper.transcribe(input, params) {|text|
  puts text.inspect
}
```

There are no doubt things that could be better and I was not sure how you felt bindings like this should integrate into the rest of the build system.  The main thing I was not sure about is I copy when running extconf.rb the source file dependencies needed e.g. whisper.cpp etc...


ggerganov:
That's great!
If you can provide a [Github action](https://github.com/ggerganov/whisper.cpp/tree/master/.github/workflows) for building/running the bindings I'll happily merge this.

taf2:
@ggerganov thanks, I think I added this into the right place in my most recent push - but I am also now thinking it'd be great to have a ruby gem for people to install easily via ``` gem install whisper ```

One issue with how I am doing a cp to move the whisper.cpp files in extconf.rb is that if we package this into a rubygem those files will need to be in the ruby gem already... I'll work on a spec file now if you think it'd be hepful and maybe overcome this issue.

ggerganov:
@taf2 I added a separate Ruby action based on your commit.

I'm not familiar with the ruby ecosystem. Whatever you think would be useful for other people is OK.
Let me know if this PR is OK to merge, or if you prefer to add the gem stuff before merging.

Regarding the copying of the sources - I don't fully understand why it is necessary. Would be useful to avoid it if possible.

taf2:
Agreed - yes I think this is a good starting place for ruby to be merged.  I'm going to spend some more time today and hopefully not have to copy those files and maybe recruit some help too for getting it published to rubygems which i think would be ideal.

Issue #445, opened 2023-01-24
Language auto-detect "auto" flag does not work using the stream tool


bdrelling:
Attempting to use the `--language auto` option to enable auto-detect with the multi-language model throws an error for the `stream` tool.

**Command**

```sh
./stream -m ./models/ggml-base.bin --threads 8 --step 500 --length 5000 --language auto
```

**Error**

```
whisper_lang_id: unknown language 'auto'
error: unknown language 'auto'
```

gourcetools:
Same here, how to make this workproperly?
```
whisper_lang_id: unknown language 'auto'
error: unknown language 'auto'

usage: ./stream [options]

```

AeroXi:
same

pannous:
you can hack one line in stream.cpp as in https://github.com/ggerganov/whisper.cpp/pull/501#pullrequestreview-1376780021

gourcetools:
> you can hack one line in stream.cpp as in [#501 (review)](https://github.com/ggerganov/whisper.cpp/pull/501#pullrequestreview-1376780021)

This works. Thanks

PR #501, opened 2023-02-14
Update stream.cpp

AliAlameh:
#445  fix Language auto-detect "auto" flag does not work using the stream tool

### DIFF
--- a/examples/stream/stream.cpp
+++ b/examples/stream/stream.cpp
@@ -451,7 +451,7 @@ int main(int argc, char ** argv) {
 
     // whisper init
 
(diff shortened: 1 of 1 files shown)


ggerganov:
This way, the stream example will keep on auto-detecting the language forever.

Maybe it would be better after it auto-detects the spoken language to select it explicitly in order to avoid running the "auto-detect" code

pannous (approved):
Fantastic pull request exactly what I was looking for why isn't it merged yet?

By the way the line that needs fixing has moved to line 150

gourcetools:
Thank you, please merge.

ggerganov (approved):

PR #534, opened 2023-02-25
adding build files for rubygem whispercpp

taf2:
This add a Rakefile and gemspec file to build whispercpp gem.  This allows endusers in the ruby ecosystem just type:

```
gem install whispercpp
```

### DIFF
--- a/whisper.h
+++ b/whisper.h
@@ -225,6 +225,15 @@ extern "C" {
     // Print system information
     WHISPER_API const char * whisper_print_system_info(void);
 
+    // Abort a running whisper_full_parallel or whisper_full
+    WHISPER_API void whisper_running_abort(struct whisper_context * ctx);
+
+    // Resume whisper context from an aborted state allowing it run again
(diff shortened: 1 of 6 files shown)


whisper.h
```diff
@@ -225,6 +225,15 @@ extern "C" {
...
 
+    // Abort a running whisper_full_parallel or whisper_full
+    WHISPER_API void whisper_running_abort(struct whisper_context * ctx);
+
+    // Resume whisper context from an aborted state allowing it run again
+    WHISPER_API void whisper_running_restore(struct whisper_context * ctx);
+
+    // Check the whisper context state if true then it can run if false it can not
+    WHISPER_API bool whisper_running_state(struct whisper_context * ctx);
+
```
ggerganov, line 236: See https://github.com/ggerganov/whisper.cpp/discussions/507#discussioncomment-5127320
I don't think this is needed

ggerganov (changes requested):

PR #537, opened 2023-02-25
Go binding: NewContext now returns a clean context

polarmoon:
It makes more sense to return a clean context upon `NewContext` call


polarmoon:
@ggerganov in what case the lib user wants to create a new Golang context but keep the past transcription text? Seems to better just use the existing Go context?

On the other hand, I believe it makes more sense that NewContext returns a truly new context, rather than a context remembers past transcription.

ggerganov (approved):

ggerganov:
@polarmoon 
I apologise - somehow I was confused about what `no_context` does.
You are correct - it is better to set it to `true` by default. I will also make it to be the default value in `whisper.cpp` as well.
Thanks 👍

PR #542, opened 2023-02-27
whisper.android: Support benchmark for Android example.

tinoue:
- Support to show system info, voice input length and time to transcribe.

Screenshot_2023-02-27-23-44-26-640_com whispercppdemo

- Add "benchmark" button which runs memcpy and ggml mul mat benchmarks.

Screenshot_2023-02-28-00-26-40-363_com whispercppdemo

### DIFF
--- a/whisper.h
+++ b/whisper.h
@@ -383,10 +383,10 @@ extern "C" {
 
     // Temporary helpers needed for exposing ggml interface
 
-    WHISPER_API int whisper_print_bench_memcpy(int n_threads);
-    WHISPER_API const char * whisper_bench_memcpy(int n_threads);
-    WHISPER_API int whisper_print_bench_ggml_mul_mat(int n_threads);
-    WHISPER_API const char * whisper_bench_ggml_mul_mat(int n_threads);
+    WHISPER_API int whisper_bench_memcpy(int n_threads);
+    WHISPER_API const char * whisper_bench_memcpy_str(int n_threads);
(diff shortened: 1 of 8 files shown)


whisper.h
```diff
@@ -383,8 +383,10 @@ extern "C" {
 
     // Temporary helpers needed for exposing ggml interface
 
-    WHISPER_API int whisper_bench_memcpy(int n_threads);
-    WHISPER_API int whisper_bench_ggml_mul_mat(int n_threads);
+    WHISPER_API int whisper_print_bench_memcpy(int n_threads);
+    WHISPER_API const char * whisper_bench_memcpy(int n_threads);
+    WHISPER_API int whisper_print_bench_ggml_mul_mat(int n_threads);
+    WHISPER_API const char * whisper_bench_ggml_mul_mat(int n_threads);
```
ggerganov, line 389: Please rename as follows:

```suggestion
    WHISPER_API int whisper_bench_memcpy(int n_threads);
    WHISPER_API const char * whisper_bench_memcpy_str(int n_threads);
    WHISPER_API int whisper_bench_ggml_mul_mat(int n_threads);
    WHISPER_API const char * whisper_bench_ggml_mul_mat_str(int n_threads);
```

Even though these are only temporary, I prefer to not change the existing API.

ggerganov (changes requested):
Thanks - this is very useful addition!

tinoue:
@ggerganov Thank you for your comment. Restored original API names.

PR #563, opened 2023-03-03
CMake support on Linux for absence of AVX2

pkeir:
This adds support within CMakeLists.txt for Linux builds on machines which don't support AVX2. Avoids "Illegal instruction (core dumped)" at runtime.


CMakeLists.txt
```diff
@@ -166,7 +166,8 @@ else()
             if(NOT WHISPER_NO_AVX)
                 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mavx")
             endif()
-            if(NOT WHISPER_NO_AVX2)
+            execute_process(COMMAND cat /proc/cpuinfo COMMAND grep avx2 OUTPUT_VARIABLE FOUND_AVX2)
```
ggerganov, line 169: I think this will fail on Intel Macs - there is no `/proc/cpuinfo`

ggerganov (changes requested):

PR #576, opened 2023-03-06
Fix MUSL Linux build

### LINKED ISSUE #37: MUSL Linux builds
Hi there! I'm attempting to build whisper.cpp for MUSL Linux for some lightweight systems, and I figured I would note the issues I ran into during the build.

1) Alpine appears to not include `stdint.h`, or `alloca.h` in its standard library when you install gcc. This results in a slew of errors:
```
localhost:~/whisper.cpp# make libwhisper.a
cc  -O3 -std=c11   -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -pthread   -c ggml.c
In file included from ggml.h:7,
                 from ggml.c:1:
/usr/lib/gcc/aarch64-alpine-linux-musl/12.2.1/include/stdint.h:9:26: error: no include path in which to search for stdint.h
    9 | # include_next <stdint.h>
      |                          ^
ggml.h:107:5: error: unknown type name 'int64_t'
  107 |     int64_t perf_cycles;
      |     ^~~~~~~
~~snip~~

ggml.c:6:10: fatal error: alloca.h: No such file or directory
    6 | #include <alloca.h>
      |          ^~~~~~~~~~
compilation terminated.
make: *** [Makefile:58: ggml.o] Error 1
localhost:~/whisper.cpp#
```

This fix is relatively simple, just install g++:
```
apk add g++
```

2) `clock_gettime` and `CLOCK_MONOTONIC` are seemingly undefined regardless of compiler used.
```
localhost:~/whisper.cpp# make libwhisper.a
cc  -O3 -std=c11   -Wall -Wextra -Wno-unused-parameter -Wno-unused-function -pthread   -c ggml.c
ggml.c: In function 'ggml_time_ms':
ggml.c:155:5: warning: implicit declaration of function 'clock_gettime' [-Wimplicit-function-declaration]
  155 |     clock_gettime(CLOCK_MONOTONIC, &ts);
      |     ^~~~~~~~~~~~~
ggml.c:155:19: error: 'CLOCK_MONOTONIC' undeclared (first use in this function)
  155 |     clock_gettime(CLOCK_MONOTONIC, &ts);
      |                   ^~~~~~~~~~~~~~~
ggml.c:155:19: note: each undeclared identifier is reported only once for each function it appears in
ggml.c: In function 'ggml_time_us':
ggml.c:161:19: error: 'CLOCK_MONOTONIC' undeclared (first use in this function)
  161 |     clock_gettime(CLOCK_MONOTONIC, &ts);
      |                   ^~~~~~~~~~~~~~~
make: *** [Makefile:58: ggml.o] Error 1
localhost:~/whisper.cpp# 
```
Digging around the internet shows [a fix](https://stackoverflow.com/questions/40515557/compilation-error-on-clock-gettime-and-clock-monotonic) for this as inserting `#define _POSIX_C_SOURCE 199309L` before including the time.h header. This appears to work successfully, placing it on line 10 of `ggml.c`. It would be nice if this issue could be fixed in some way. I would make a PR if I had sufficient knowledge to implement the required changes, which I don't.

ggerganov:
close #37 

@mikeslattery Could you please check if this works on MUSL Linux?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,12 @@ CFLAGS   = -I.              -O3 -DNDEBUG -std=c11   -fPIC
 CXXFLAGS = -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC
 LDFLAGS  =
 
+# ref: https://github.com/ggerganov/whisper.cpp/issues/37
+ifneq ($(wildcard /usr/include/musl/*),)
+	CFLAGS   += -D_POSIX_SOURCE -D_GNU_SOURCE
+	CXXFLAGS += -D_POSIX_SOURCE -D_GNU_SOURCE
+endif
+
 # OS specific
 # TODO: support Windows
 ifeq ($(UNAME_S),Linux)


mikeslattery:
It did not.  However, when I commented out the `ifneq` it did.  `$(CCV)` is the `CC` value below:

```
I whisper.cpp build info:
I UNAME_S:  Linux
I UNAME_P:  unknown
I UNAME_M:  x86_64
I CFLAGS:   -I.              -O3 -DNDEBUG -std=c11   -fPIC -pthread -mavx -mavx2 -mfma -mf16c -msse3
I CXXFLAGS: -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC -pthread
I LDFLAGS:
I CC:       cc (Alpine 12.2.1_git20220924-r4) 12.2.1 20220924
I CXX:      g++ (Alpine 12.2.1_git20220924-r4) 12.2.1 20220924
```

ggerganov:
Directly adding `-D_POSIX_SOURCE -D_GNU_SOURCE` to the flags is not desired since it breaks the build for other systems. We need a way to detect MUSL in the Makefile

mikeslattery:
> We need a way to detect MUSL in the Makefile

I asked ChatGPT

> List 3 ways within a makefile how you can detect if the system uses musl instead of libc


```makefile
ifeq ($(wildcard /usr/include/musl/*),)
        CFLAGS += -D_POSIX_SOURCE -D_GNU_SOURCE
        CXXFLAGS += -D_POSIX_SOURCE -D_GNU_SOURCE
endif

ifeq ($(shell ldd /bin/ls | grep -q musl && echo true),true)
endif

ifeq ($(shell pkg-config --exists musl && echo true),true)
endif
```

I tested the first and it worked.

ggerganov:
Added the suggestion by ChatGPT

PR #624, opened 2023-03-17
stream : output only completed lines of text in saved file.

dmahurin:
This change only outputs the completed lines to the text file, the lines that are later overwritten  are not output.  This allows stream to be better used for dictation.


ggerganov:
This breaks the "real-time" stream usage. For example see the videos here:

https://github.com/ggerganov/whisper.cpp/tree/master/examples/whisper.nvim

dmahurin:
The change has been amended to use std::flush, which allows the nvim operation.

An additional change is added to output the current incomplete line on interrupt.

Note, with these changes, the nvim configuration could also be updated to allow multiline entry (removing the tail -n 1 newline deletion)

PR #628, opened 2023-03-18
Add a cmake flag to disable F16C

a5huynh:
I'm running whisper in a virtualized environment for tests. The `Makefile` is able to build the binary just fine, but as a library I was running in an `illegal instruction` error.

Adding a flag to disable `-mf16c` did the trick.


j-f1 (approved):

thewh1teagle:
@ggerganov 
Can we use `-mf16c` conditionally in runtime only if supported?

Issue #637, opened 2023-03-21
Add `params.initial_prompt` to set an initial prompt string


sindresorhus:
It would be useful if the API had an `initial_prompt` parameter to set a string as the initial prompt. The OpenAI docs recommend always setting it to fix punctation, so it's quite a common need: https://platform.openai.com/docs/guides/speech-to-text/prompting

I'm aware I could already do something like [this](https://github.com/rock3125/whisper.cpp/commit/cd0ae21ac0436a595023074b86a8bfe7cc7f4797#diff-2d3599a9fad195f2c3c60bd06691bc1815325b3560b5feda41a91fa71194e805R468-R482), but I don't think it makes sense for everyone to re-implement that logic.

jhen0409:
Agree, that option will be very useful!

I did the same in https://github.com/mybigday/whisper.rn/pull/8, I think it can be built directly into the whisper.cpp. Maybe I will try to send a PR later.

User1231300:
I also would appreciate this

PR #645, opened 2023-03-23
whisper : add initial_prompt param

jhen0409:
Related to #637.

Add `initial_prompt` so we can avoid some re-implement of `prompt_tokens` and `n_prompt_tokens`.

This changes removes the print of prompt tokens because it may not ideal in whisper_full.


jhen0409:
It seems the `Example Tests` CI was broken after #642.

PR #650, opened 2023-03-24
Adds example that does streamming text via ZeroMQ in pub/sub model

whitead:
This adds a new `streammq` example that streams text over ZeroMQ in a pub/sub model. This enables other tools to access what is being dictated from a whisper model running.


fire:
Now I'm interested to see if I can do OSC based streaming.

PR #668, opened 2023-03-28
Add Alpaca support to talk-llama example

ejones:
I made a couple edits locally to get `talk-llama` working with alpaca-13b, and thought I'd contribute back if that's helpful.

The invocation I used is:
```
./talk-llama -mw ./models/ggml-base.en.bin -ml path/to/ggml-alpaca-13b-q4.bin -p Evan -t 8 --n-parts-llama 1 --verbose-prompt --prompt-file examples/talk-llama/prompts/talk-alpaca.txt
```

This basically boils down to making the prompt and n_parts configurable in the CLI as well as adding a sample prompt in the Alpaca format. Seems to work well with alpaca-13b; I don't have any other models to test it with. Also I admit that making the prompt configurable here exposes the `{N}`-style interpolation that `talk-llama.cpp` does, which may warrant more thought.

### DIFF
--- a/examples/talk-llama/prompts/talk-alpaca.txt
+++ b/examples/talk-llama/prompts/talk-alpaca.txt
@@ -1,0 +1,23 @@
+Below is an instruction that describes a task. Write a response that appropriately completes the request.
+
+### Instruction:
+
+Write a text transcript of a never ending dialog, where {0} interacts with an AI assistant named {1}.
+{1} is helpful, kind, honest, friendly, good at writing and never fails to answer {0}’s requests immediately and with details and precision.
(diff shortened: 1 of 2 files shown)


ggerganov (approved):

JKeddo95:
Have you seen the `llama_model_load: invalid model file '../llama.cpp/models/Alpaca/13B/ggml-alpaca-13B-q4.bin' (bad magic)` message? I am able to run fine in llama.cpp, but can't get it started from whisper. Can you confirm you are running newest Alpaca.cpp? Which converter (if any) did you use for your files?

jmtatsch:
I would very much like to talk to my vicuna but am suffering from the same bad magic error.
Is the ggml format maybe too new for the llama.cpp contained in whisper.cpp?

PR #682, opened 2023-03-29
Update README.md

razodactyl:
- Slight tweaks to README for improved comprehension.


razodactyl:
Hmm. Is it required to edit `Models/WhisperState` as there's a hardcoded URL for the specific model file.

PR #704, opened 2023-03-31
ggml : use vzip_s8 for interleave

jhen0409:
Fix 32-bit build in whisper.android.

I got the same issue from https://github.com/ggerganov/llama.cpp/issues/482 after #664, it can be fixed with [`vzip_s8`](https://developer.arm.com/architectures/instruction-sets/intrinsics/vzip_s8) which also supports A32.


ggerganov:
With the latest version, do you still have this issue?

jhen0409:
Confirmed it fixed by https://github.com/ggerganov/whisper.cpp/commit/2f889132c66051b14c6f8770e9b3d4e3f159821d. Thanks! Also I got an another build issue on 64-bit, may file an issue or PR later.

Issue #199, opened 2022-11-28
Make failure - error: inlining failed in call to always_inline 


JJGO:
Thanks for releasing this software, I'm trying to build this in a Debian system but make is failing. The Compiler versions are
```
❯ cc --version
cc (Debian 8.3.0-6) 8.3.0

❯ c++ --version
c++ (Debian 8.3.0-6) 8.3.0
```
When trying `make base.en` in the `master` branch the following error happens

```
❯ make base.en
cc  -I.              -O3 -std=c11   -pthread -mavx   -c ggml.c -o ggml.o
In file included from /usr/lib/gcc/x86_64-linux-gnu/8/include/immintrin.h:99,
                 from ggml.c:128:
ggml.c: In function ‘ggml_vec_mad_f16’:
/usr/lib/gcc/x86_64-linux-gnu/8/include/f16cintrin.h:73:1: error: inlining failed in call to always_inline ‘_mm256_cvtps_ph’: target specific option mismatch
 _mm256_cvtps_ph (__m256 __A, const int __I)
 ^~~~~~~~~~~~~~~
ggml.c:1024:9: note: called from here
         _mm_storeu_si128((__m128i*)(y + i + 24), _mm256_cvtps_ph(y3, 0));
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/lib/gcc/x86_64-linux-gnu/8/include/immintrin.h:99,
                 from ggml.c:128:
/usr/lib/gcc/x86_64-linux-gnu/8/include/f16cintrin.h:73:1: error: inlining failed in call to always_inline ‘_mm256_cvtps_ph’: target specific option mismatch
 _mm256_cvtps_ph (__m256 __A, const int __I)
 ^~~~~~~~~~~~~~~
ggml.c:1023:9: note: called from here
         _mm_storeu_si128((__m128i*)(y + i + 16), _mm256_cvtps_ph(y2, 0));
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/lib/gcc/x86_64-linux-gnu/8/include/immintrin.h:99,
                 from ggml.c:128:
[...]

PR #706, opened 2023-04-01
Makefile: disable avx in case f16c is not available

duthils:
Why:

* ggml.c does not support AVX without F16C

Fixes https://github.com/ggerganov/whisper.cpp/issues/199


ggerganov:
Hmm, we should support it. What error do you get?

ggerganov (approved):

Issue #724, opened 2023-04-06
conversion script pt to ggml not working


ivan-gorin:
Whisper seem to have switched to a different tokenizer library: [pull request](https://github.com/openai/whisper/pull/1044), and the conversion script [convert-pt-to-ggml.py](https://github.com/ggerganov/whisper.cpp/blob/0a2d1210bcb98978214bbf4e100922a413afd39d/models/convert-pt-to-ggml.py) has stopped working. I think it needs to be updated to use the new .tiktoken files.

PR #725, opened 2023-04-06
Changed convert-pt-to-ggml.py to use .tiktoken tokenizer files

ivan-gorin:
Fixes my issue #724 
Changed the way vocabulary is read and converted to ggml. Used the same [code](https://github.com/openai/tiktoken/blob/46287bfa493f8ccca4d927386d7ea9cc20487525/tiktoken/load.py#L115-L118) as in original tiktoken library.

The resulting ggml files are the same as ones downloaded from huggingface. For multilingual exactly the same, for en only there is a difference of 17 bytes, due to the difference in whisper's vocab files. 
```
-rw-rw-r--  1 ivan ivan 77691713 Mar 22 11:09 ggml-tiny.bin
-rw-rw-r--  1 ivan ivan 77691713 Apr  6 06:13 ggml-tiny.bin.new
-rw-rw-r--  1 ivan ivan 77704715 Mar 22 11:08 ggml-tiny.en.bin
-rw-rw-r--  1 ivan ivan 77704698 Apr  6 06:14 ggml-tiny.en.bin.new
```
In the old [vocab.json](https://github.com/openai/whisper/blob/ad3250a846fe7553a25064a2dc593e492dadf040/whisper/assets/gpt2/vocab.json) there are 50257 tokens, the last one `<|endoftext|>` with index 50256. In the new [gpt2.tiktoken](https://github.com/openai/whisper/blob/b5851c6c40e753606765ac45b85b298e3ae9e00d/whisper/assets/gpt2.tiktoken) there are only 50256 tokens, the endoftext is removed. 4 bytes for the int storing string length + 13 bytes for the string itself = 17 byte difference. It doesn't seem to make a difference in the model output anyway, not sure why this token was in the vocab.json previously, since it is probably a special token.

I have only tested with tiny models, but the only change is in the tokenizer so it should work for all the others.

### DIFF
--- a/models/convert-pt-to-ggml.py
+++ b/models/convert-pt-to-ggml.py
@@ -39,6 +39,7 @@ import json
 import code
 import torch
 import numpy as np
+import base64
 
 #from transformers import GPTJForCausalLM
 #from transformers import GPT2TokenizerFast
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
Tested with `medium` and it produces the same model as before.
Thank you very much!

PR #768, opened 2023-04-15
C++11style

wuyudi:
tested by `./tests/run-tests.sh tiny.en`

- use STL to rewrite some for-loops.
- use `Path` to avoid catenating string path.
- `self.field = v` instead of `setattr(self, "field", v)`

about designated initializers, it seems to work on gcc, clang, and msvc. https://developercommunity.visualstudio.com/t/c-designated-initializers/518768

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -2902,9 +2902,9 @@ int whisper_lang_max_id() {
     return max_id;
 }
 
-int whisper_lang_id(const char *lang) {
+int whisper_lang_id(const char * lang) {
     if (!g_lang.count(lang)) {
-        for (const auto &kv : g_lang) {
+        for (const auto & kv : g_lang) {
             if (kv.second.second == lang) {
                 return kv.second.first;
             }
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -2901,7 +2901,7 @@ int whisper_lang_id(const char *lang) {
     if (!g_lang.count(lang)) {
         auto fd = std::find_if(g_lang.cbegin(), g_lang.cend(),
                                [=](it x) { return x.second.second == lang; });
-        if (fd == g_lang.end()) {
+        if (fd == g_lang.cend()) {
             fprintf(stderr, "%s: unknown language '%s'\n", __func__, lang);
             return -1;
         }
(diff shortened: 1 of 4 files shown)


whisper.cpp
```diff
@@ -2890,24 +2889,23 @@ int whisper_tokenize(struct whisper_context * ctx, const char * text, whisper_to
...
-    for (const auto & kv : g_lang) {
-        max_id = std::max(max_id, kv.second.first);
-    }
-
-    return max_id;
+    using it = decltype(*g_lang.cbegin());
+    return std::max_element(
+               g_lang.cbegin(), g_lang.cend(),
+               [](it x, it y) { return x.second.first > y.second.first; })
+        ->second.first;
```
ggerganov, line 2896: All changes starting from here to the end are not necessary - it just makes harder to read the code

whisper.cpp
```diff
@@ -2434,23 +2432,16 @@ static bool log_mel_spectrogram(
...
-        if (mel.data[i] < mmax) {
-            mel.data[i] = mmax;
-        }
-
-        mel.data[i] = (mel.data[i] + 4.0)/4.0;
-    }
+    std::transform(
+        mel.data.cbegin(), mel.data.cbegin() + (mel.n_mel * mel.n_len),
+        mel.data.begin(),
+        [=](double datai) { return (std::max(datai, mmax) + 4.0) / 4.0; });
```
ggerganov, line 2444: Keep the original

whisper.cpp
```diff
@@ -2434,23 +2432,16 @@ static bool log_mel_spectrogram(
...
 
     // clamping and normalization
-    double mmax = -1e20;
-    for (int i = 0; i < mel.n_mel*mel.n_len; i++) {
-        if (mel.data[i] > mmax) {
-            mmax = mel.data[i];
-        }
-    }
+    double mmax = *std::max_element(
+        mel.data.cbegin(), mel.data.cbegin() + (mel.n_mel * mel.n_len));
```
ggerganov, line 2436: Keep the original

whisper.cpp
```diff
@@ -1,3 +1,5 @@
+#include <iterator>
+#include <numeric>
```
ggerganov, line 2: I think these are not needed

ggerganov (changes requested):
The python changes are ok
The C++ changes are mostly not needed - see comments

ggerganov (approved):

Issue #680, opened 2023-03-28
UI: speed_up doubles the duration_ms


heshpdx:
This is an easy one to fix for anyone wanting to contribute. (I can get to it later if nobody else does).

If one supplies `-su` as a parameter alongside `-d`, then whatever value is passed to `-d` is effectively doubled. So if you specify `-d 10000 -su` to process the first ten seconds of audio at twice the rate, whisper.cpp will not exit until it completes twenty seconds of transcription.

PR #810, opened 2023-04-21
Allow duration knob to work correctly with speed_up knob

heshpdx:
Makes sure that inference stops at the correct timestamp.  Closes issue #680 
Also fixes build for gcc-13.

```
./main -m models/ggml-base.en.bin -su -d 11000 -f samples/jfk.wav
```
Original output:
```
main: processing 'samples/jfk.wav' (176000 samples, 11.0 sec), 4 threads, 1 processors, lang = en, task = transcribe, timestamps = 1 ...

[00:00:00.000 --> 00:00:10.560]   And so my fellow Americans, ask not what your country can do for you, ask what you can do for your country.
[00:00:10.560 --> 00:00:30.560]   [BLANK_AUDIO]
```

Fixed:
```
main: processing 'samples/jfk.wav' (176000 samples, 11.0 sec), 4 threads, 1 processors, lang = en, task = transcribe, timestamps = 1 ...

[00:00:00.000 --> 00:00:10.560]   And so my fellow Americans, ask not what your country can do for you, ask what you can do for your country.
```

### DIFF
--- a/ggml/src/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan.cpp
@@ -2042,9 +2042,9 @@ void ggml_vk_instance_init() {
 }
 
 static void ggml_vk_init(ggml_backend_vk_context * ctx, size_t idx) {
-    GGML_ASSERT(idx < vk_instance.device_indices.size());
     VK_LOG_DEBUG("ggml_vk_init(" << ctx->name << ", " << idx << ")");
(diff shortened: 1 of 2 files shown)


ggerganov (commented):
Shouldn't we do this internally in `whisper.cpp` so that it is seamless for the user?
With this proposal, the user of the library has to know to divide the duration by 2

Issue #769, opened 2023-04-15
Escape Windows path seperator in JSON output


graham-walker:
When setting the model path on Windows using Windows path separators, the separator character is not escaped in the JSON output, resulting in invalid JSON. This issue is present in the 1.3.0 release.

Running:
`main --language english --model "D:\whisper.cpp 1.3.0\ggml-tiny.bin" --file "D:\whisper.cpp 1.3.0\audio_sample.wav" --output-file "D:\whisper.cpp 1.3.0\output" --output-json`

Results in the following invalid output in output.json:
```json
"params": {
  "model": "D:\whisper.cpp 1.3.0\ggml-tiny.bin",
  "language": "english",
  "translate": false
},
```

tauseefmohammed2:
Hi!
I would love to contribute to the repo. I saw that the issue is labelled "first time", which is perfect for me as I am new to this project and eager to get started. Could you please assign this issue to me?

PR #815, opened 2023-04-23
Escape quotes in csv output

laytan:
If Whisper outputs a quote the whole CSV is invalid at the moment, this runs the same function to escape them as the JSON output on the CSV output.

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -352,29 +352,6 @@ bool output_srt(struct whisper_context * ctx, const char * fname, const whisper_
     return true;
 }
 
-bool output_csv(struct whisper_context * ctx, const char * fname) {
-    std::ofstream fout(fname);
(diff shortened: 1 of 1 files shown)


dgm3333:
Why is this different to the preexisting pull request which would appear to be fixing the same issue?
https://github.com/ggerganov/whisper.cpp/pull/776
https://github.com/ggerganov/whisper.cpp/issues/769
PS you've got a spelling mistake in your title

laytan:
@dgm3333 Thanks for the heads up about the spelling mistake. The issues you mention are changing the approach of the escaping function. This PR just applies that function to the text of the CSV output.

ggerganov:
I think this needs update after the merge of #776

laytan:
@ggerganov I have rebased my branch with the changes from that merge

ggerganov (approved):

PR #835, opened 2023-04-29
command: fix out of bounds access in levenshtein distance function

grahameth:


ggerganov:
Fixed via 4a7d49af9550725b5095fdcba3306a075b3ea32e

PR #837, opened 2023-04-29
Update elevenlabs example to use official python API

DGdev91:
Follow-up to #728 
Seems like there was some confusion recently with elevenlabs-related pakages in pip.
A week ago the "elevenlabs" package pointed to the unofficial library from benaptist. Then, an official one was released.
The unofficial package was renamed as benbaptist-elevenlabs and  the official library took it's place.

More info here: https://github.com/benbaptist/elevenlabs/commit/add292a9b61b406e052a268ff351f0a51d9ddbc4

I adapted the example to use the official library.

Also, seems like the official library allows to use the API for free without setting an API key, but with limited character count. i clarified this inside speak.sh script

### DIFF
--- a/examples/talk-llama/eleven-labs.py
+++ b/examples/talk-llama/eleven-labs.py
@@ -1,23 +1,20 @@
 import sys
 import importlib.util
 
-api_key = "" #Write your https://beta.elevenlabs.io api key here
-if not api_key:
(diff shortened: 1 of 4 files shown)


DGdev91:
@ggerganov this PR has beed here for a while. Is there a specific reason to not merge it or has just been forgotten?

ggerganov (approved):
Sorry - I have lost track of this

PR #853, opened 2023-05-01
--detect-language mode

### LINKED ISSUE #826: Use whisper.cpp for language detection only
I would like to use whisper.cpp to just get information about the spoken language. I am aware of the -l auto flag and the output but I would rather see a switch that allows me to only get the language output and nothing else. Using -dl (detectlanguage) or similar, the program should guess the language of a given input file and exit without making attempts to transscribe it.

CRD716:
Closes #826 

Just a basic additional flag that ends main after detecting the language.


CRD716:
ready for review

PR #854, opened 2023-05-02
talk-llama: updating session prompts load

herrera-luis:
After further testing, I discovered that each time I restarted the talk-llama tool it was able to load the session file, but it was not able to get context about the conversation (prompts), so it provided different answers to the same question each time I restarted the tool.

This PR is an attempt to provide session information to the LLaMA context so that it can maintain a coherent conversation based on inputs provided in previous sessions.

### DIFF
--- a/examples/talk-llama/talk-llama.cpp
+++ b/examples/talk-llama/talk-llama.cpp
@@ -333,27 +333,10 @@ int main(int argc, char ** argv) {
 
     prompt_llama = ::replace(prompt_llama, "{4}", chat_symb);
 
-    // evaluate the initial prompt
-
-    auto embd_inp = ::llama_tokenize(ctx_llama, prompt_llama, true);
-
-    printf("\n");
-    printf("%s : initializing - please wait ...\n", __func__);
-
-    if (llama_eval(ctx_llama, embd_inp.data(), embd_inp.size(), 0, params.n_threads)) {
-        fprintf(stderr, "%s : failed to eval\n", __func__);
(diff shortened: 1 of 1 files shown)


examples/talk-llama/talk-llama.cpp
```diff
@@ -370,13 +353,32 @@ int main(int argc, char ** argv) {
                 return 1;
             }
             session_tokens.resize(n_token_count_out);
+            for (size_t i = 0; i < session_tokens.size(); i++) {
+                embd_inp[i] = session_tokens[i];
+            }
```
ggerganov, line 358: Isn't this overwriting on top of the prompt tokenized on line 339?
Is this what we want?

herrera-luis: @ggerganov with this approach I'm replacing the default prompt: https://github.com/ggerganov/whisper.cpp/blob/b8064208733a3d9f5a9f418448ec46a39333e272/examples/talk-llama/talk-llama.cpp#L205
And instead I'm loading the prompts from the session, so it will resume the context and keep the consistency of the previous conversations

ggerganov (approved):

ggerganov:
Got it

Btw, consider adding the following change: https://github.com/ggerganov/llama.cpp/pull/1272
It will reduce the size of the session files significantly

herrera-luis:
thank you @ggerganov I will follow the Evan's PR and will add the changes after it gets merged, it looks very good 💯

Issue #840, opened 2023-04-30
Problems compiling with cuBLAS support on Windows using CMake


RelatedTitle:
I am trying to compile whisper.cpp with cuBLAS support on Windows using CMake but ran into an error. Am I doing something wrong? 

CUDA Compiler version is 12.1.105

```
A:\AI\Whisper\src\release>cmake -DCMAKE_BUILD_TYPE=Release -DWHISPER_CUBLAS=1 ..
-- Building for: Visual Studio 17 2022
-- The C compiler identification is MSVC 19.34.31937.0
-- The CXX compiler identification is MSVC 19.34.31937.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.34.31933/bin/Hostx64/x64/cl.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.34.31933/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Warning (dev) at CMakeLists.txt:5 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "MSVC" will no longer be dereferenced when the policy
  is set to NEW.  Since the policy is not set the OLD behavior will be used.
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Found Git: C:/Program Files/Git/cmd/git.exe (found version "2.37.1.windows.1")
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - not found
-- Found Threads: TRUE
-- Found CUDAToolkit: C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v12.1/include (found version "12.1.105")
[...]

PR #867, opened 2023-05-03
Add cuBLAS build workflow and fix error causing lines in CMakeLists

RelatedTitle:
Adds a build workflow for whisper.cpp on Windows with cuBLAS support. Removes error causing lines in CMakeLists that prevented building successfully.

Credit: @giterhug for the fix.

https://github.com/ggerganov/whisper.cpp/issues/840

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -235,6 +235,61 @@ jobs:
         with:
           name: whisper-blas-bin-${{ matrix.arch }}
           path: build/bin/${{ matrix.build }}
+          
+  windows-cublas:
+    runs-on: windows-latest
+    
+    strategy:
+      matrix:
+        build: [Release]
+        arch: [x64]
+        cublas: [ON]
+        sdl2: [ON]
+        include:
+          - arch: x64
+            s2arc: x64
+          - sdl2: ON
+            s2ver: 2.26.0
+            
+    steps:
+      - name: Clone
+        uses: actions/checkout@v1
+
+      - name: Add msbuild to PATH
+        uses: microsoft/setup-msbuild@v1
+        
+      - name: Install CUDA Toolkit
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -235,6 +235,61 @@ jobs:
...
+      - name: Configure
+        run: >
+          cmake -S . -B ./build -A ${{ matrix.arch }}
+          -DCMAKE_BUILD_TYPE=${{ matrix.build }}
+          -DWHISPER_CUBLAS=1
+      
+      - name: Build
+        run: |
+          cd ./build
+          msbuild ALL_BUILD.vcxproj -t:build -p:configuration=${{ matrix.build }} -p:platform=${{ matrix.arch }}
```
Green-Sky, line 281: prefer the use of `cmake --build`

Green-Sky (changes requested):

Green-Sky:
xref llama.cpp https://github.com/ggerganov/llama.cpp/pull/1271

.github/workflows/build.yml
RelatedTitle: Could you explain why this is better than the current one? I'm not really that well-versed in CMake or compiling C++ projects in general so I just used the same as the other Windows builds.

Green-Sky: It makes it buildsystem agnostic. In theory we could switch to eg. Ninja, without touching that line.
Consider it best practice.

RelatedTitle: Hmm, wouldn't it be best to switch all the builds to that all at once?

ggerganov: Sure, we can do that in a follow up PR

ggerganov (approved):

Issue #777, opened 2023-04-17
Build fails on Fedora 38


jkrimmer:
Building whisper.cpp on Fedora 38 (GCC 13) fails due to a multitude of undefined symbols. The first error occurs since `uint8_t` has not been defined in this scope. As far as I have understood, this behavior is caused by a [bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=107134) in GCC 13. Adding 
```
#include <stdint.h>
```
in `examples/common.cpp`, resolves this issue.

PR #871, opened 2023-05-04
fix gcc13 build

frostworx:
[this issue](https://github.com/ggerganov/whisper.cpp/issues/777) is reproducible also on Arch Linux _(using [this AUR package](https://aur.archlinux.org/packages/whisper.cpp))_
The simple commit fixes the issue

### DIFF
--- a/examples/common.cpp
+++ b/examples/common.cpp
@@ -5,7 +5,7 @@
 #define DR_WAV_IMPLEMENTATION
 #include "dr_wav.h"
 
-#include <stdint.h>
+#include <cstdint>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


examples/common.cpp
```diff
@@ -5,6 +5,7 @@
 #define DR_WAV_IMPLEMENTATION
 #include "dr_wav.h"
 
+#include <stdint.h>
```
ggerganov, line 8: Does this work:

```suggestion
#include <cstdint>
```

frostworx:
it does. just bumped the PR. thanks for looking into it.

PR #875, opened 2023-05-05
fixed vs2017 build

mtdxc:


ggerganov:
Yes, let's limit to VS2019 and above

PR #883, opened 2023-05-05
nix: add flake

cmoog:
This PR adds a Nix Flake for building and distributing the binary.

Derivation adapted from [nixpkgs-unstable](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/tools/audio/openai-whisper-cpp/default.nix).

### Example

```
$ nix run github:ggerganov/whisper.cpp -- --help

usage: /nix/store/z4zqr5nklvf8n80bxqxaphc3a3j9xddb-whisper-cpp/bin/whisper-cpp [options] file0.wav file1.wav ...

options:
  -h,        --help              [default] show this help message and exit
  -t N,      --threads N         [4      ] number of threads to use during computation
  -p N,      --processors N      [1      ] number of processors to use during computation
  -ot N,     --offset-t N        [0      ] time offset in milliseconds
  -on N,     --offset-n N        [0      ] segment index offset
  -d  N,     --duration N        [0      ] duration of audio to process in milliseconds
  -mc N,     --max-context N     [-1     ] maximum number of text context tokens to store
  -ml N,     --max-len N         [0      ] maximum segment length in characters
.......
```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1681202837,
+        "narHash": "sha256-H+Rh19JDwRtpVPAWp64F+rlEtxUWBAQW28eAi3SRSzg=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "cfacdce06f30d2b68473a46042957675eebb3401",
+        "type": "github"
+      },
+      "original": {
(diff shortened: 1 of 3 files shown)


pkese:
I'd be interested in having a flake that would provide libwhisper.so  
built with `WHISPER_CUBLAS=1 make -j`.

I'm not very fluent in nix. Can this flake be extended (parametrized) to also provide such a library, or should I take your nix file as example and change it according to my needs (e.g. build a library instead of an executable)?

Issue #904, opened 2023-05-10
fix memory leak has introduced a segmentation fault in stream while using coreml


dev-mush:
The changes introduced in this pr https://github.com/ggerganov/whisper.cpp/pull/899 are causing segmentation fault crashes when executing ./stream. 

I've inspected the code a little, and my safe bet is that the program is attempting to release possibily nil objects, I don't see any other big changes from the previous version that could justify the segfault... but am not touching c++/objc code since ages so might be wrong.

multispecies-lab:
Same issue here...

ggerganov:
Should be fixed now

PR #910, opened 2023-05-10
whisper.objc : enable Core ML in example & fix segmentation fault

jhen0409:
I got some build issues from https://github.com/mybigday/whisper.rn/pull/41, I think this can be reproduced directly here. I also updated the example.

This PR is mainly turned on ARC for whisper-encoder.mm (as another files already did). It also fix segmentation fault issue. (It can be reproduced by stream example)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -333,7 +333,9 @@ jobs:
         uses: actions/checkout@v1
 
       - name: Configure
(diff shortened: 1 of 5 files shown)


jhen0409:
It also seems to fix #904.

@ggerganov please review this. Thanks!

ggerganov (approved):

PR #927, opened 2023-05-15
ci: Build with any BLAS compatible library

akharlamov:
The change allows building whisper.cpp with any BLAS implementation, which in our case required for Intel oneMKL.


CMakeLists.txt
```diff
@@ -147,7 +159,7 @@ if (WHISPER_CUBLAS)
 
     if (CUDAToolkit_FOUND)
         message(STATUS "cuBLAS found")
-
+        set(CMAKE_CUDA_COMPILER /usr/local/cuda/bin/nvcc)
```
ggerganov, line 162: Is this needed?
In `llama.cpp` we don't have this line: 

https://github.com/ggerganov/llama.cpp/blob/29cf5596fe0c37213f9b74e80d8f631193a93f0f/CMakeLists.txt#L186-L188

ggerganov (approved):

CMakeLists.txt
akharlamov: Removed the explicit search path for nvcc and rebased to upstream/master

PR #939, opened 2023-05-20
Add Swift Package Manager manifest

afarnham:
This PR add a manifest so this repo can be targeted from Swift projects and Xcode using Swift Package Manager.

In order to expose only whisper.h outside the package I've added a new directory, `swiftpm` with a shim header. If the publicHeadersPath were specified as the root of the repo then all headers in all directories become public as Swift Package manager recurses its public headers directory without any option to turn that off, nor is there an option to specify specific headers. 

I've also updated .gitignore prevent check in of any artifacts Swift Package Manager introduces when building and editing.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,48 @@
+// swift-tools-version:5.3
+import PackageDescription
+
+let package = Package(
+    name: "whisper.cpp",
+    platforms: [
+        .iOS(.v11),
+        .macOS(.v10_13),
+        .watchOS(.v4),
+        .tvOS(.v11)
+    ],
+    products: [
+        .library(
+            name: "whispercpp",
+            targets: ["whispercpp"]
+        )
+    ],
+    targets: [
+        .target(
+            name: "whispercpp",
+            dependencies: [],
+            path: ".",
+            exclude: [
+                "bindings",
+                "cmake",
+                "coreml",
+                "examples",
+                "extra",
+                "models",
+                "samples",
+                "tests",
+                "CMakeLists.txt",
+                "ggml-cuda.cu",
+                "ggml-cuda.h",
+                "ggml-opencl.c",
+                "ggml-opencl.h",
+                "Makefile"
+            ],
+            sources: [
+                "whisper.cpp",
+                "ggml.c"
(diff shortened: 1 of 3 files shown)


chidiwilliams:
@afarnham, there's a Swift package at https://github.com/ggerganov/whisper.spm

sindresorhus:
I think it would be better to add the manifest here as it will automatically be kept up to date with Whisper.cpp and it would be easier to use a specific Whisper.cpp commit in a Swift project.

Package.swift
```diff
@@ -0,0 +1,48 @@
...
+                "ggml-cuda.h",
+                "ggml-opencl.c",
+                "ggml-opencl.h",
+                "Makefile"
+            ],
+            sources: [
+                "whisper.cpp",
+                "ggml.c"
+            ],
+            publicHeadersPath: "swiftpm"
```
sindresorhus, line 43: You probably need `cSettings` too: https://github.com/ggerganov/whisper.spm/blob/master/Package.swift#L12

Package.swift
```diff
@@ -0,0 +1,48 @@
+// swift-tools-version:5.3
```
sindresorhus, line 1:
```suggestion
// swift-tools-version:5.7
```

The convention is to target the current stable Xcode version's Swift, which is 5.7.

sindresorhus (changes requested):

sindresorhus:
@ggerganov I would be happy to make an updated pull request for this and commit to maintain this file if it's merged. This would make it easier for you too as you don't have to maintain another repository.

sindresorhus:
I just noticed that llama.cpp has a Package.swift in the main repo, which is another argument for doing it here too: https://github.com/ggerganov/llama.cpp/blob/master/Package.swift

The easist would be to copy-paste that one and simply change a few names.

ggerganov:
@sindresorhus 

I think this would be great! Definitely need a helping hand and I agree the current solution is not the best.

Btw, I'm planning a new release with parallel decoding support for efficient beam-search. Hoping to do it in the following week, though some more delay can be expected.

Issue #427, opened 2023-01-19
Add option to extend Confidence color-coding to CSV or some other output file format


DIGist:
So for spot checking long transcriptions, it would be helpful to have the confidence score output to some file.. either as an extra column in the CSV or maybe some metadata in VTT or even it's own log or something.

Thanks! This is amazing work!

RYucel:
This feature could be very helpful indeed for further correction of outputted text manually

CodeFromAnywhere:
UPVOTE! This would be incredibly powerful for my use-case.

CyberGoat:
One can log the output into a file then open it via this

> vim -c ":term ++hidden ++open cat <FILENAME>" 

This will enable one to view it as it was in vim to view.

PR #971, opened 2023-05-30
Add support for outputting token-level confidence

pgdr:
This is currently a **request for feedback**.

It is an attempt of addressing #427 and #571.

This commit adds support for outputting confidence (0-100%) on a token level for the `.txt` and `.csv` outputs.

One new argument is added: `-oc | --output-confidence`

When run with `otxt`:

```
Adding (65.8%)  the (94.7%)  -- (7.3%) out (81.7%) put (99.9%)  confidence (2.7%)  argument (97.8%) , (63.7%)  the (32.1%)  output (97.9%)  of (95.0%)  the (92.8%)  trans (92.6%) cribed (98.2%)  text (94.3%)  will (70.2%)  be (89.2%)  annot (53.7%) ated (99.5%)  with (44.4%)  the (73.8%)  probability (95.4%)  of (93.3%)  the (94.9%)  specified (98.8%)  token (96.6%)  in (94.2%)  the (97.5%)  specified (93.7%)  segment (96.9%) . (53.7%) 
```

When run with `ocsv`:

|   start |   end | text        |   confidence |
|---------|-------|-------------|--------------|
|       0 | 12480 | Adding      |         65.8 |
|       0 | 12480 | the         |         94.7 |
|       0 | 12480 | --          |          7.3 |
|       0 | 12480 | out         |         81.7 |
|       0 | 12480 | put         |         99.9 |
|       0 | 12480 | confidence  |          2.7 |
|       0 | 12480 | argument    |         97.8 |
|       0 | 12480 | ,           |         63.7 |
|       0 | 12480 | the         |         32.1 |
|       0 | 12480 | output      |         97.9 |

---

It probably makes more sense to add the confidence when the output is `json` files, but I would like to get some feedback first.

I have some questions:

1. Is it possible to get timing on token level?  E.g. instead of this signature
    * `WHISPER_API int64_t whisper_full_get_segment_t0           (struct whisper_context * ctx, int i_segment);`, I need this:
    * `WHISPER_API int64_t whisper_full_get_segment_t0           (struct whisper_context * ctx, int i_segment, int i_token);`
2. Is it possible to get _word-level_ confidence?
3. Should we allow confidence in `txt` and `csv`, or only in `json`?
4. Should we output _both_ the full-text and also word/token level confidence?  If so, what should the JSON schema be?
5. If we output confidence to `txt`, why not also `srt`?
6. We could alternatively add a _confidence threshold_ such that every word/token with confidence < threshold is simply given a `[?]` postfix.

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -9,6 +9,7 @@
 #include <thread>
 #include <vector>
 #include <cstring>
+#include <iomanip>
 
 // Terminal color map. 10 colors grouped in ranges [0.0, 0.1, ..., 0.9]
 // Lowest is red, middle is yellow, highest is green.
@@ -77,6 +78,7 @@ struct whisper_params {
     bool output_csv     = false;
     bool output_jsn     = false;
     bool output_lrc     = false;
+    bool output_confidence = false;
     bool print_special  = false;
     bool print_colors   = false;
     bool print_progress = false;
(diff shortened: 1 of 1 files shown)


ggerganov:
1. You can use: https://github.com/ggerganov/whisper.cpp/blob/77eab3fbfe5e5462021d92dd230076bba06eefbc/whisper.h#L475-L478
2. Currently, only token-level probs are supported
3. I think just `json` is enough
4. No preference

PR #972, opened 2023-05-30
Support decode wav file has 2 channels.

geniusnut:


geniusnut:
Support decode wav file which has two channels in Android Demo.

mrfragger:
This will limit the max duration of audio to 18 hours as it's twice the size of mono wav file. Current max is 36 hours.

ggerganov (approved):

PR #1001, opened 2023-06-11
Make convert-pt-to-ggml.py backwards compatible with older vocab.json tokenizer files

akashmjn:
Patches the script to determine what type of tokenizer files are present and convert appropriately.

Mostly borrows from  https://github.com/ggerganov/whisper.cpp/pull/725 as a reference. For the same reason mentioned in that PR, converted multilingual checkpoints will continue to exactly match while `.en` checkpoints have a 17 byte difference compared to ggml files downloaded by whisper.cpp.

The script will also now produce exactly the same files regardless of which type of tokenizer files were used for conversion.

```
-rw-r--r--  1 Akash  staff  487614184 Jun 10 21:30 ggml-small.en.hf.bin
-rw-r--r--  1 Akash  staff  487614184 Jun 10 21:28 ggml-small.en.tiktoken.bin
-rw-r--r--  1 Akash  staff  487601967 Jun 10 21:32 ggml-small.hf.bin
-rw-r--r--  1 Akash  staff  487601967 Jun 10 21:35 ggml-small.tiktoken.bin
```

### DIFF
--- a/models/convert-pt-to-ggml.py
+++ b/models/convert-pt-to-ggml.py
@@ -235,7 +235,7 @@ if not tokenizer.is_file():
     tokenizer = dir_whisper / "whisper" / "assets" / (multilingual and "multilingual" or "gpt2") / "vocab.json"
     tokenizer_type = "hf_transformers"
     if not tokenizer.is_file():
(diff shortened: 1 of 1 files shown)


ggerganov (approved):

ggerganov:
Thank you!

PR #1003, opened 2023-06-11
Change ggml.c to make it compatible with C++ compilers.

Alejandro-FA:
In a small cross-platform mobile project that I'm working on I've been forced to compile `ggml.c` with a C++ compiler (`clang++`). This has exposed some minor incompatibilities of `ggml.c` with C++ compilers, which can be easily solved.

### DIFF
--- a/ggml.c
+++ b/ggml.c
@@ -27,6 +27,10 @@
 #define static_assert(cond, msg) struct global_scope_noop_trick
 #endif
 
+#ifdef __cplusplus
+#define restrict __restrict
+#endif
+
 #if defined(_WIN32)
 
 #include <windows.h>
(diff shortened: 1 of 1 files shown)


fire:
Are the binary files supposed to be included?

Alejandro-FA:
The binary files were not supposed to be included. I have deleted them from the branch, sorry for the trouble.

fire:
Do you know if whisper.cpp has a 1 merge commit policy? The binary files are still in the pull request. You probably need to squash.

Alejandro-FA:
> Do you know if whisper.cpp has a 1 merge commit policy? The binary files are still in the pull request. You probably need to squash.

I don't know if they have such a policy, but as far as I understand it is a decision to take when merging, not now.

Are you sure about the binary files? I only see 1 file changed in the pull request.

ziegenberg:
Commit ec6c2cd79f15b0b477d1640141b05af07dc7abdd, which is part of this PR, does indeed include the binary files. You could squash your commits and force-push.

Issue #876, opened 2023-05-05
"native" is not a defined option for 'gpu-architecture'


cybersandwich:
I don't know if this is specific to me. I am running a headless GTX 1080. I have the nvidia-530 drivers and the nvidia-cuda-toolkit installed (running pop_os so these are likely whats in the ubuntu repos as well). 

When I attempted to perform: 

```
make clean
WHISPER_CUBLAS=1 make -j
```

I received this error

```
nvcc fatal   : Value 'native' is not defined for option 'gpu-architecture'
make: *** [Makefile:171: ggml-cuda.o] Error 1
make: *** Waiting for unfinished jobs....
```

`nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2021 NVIDIA Corporation
Built on Thu_Nov_18_09:45:30_PST_2021
Cuda compilation tools, release 11.5, V11.5.119
Build cuda_11.5.r11.5/compiler.30672275_0`

I fixed this by editing the make file with:

NVCCFLAGS   = --forward-unknown-to-host-compiler -arch=compute_75

The compute_75 was one of the listed available options when I did nvcc --help 

`--gpu-architecture <arch>                       (-arch)
        Specify the name of the class of NVIDIA 'virtual' GPU architecture for which
        the CUDA input files must be compiled.
        With the exception as described for the shorthand below, the architecture
        specified with this option must be a 'virtual' architecture (such as compute_50).
        Normally, this option alone does not trigger assembly of the generated PTX
        for a 'real' architecture (that is the role of nvcc option '--gpu-code',
        see below); rather, its purpose is to control preprocessing and compilation
        of the input to PTX.
        For convenience, in case of simple nvcc compilations, the following shorthand
        is supported.  If no value for option '--gpu-code' is specified, then the
        value of this option defaults to the value of '--gpu-architecture'.  In this
        situation, as only exception to the description above, the value specified
        for '--gpu-architecture' may be a 'real' architecture (such as a sm_50),
        in which case nvcc uses the specified 'real' architecture and its closest
        'virtual' architecture as effective architecture values.  For example, 'nvcc
        --gpu-architecture=sm_50' is equivalent to 'nvcc --gpu-architecture=compute_50
        --gpu-code=sm_50,compute_50'.
        -arch=all         build for all supported architectures (sm_*), and add PTX
        for the highest major architecture to the generated code.
        -arch=all-major   build for just supported major versions (sm_*0), plus the
[...]

PR #1012, opened 2023-06-13
Fix for issue #876

burningion:
For the following issue:

https://github.com/ggerganov/whisper.cpp/issues/876#issue-1696896105

GPU support won't build without the change from `native` to `any`. Confirmed working with a 4090.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ ifdef WHISPER_CUBLAS
 	LDFLAGS     += -lcublas -lculibos -lcudart -lcublasLt -lpthread -ldl -lrt -L/usr/local/cuda/lib64 -L/opt/cuda/lib64 -L$(CUDA_PATH)/targets/x86_64-linux/lib
 	WHISPER_OBJ += ggml-cuda.o
 	NVCC        = nvcc
-	NVCCFLAGS   = --forward-unknown-to-host-compiler -arch=native
+	NVCCFLAGS   = --forward-unknown-to-host-compiler -arch=any
 
 ggml-cuda.o: ggml-cuda.cu ggml-cuda.h
 	$(NVCC) $(NVCCFLAGS) $(CXXFLAGS) -Wno-pedantic -c $< -o $@


ggerganov (approved):
Do we know that `all` does not reduce the performance compared to `native` ?

byte-6174:
It most certainly would in some sense. For example when I compile with `all` the binary is 1.5MB vs. 812KB when I give the exact architecture I am compiling for. 
The exact architecture is the best way to make sure you have the most optimized code for it, as it does not create the intermediate ptx code. With option `all`, nvcc would generate the ptx and postpone compilation at runtime. Nvidia notes this would lead to app startup delays. It could be addressed by caching, however.

byte-6174:
for example I just ran w a binary the load time for a model (`models/ggml-base.en.bin`) was:

`-gpu-architecture=all`:
`whisper_print_timings:     load time =  4452.92 ms`

vs 
`-gpu-architecture=sm_72`:
`whisper_print_timings:     load time =   224.77 ms`

ggerganov:
So, is it worth switching to `all`?
Maybe have `native` by default and only use `all` upon setting certain build option

cdeisler:
As an FYI, this technically still breaks on WSL2 (Ubuntu 22).  The quick workaround is to use "all".

```
nvcc --forward-unknown-to-host-compiler -arch=any -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC -D_XOPEN_SOURCE=600 -pthread -DGGML_USE_CUBLAS -I/usr/local/cuda/include -I/opt/cuda/include -I/targets/x86_64-linux/include -Wno-pedantic -c ggml-cuda.cu -o ggml-cuda.o
nvcc fatal   : Value 'any' is not defined for option 'gpu-architecture'
```

byte-6174:
yes, cuda optimizations are highly specific to the hardware. `all` will allow compilation without errors, but with suboptimal binary.
whereas `native` would work sometimes based on arch. / cuda version, but when it works will produce the best binary 😄 

In that regard, I like the above suggestion. However, I'm reading in other [comments](https://github.com/ggerganov/whisper.cpp/issues/876#issuecomment-1542210182) though for some arch/cuda combos neither works :( and one has to specify the exact arch like `sm_72` etc.
So there is some finetuning of the makefile when it comes to cuda.

FerLuisxd:
It doesn't work on 2080 and version 12.1

PR #1027, opened 2023-06-17
Do not use _GNU_SOURCE gratuitously.

przemoc:
What is needed to build whisper.cpp and examples is availability of stuff defined in The Open Group Base Specifications Issue 6 (https://pubs.opengroup.org/onlinepubs/009695399/) known also as Single Unix Specification v3 (SUSv3) or POSIX.1-2001 + XSI extensions.

There is no need to penalize musl libc which simply follows standards.

Not having feature test macros in source code gives greater flexibility to those wanting to reuse it in 3rd party app, as they can build it with minimal FTM (_XOPEN_SOURCE=600) or other FTM depending on their needs.

It builds without issues in Alpine (musl libc), Ubuntu (glibc), MSYS2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,11 +36,12 @@ CFLAGS   = -I.              -O3 -DNDEBUG -std=c11   -fPIC
 CXXFLAGS = -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC
 LDFLAGS  =
 
-# ref: https://github.com/ggerganov/whisper.cpp/issues/37
-ifneq ($(wildcard /usr/include/musl/*),)
-	CFLAGS += -D_POSIX_SOURCE -D_GNU_SOURCE
-	CXXFLAGS += -D_POSIX_SOURCE -D_GNU_SOURCE
-endif
+# clock_gettime came in POSIX.1b (1993)
+# CLOCK_MONOTONIC came in POSIX.1-2001 / SUSv3 as optional
+# posix_memalign came in POSIX.1-2001 / SUSv3
+# M_PI is an XSI extension since POSIX.1-2001 / SUSv3, came in XPG1 (1985)
+CFLAGS   += -D_XOPEN_SOURCE=600
+CXXFLAGS += -D_XOPEN_SOURCE=600
 
 # OS specific
 # TODO: support Windows
--- a/examples/command/command.cpp
+++ b/examples/command/command.cpp
@@ -6,8 +6,8 @@
 // ref: https://github.com/ggerganov/whisper.cpp/issues/171
 //
 
-#include "common.h"
 #include "common-sdl.h"
+#include "common.h"
 #include "whisper.h"
 
 #include <sstream>
--- a/examples/stream/stream.cpp
+++ b/examples/stream/stream.cpp
@@ -3,8 +3,8 @@
 // A very quick-n-dirty implementation serving mainly as a proof of concept.
 //
 
-#include "common.h"
 #include "common-sdl.h"
+#include "common.h"
 #include "whisper.h"
 
 #include <cassert>
--- a/examples/talk-llama/llama.cpp
+++ b/examples/talk-llama/llama.cpp
@@ -1,11 +1,3 @@
-// Defines fileno on msys:
-#ifndef _GNU_SOURCE
-#define _GNU_SOURCE
-#include <cstddef>
-#include <cstdint>
-#include <cstdio>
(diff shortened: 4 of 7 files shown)


ggerganov:
The `macOS-latest` action seems to fail with this change - SDL2 is not happy for some reason

przemoc:
Thanks for running CI checks. I do not have access to any macOS platform, so debugging it may be a bit challenging.

```
I whisper.cpp build info: 
I UNAME_S:  Darwin
I UNAME_P:  i386
I UNAME_M:  x86_64
I CFLAGS:   -I.              -O3 -DNDEBUG -std=c11   -fPIC -D_XOPEN_SOURCE=600 -pthread -mf16c -mavx -DGGML_USE_ACCELERATE
I CXXFLAGS: -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC -D_XOPEN_SOURCE=600 -pthread
I LDFLAGS:   -framework Accelerate
I CC:       Apple clang version 14.0.0 (clang-1400.0.29.202)
I CXX:      Apple clang version 14.0.0 (clang-1400.0.29.202)

c++ -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC -D_XOPEN_SOURCE=600 -pthread examples/stream/stream.cpp examples/common.cpp examples/common-ggml.cpp examples/common-sdl.cpp ggml.o whisper.o -o stream `sdl2-config --cflags --libs`  -framework Accelerate
In file included from examples/stream/stream.cpp:7:
In file included from ./examples/common-sdl.h:3:
In file included from /usr/local/include/SDL2/SDL.h:32:
In file included from /usr/local/include/SDL2/SDL_main.h:25:
/usr/local/include/SDL2/SDL_stdinc.h:532:5: error: use of undeclared identifier 'memset_pattern4'
    memset_pattern4(dst, &val, dwords * 4);
    ^
1 error generated.
```

przemoc:
Looks like my attempt at fixing macOS issue was successful, others checks will be most likely successful too, but let's wait for them to finish.

ggerganov:
I synced the latest `ggml` from `llama.cpp` and the new NUMA-related changes fail to compile:

https://github.com/ggerganov/whisper.cpp/actions/runs/5438047714/jobs/9889005968

Temporary reverted the change until we resolve the build issues.
I think we should first fix https://github.com/ggerganov/llama.cpp/pull/2035 and then I will propagate the changes back to `whisper.cpp`

przemoc:
Sorry, @ggerganov, it turned out I didn't have time this weekend to play with `llama.cpp` and cmake. :(
Adding [selected lines in 1st commit](https://github.com/ggerganov/llama.cpp/pull/2035/commits/09fcec856b982d52eee2fdcd080c77b24d7d93a9#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R62-R68) from https://github.com/ggerganov/llama.cpp/pull/2035 would most likely solve failing build in `whisper.cpp`, but of course only for Makefile-driven builds.

BTW What's cmake usage state in `whisper.cpp`? I don't think CI was doing any cmake-driven build last time I've seen those PR actions. Is it expected to be working? If yes, then I guess I may have broken it with my FTM sanitization patches that weren't touching CMakeLists.txt.

ggerganov:
The main purpose for having CMake build system is to support Windows.

I think the CI wasn't failing before because the NUMA changes were not present.
Yesterday I synced them from the `llama.cpp` repo and they started causing problems.

Don't worry - whenever you have the time. It's not a big problem

przemoc:
I looked at `llama.cpp` a bit on Sunday and went with Howard's approach for CMake changes to simply follow recent FTM improvements from Makefile - seemed like viable way way of doing that. I had to fix other things first, though, to make it build in MSYS2.

@ggerganov, when you'll have a spare minute, please trigger PR checks in ggerganov/llama.cpp#2035 and the other one mentioned there that should be merged first, so that I could see if there are still any issues that I should resolve.

And when that will be clean, I can create analogous remove-gnu-source-ftm for `whisper.cpp`. Not sure if GitHub allows merging second time from same PR (considering it was reverted), as I would prefer to have updated rebased version here and not create yet another PR.

przemoc:
Thanks for triggering PR checks in `llama.cpp`. Stuff seem to be building fine there. I ended up creating new PR in `whisper.cpp`, #1129, to avoid potential issues with GitHub merging UX.

PR #1034, opened 2023-06-21
typofix README.md

DIRAKHIL:


bobqianic:
Actually, we should use `naïve` or `naive` in this case, not `native`. It's because the implementation isn't quite refined.

Issue #1038, opened 2023-06-23
Model download fails if the destination folder includes a space


matthewloewen:
Issue: using `bash ./models/download-ggml-model.sh base.en` as described in Quick Start fails to save the file on macOS 13.4.1.

The folder I was trying to download to has the path of `/Users/name/App\ Development/Tools/whisper.cpp`.  The file is downloaded, but you can see from the path that it is truncated at the space. Unfortunately, while the error is noted, the final message suggests that everything was completed successfully.

```
Downloading ggml model base.en from 'https://huggingface.co/ggerganov/whisper.cpp' ...
./models/download-ggml-model.sh: line 57: cd: /Users/name/App: No such file or directory
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1206  100  1206    0     0   1375      0 --:--:-- --:--:-- --:--:--  1378
100  141M  100  141M    0     0  18.3M      0  0:00:07  0:00:07 --:--:-- 29.3M
Done! Model 'base.en' saved in 'models/ggml-base.en.bin'
You can now use it like this:

  $ ./main -m models/ggml-base.en.bin -f samples/jfk.wav
```

Running the ./main results in:

```
 ./main -f samples/jfk.wav
whisper_init_from_file_no_state: loading model from 'models/ggml-base.en.bin'
whisper_init_from_file_no_state: failed to open 'models/ggml-base.en.bin'
error: failed to initialize whisper context
```

At first I thought I made a mistake, so I ran `make base.en` which confusingly told me that the folder didn't exist and that base.en already existed.

```
bash ./models/download-ggml-model.sh base.en
Downloading ggml model base.en from 'https://huggingface.co/ggerganov/whisper.cpp' ...
./models/download-ggml-model.sh: line 57: cd: /Users/name/App: No such file or directory
Model base.en already exists. Skipping download.
```

Changing whisper.cpp location to a path that does not include a space allowed me to complete the download / install successfully.

macOS 13.4.1
whisper.cpp 1.4.0

janngobble:
This was ONCE taken care of in this issue (I opened it awhile back in Dec 2022 and it was immediately addressed):

https://github.com/ggerganov/whisper.cpp/issues/315

The fix was:

double-quoting:

models_path=$(get_script_path)

as

models_path="$(get_script_path)"


Did something change, @ggerganov?

ggerganov:
[...]

PR #1041, opened 2023-06-24
Fix cd statements to allow spaces in model path

roddurd:
Hi @ggerganov, 

`download-ggml-model.sh` and a couple other files currently do not quote `cd` statements, leading to issues when working in directories with spaces in the path:

whisperbug

whisperfix

I've fixed this issue.


roddurd:
Just noticed this PR addresses [this](https://github.com/ggerganov/whisper.cpp/issues/1038) recently opened issue. Looks like it was addressed before but something changed since then.

ggerganov (approved):

PR #1046, opened 2023-06-25
whisper : `split_on_word` no longer trims

### LINKED ISSUE #1030: `split_on_word = true` should not trim the segment text
When it's `false`, I get `" it, to get…"`, but when it's `true`, I get `"it, to get…"`. Notice the missing leading space.

It seems this was an intentional change by https://github.com/ggerganov/whisper.cpp/pull/476. However, that pull request is flawed. It was supposed to fix subtitles incorrectly having leading spaces (https://github.com/ggerganov/whisper.cpp/issues/397), but it's fixing the problem in the wrong place. The problem should be fixed in the code that generates the subtitles, not by changing the segments when `split_on_word = true`. That change affects a lot more than just subtitles.

My use-case: I want to use this option in my transcription app, but my app allows toggling between timestamp (subtitle) view and reader view. I need the leading spaces preserved in the segments, so I can combine them into text when showing reader view.

ggerganov:
close #1030

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -3401,26 +3401,6 @@ static void whisper_exp_compute_token_level_timestamps(
                          float   thold_pt,
                          float   thold_ptsum);
 
-// trim from start (in place)
-static inline void ltrim(std::string &s) {
-    s.erase(s.begin(), std::find_if_not(s.begin(), s.end(), [](unsigned char ch) {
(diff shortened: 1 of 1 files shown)


ggerganov:
cc @sindresorhus

sindresorhus (approved):

PR #1060, opened 2023-06-28
expose api to let user control log output

evmar:
Add
  whisper_set_log_callback()
that lets user set a callback for log messages.

Change all the
  fprintf(stderr, ...)
to call via the above.

This is nice when embedding whisper.cpp into a larger program that doesn't want to print stuff to stderr.


evmar:
Note: I also considered making a log macro that included the `__func__` bit automatically, lemme know if you think that's a good idea.  Also if it's literally named `log` then it shadows `std::log` the math function so it would need a different name.

whisper.cpp
```diff
@@ -711,6 +711,20 @@ struct whisper_context {
...
 };
 
+static void whisper_default_log(const char * text) {
+    fprintf(stderr, "%s\n", text);
+}
+
+static whisper_log_callback whisper_log = whisper_default_log;
+
+static void log(const char * fmt, ...) {
+    char buf[256];
```
ggerganov, line 721:
```suggestion
    char buf[1024];
```

whisper.cpp
```diff
@@ -711,6 +711,20 @@ struct whisper_context {
     std::string path_model; // populated by whisper_init_from_file()
 };
 
+static void whisper_default_log(const char * text) {
+    fprintf(stderr, "%s\n", text);
```
ggerganov, line 715: New lines must be explicitly logged

```suggestion
    fprintf(stderr, "%s", text);
```

whisper.cpp
```diff
@@ -711,6 +711,20 @@ struct whisper_context {
...
+}
+
+static whisper_log_callback whisper_log = whisper_default_log;
+
+static void log(const char * fmt, ...) {
+    char buf[256];
+    va_list args;
+    va_start(args, fmt);
+    vsnprintf(buf, sizeof(buf), fmt, args);
+    whisper_log(buf);
```
ggerganov, line 725: Check if `whisper_log` is `nullptr`. If it is - do nothing

ggerganov (changes requested):

evmar:
OK, I applied your suggestions and rebased.

PR #1062, opened 2023-06-28
Fix talk-llama build on macOS.

przemoc:
`whisper.cpp` CI does not do all available make targets, so it has been overlooked. Follow up to #1027.


przemoc:
@jasontitus, thanks for reporting issue with talk-llama, but you did it in the unrelated PR.
Please try this PR to see if the issue is fixed for you here.

ggerganov (approved):

PR #1101, opened 2023-07-12
Multi platforms CI

alonfaraj:
Related https://github.com/ggerganov/ggml/issues/295.

This pull request enhances the capability to execute **Linux** builds & tests across multiple architectures by leveraging QEMU and docker buildx. It also offers easy extensibility for supporting additional architectures. 
Additionally, a build for **FreeBSD** was added.
The supported platforms for Linux include:

-   **linux/amd64**
 -   **linux/arm64**
-   linux/riscv64
-   **linux/ppc64le**
-   linux/s390x
-   linux/386
-   linux/mips64le
-   linux/mips64
-   **linux/arm/v7**
-   linux/arm/v6

Architectures included in this pull request are highlighted in bold, but it ultimately falls upon the maintainers to determine their relevance. 
Note that not all the jobs passed successfully, as some test failed in ubuntu gcc sanitized .
These issues can be addressed in separate pull requests.

Additionally, this pull request incorporates a few minor changes:

- Bump `actions/checkout` to v3
- Utilized a dedicated action for Emscripten.
- CMake improvements


ggerganov:
It's strange that the sanitizer builds fail for armv7 and linux/arm64 builds.
I don't have any way to test and fix this, so if we cannot figure out what is causing the issue, we should remove these builds from the build matrix before merging

alonfaraj:
I was trying to investigate it but no success so far. 
I will remove them for now.

PR #1113, opened 2023-07-16
Now make tests can be called as make tests base.en

Jerry-Master:


Jerry-Master:
Or make tests large, I simply forwarded the argument from the command line to the bash script. It seems more comfortable this way.

ggerganov (approved):

PR #1120, opened 2023-07-19
Fix CLBlast build on MacOS

iceychris:


marty1885:
Second this. I also run into the same issue on my MBP.

PR #1129, opened 2023-07-21
Do not use _GNU_SOURCE gratuitously.

przemoc:
Follow up to previous attempt #1027 that had to be reverted due to not being complete, which was exposed after syncing new stuff from `llama.cpp`. It uses improvements from ggerganov/llama.cpp#2035 (courtesy of @howard0su).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -321,6 +321,49 @@ else()
     endif()
 endif()
 
+# clock_gettime came in POSIX.1b (1993)
+# CLOCK_MONOTONIC came in POSIX.1-2001 / SUSv3 as optional
+# posix_memalign came in POSIX.1-2001 / SUSv3
+# M_PI is an XSI extension since POSIX.1-2001 / SUSv3, came in XPG1 (1985)
+add_compile_definitions(_XOPEN_SOURCE=600)
+
+# Somehow in OpenBSD whenever POSIX conformance is specified
+# some string functions rely on locale_t availability,
+# which was introduced in POSIX.1-2008, forcing us to go higher
+IF (CMAKE_SYSTEM_NAME MATCHES "OpenBSD")
+    remove_definitions(-D_XOPEN_SOURCE=600)
+    add_compile_definitions(_XOPEN_SOURCE=700)
+ENDIF()
+
+# Data types, macros and functions related to controlling CPU affinity
+# are available on Linux through GNU extensions in libc
+IF (CMAKE_SYSTEM_NAME MATCHES "Linux")
+    add_compile_definitions(_GNU_SOURCE)
+ENDIF()
+
+# RLIMIT_MEMLOCK came in BSD, is not specified in POSIX.1,
+# and on macOS its availability depends on enabling Darwin extensions
+# similarly on DragonFly, enabling BSD extensions is necessary
+IF (CMAKE_SYSTEM_NAME MATCHES "Darwin")
+    add_compile_definitions(_DARWIN_C_SOURCE)
+ENDIF()
+IF (CMAKE_SYSTEM_NAME MATCHES "DragonFly")
+    add_compile_definitions(_DARWIN_C_SOURCE)
+ENDIF()
+
+# alloca is a non-standard interface that is not visible on BSDs when
+# POSIX conformance is specified, but not all of them provide a clean way
+# to enable it in such cases
+IF (CMAKE_SYSTEM_NAME MATCHES "FreeBSD")
+    add_compile_definitions(__BSD_VISIBLE)
+ENDIF()
+IF (CMAKE_SYSTEM_NAME MATCHES "NetBSD")
+    add_compile_definitions(_NETBSD_SOURCE)
+ENDIF()
+IF (CMAKE_SYSTEM_NAME MATCHES "OpenBSD")
+    add_compile_definitions(_BSD_SOURCE)
+ENDIF()
+
 if (WHISPER_PERF)
     set(WHISPER_EXTRA_FLAGS ${WHISPER_EXTRA_FLAGS} -DGGML_PERF)
 endif()
--- a/Makefile
+++ b/Makefile
@@ -42,10 +42,18 @@ CFLAGS   = -I.              -O3 -DNDEBUG -std=c11   -fPIC
 CXXFLAGS = -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC
 LDFLAGS  =
 
-# ref: https://github.com/ggerganov/whisper.cpp/issues/37
-ifneq ($(wildcard /usr/include/musl/*),)
-	CFLAGS += -D_POSIX_SOURCE -D_GNU_SOURCE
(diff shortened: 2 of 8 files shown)


przemoc:
Never ending story. We didn't have FreeBSD coverage before in `whisper.cpp`, so it naturally has to fail now. :)

```
I whisper.cpp build info: 
I UNAME_S:  FreeBSD
I UNAME_P:  amd64
I UNAME_M:  amd64
I CFLAGS:   -I.              -O3 -DNDEBUG -std=c11   -fPIC -D_XOPEN_SOURCE=600 -pthread -mavx -mavx2 -mfma -mf16c
I CXXFLAGS: -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC -D_XOPEN_SOURCE=600 -pthread
I LDFLAGS:  
I CC:       FreeBSD clang version 14.0.5 (https://github.com/llvm/llvm-project.git llvmorg-14.0.5-0-gc12386ae247c)
I CXX:      FreeBSD clang version 14.0.5 (https://github.com/llvm/llvm-project.git llvmorg-14.0.5-0-gc12386ae247c)

cc  -I.              -O3 -DNDEBUG -std=c11   -fPIC -D_XOPEN_SOURCE=600 -pthread -mavx -mavx2 -mfma -mf16c   -c ggml.c -o ggml.o
ggml.c:16438:43: warning: implicit declaration of function 'alloca' is invalid in C99 [-Wimplicit-function-declaration]
    struct ggml_compute_state * workers = alloca(sizeof(struct ggml_compute_state)*n_threads);
                                          ^
ggml.c:16438:33: warning: incompatible integer to pointer conversion initializing 'struct ggml_compute_state *' with an expression of type 'int' [-Wint-conversion]
    struct ggml_compute_state * workers = alloca(sizeof(struct ggml_compute_state)*n_threads);
                                ^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
ggml.c:18[321](https://github.com/ggerganov/whisper.cpp/actions/runs/5621134985/job/15231380392?pr=1129#step:3:324):65: warning: implicit declaration of function 'alloca' is invalid in C99 [-Wimplicit-function-declaration]
    struct ggml_opt_context * opt = (struct ggml_opt_context *) alloca(sizeof(struct ggml_opt_context));
                                                                ^
ggml.c:18321:37: warning: cast to 'struct ggml_opt_context *' from smaller integer type 'int' [-Wint-to-pointer-cast]
    struct ggml_opt_context * opt = (struct ggml_opt_context *) alloca(sizeof(struct ggml_opt_context));
                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
4 warnings generated.
c++ -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC -D_XOPEN_SOURCE=600 -pthread -c whisper.cpp -o whisper.o
c++ -I. -I./examples -O3 -DNDEBUG -std=c++11 -fPIC -D_XOPEN_SOURCE=600 -pthread examples/main/main.cpp examples/common.cpp examples/common-ggml.cpp ggml.o whisper.o -o main 
ld: error: undefined symbol: alloca
>>> referenced by ggml.c
>>>               ggml.o:(ggml_graph_compute)
>>> referenced by ggml.c
>>>               ggml.o:(ggml_opt)
>>> did you mean: allocm
>>> defined in: /lib/libc.so.7
c++: error: linker command failed with exit code 1 (use -v to see invocation)
gmake: *** [Makefile:285: main] Error 1
```

przemoc:
`alloca()` function ([FreeBSD man page](https://man.freebsd.org/cgi/man.cgi?alloca), [Linux man page](https://man7.org/linux/man-pages/man3/alloca.3.html)) is not the best thing to use, but I don't want to refactor stuff in this PR. Will need to dig into this more to understand why it fails now, but it may or may not be today, as I mostly used my quota for side things today. :)

Maybe with minimal FTMs we're adding FreeBSD limits availability of non-POSIX interfaces. Or something like that.

I may need to set up FreeBSD VM to play with it more, maybe during weekend. Haven't touched FreeBSD in many years.

przemoc:
By looking at headers only:
https://github.com/freebsd/freebsd-src/blob/release/13.2.0/include/stdlib.h#L245-L259
https://github.com/freebsd/freebsd-src/blob/release/13.2.0/sys/sys/cdefs.h#L694-L766
it seems that `alloca()` is available in FreeBSD only when `__BSD_VISIBLE` is set to 1. I don't see any cleaner way than adding it explicitly for FreeBSD, as normally `__BSD_VISIBLE` seems to be set to 1 only when no standard-related FTMs are used at all.

So I looked at other popular BSD headers too.

For OpenBSD it seems that `_BSD_SOURCE` should do the job, which looks like a cleaner solution. Just setting `__BSD_VISIBLE` to 1 would be changed to 0 in the presence of `__POSIX_VISIBLE` that is set due to FTM we use, so `_BSD_SOURCE` is better approach. (No, FreeBSD does not recognize `_BSD_SOURCE`.)

And for NetBSD `_NETBSD_SOURCE` seems to be what is needed.

I'll have to add all this to Makefile and CMakeLists.txt with appropriate comments, so that it could be most likely removed if/when `alloca()` will be gone from `ggml.c` (and examples that use it, presently only in `llama.cpp`.

przemoc:
Before updating this PR, I will create yet another PR to simply add support for building on DragonFlyBSD/NetBSD/OpenBSD, then I will rebase this one on top of that with new BSD-related refinements.

przemoc:
@ggerganov, when you'll have some spare time for `whisper.cpp` (as a break from `llama.cpp`, I guess :>), it would be great if you could merge my recent PRs: #1210, #1211, #1212. It will help continuing working on and rebasing current PR (#1129).

przemoc:
BTW There should be no issues with merging #1238 first and then this #1129.

ggerganov:
Will look to merge this after the sync: https://github.com/ggerganov/whisper.cpp/pull/1247

przemoc:
I saw that #1247 got merged.
I rebased the branch, but it's kind of untested locally yet, so theoretically it may be not clean.

przemoc:
Portable applications should employ `sysconf(_SC_PAGESIZE)` instead of `getpagesize()`, so I will create separate PR for `llama.cpp` and `whisper.cpp` to do that, to be merged before this PR will be merged. But tbh, it's unclear to me if macOS supports `sysconf(_SC_PAGESIZE)` (as I don't see it in [apple-oss-distributions/xnu](https://github.com/apple-oss-distributions/xnu), but maybe that's wrong place to look at). I guess checks will tell us the truth...

przemoc:
#1251 has been created.

przemoc:
Rebased after #1251 got merged.

ggerganov (approved):
Thank you for the effort @przemoc !

przemoc:
You're welcome!
Hopefully we won't have new issues (not covered by checks) due to this change.

PR #1130, opened 2023-07-22
allow whisper_full from mel spectrogram - no audio

jbrough:
This change allows a workflow like

`whisper_set_mel_with_state` > `whisper_full_with_state` 

Previously, whisper_full_with_state would scrub `mel.data`, but now preserve it for processing if `n_samples  == 0`.

I think it will be very handy to be able to work with (highly compressed) mel spectrograms directly, especially in a realtime situation.

I've tested this change with streaming spectrograms and works nicely (as it's otherwise fully supported by the internal whisper.cpp apis).


vadi2:
Would you mind giving an example of how to work with this in practice?

jbrough:
@vadi2 sure. I've only just started prototyping real time inference with spectrograms after finishing up a small library to generate them. That's here, and here's the specific function that formats the mel spectrogram data into the format C expects:

https://github.com/wavey-ai/mel-spec/blob/01b74a329d4345d1909593b6a4052ef8fdcf1ca1/src/lib.rs#L142


I'm getting good results with the "jfk" sample sending short spectrograms composited of 256 - 512 fft frames. It will need logic to splice them at appropriate word boundaries, that seems to be the only real issue - and I think this might be easier to manage with spectrograms than detecting attack transients in pcm audio (I know whisper does speech detection, but that still requires sending everything to the model).

I was just trying things out in Rust, calling it with whisper-rs which just provides bindings to the C functions:

```
           let handle = thread::spawn(move || {
                let mut men_filters = mel(16000.0, 400, 80, false, true);
                let n_frames = 256
                let mut frames: Vec<Array2<f64>> = Vec::new();
                let mut i = 0;
                let ctx = WhisperContext::new(
                    "../models/ggml-medium.en.bin",
                )
                .expect("failed to load model");
                let mut state = ctx.create_state().expect("failed to create key");
            
                let mut fft = AudioProcessor::new(fft_size, hop_size);

                while let Ok((idx, chan, pcm_data)) = fft_receiver_clone.recv() {
                    complex_buf = fft.add(&pcm_data);
                    let mel = log_mel_spectrogram(&complex_buf, &mel_filters);
                    frames.push(norm_mel(&mel));
                    i += 1;
                    if i == n_frames {
                        i = 0;
                        let waterfall = interleave_frames(&frames, true); // convenient byte order to render to video/canvas 
                        let whisper = interleave_frames(&frames, false);  // whisper.cpp byetstream order
                        frames.drain(0..);
                        fft_frames_sender_clone
                            .send((false, idx, chan, n_frames, waterfall))
                            .unwrap();
                        state.set_mel(&whisper).unwrap();
                        let mut params = FullParams::new(SamplingStrategy::Greedy { best_of: 0 });

                        params.set_n_threads(6);
                        params.set_single_segment(true);
                        params.set_language(Some("en"));
                        params.set_print_special(false);
                        params.set_print_progress(false);
                        params.set_print_realtime(false);
                        params.set_print_timestamps(false);

                        let empty = vec![0.0; 0];
                        state.full(params, &empty[..]).unwrap()
                        let num_segments = state
                            .full_n_segments()
                            .expect("failed to get number of segments");
                        for i in 0..num_segments {
                            let segment = state
                                .full_get_segment_text(i)
                                .expect("failed to get segment");
                            println!("text: {}\n", segment);
                        }
                    }
                }
            });
```

here's what I got

Screenshot 2023-07-22 at 09 17 25

jbrough:
here's whisper-rs, which provided bindings to whisper_set_mel_with_state - I've had to fork both though and make these small changes to get it to work.

vadi2:
Thank you!

jbrough:
@vadi2 added some more info: https://github.com/wavey-ai/mel-spec/blob/main/README.md

jbrough:
PR got closed automatically on my side due to merge conflicts, reopening.

there's a thorough use case here:

https://github.com/wavey-ai/mel-spec

happy to make any changes that are needed

vadi2:
Unfortunately I don't, would need @ggerganov to weigh in here. I think this a good improvement myself and I'll be having a streaming usecase myself soon as well.

bobqianic (approved):

ggerganov (approved):

ggerganov:
Merged via #1214

Issue #1133, opened 2023-07-25
Windows CLBlast: kernel compile error 


gpapp:
When building under Windows, the kernel build fails on the line endings.

This issue is identical to that of LLama.cpp.

https://github.com/ggerganov/llama.cpp/issues/2065

PR #1134, opened 2023-07-25
Fixing Windows line endings for CLBlast kernel. #1133

gpapp:
Fixes the OpenCL kernel compile issue on Windows, similar to LLama.cpp.

https://github.com/ggerganov/llama.cpp/pull/2086


ggerganov:
Should be fixed now

PR #1142, opened 2023-07-30
Add precalculated values of sin/cos for speeding up FFT

AlexandrGraschenkov:
The code works in the same way as before, except that we don't spend time on sin/cos calculations. This is a real bottleneck for FFT calculations.

System info:
```
Apple M1 Pro, Mac OS Ventura 13.3
n_threads = 4 / 10 | AVX = 0 | AVX2 = 0 | AVX512 = 0 | FMA = 0 | NEON = 1 | ARM_FMA = 1 | F16C = 0 | FP16_VA = 1 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 0 | VSX = 0 | COREML = 0 | OPENVINO = 0 |
```

Testing `base.en` model on: `./main -f samples/a13.wav`
| Version | mel time |
| --- | -- |
| Original | 55.17 ms |
| Cos/sin table | **19.15 ms** |

Testing `base.en` model on:  `./main -f samples/diffusion2023-07-03.wav`
| Version | mel time | 
| --- | -- |
| Original | 2142.07 ms |
| Cos/sin table | **734.42 ms** |

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -2359,7 +2359,7 @@ static void fill_sin_cos_table() {
     for (int i = 0; i < SIN_COS_N_COUNT; i++) {
         double theta = (2*M_PI*i)/SIN_COS_N_COUNT;
         sin_vals[i] = sin(theta);
-        cos_vals[i] = cos(theta);
+        cos_vals[i] = cosf(theta);
     }
     is_filled = true;
 }
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -2347,6 +2347,23 @@ static std::string to_timestamp(int64_t t, bool comma = false) {
     return std::string(buf);
 }
 
+#define SIN_COS_N_COUNT WHISPER_N_FFT
+static float sin_vals[SIN_COS_N_COUNT];
+static float cos_vals[SIN_COS_N_COUNT];
+
+// In FFT, we frequently use sine and cosine operations with the same values. 
+// We can use precalculated values to speed up the process.
+static void fill_sin_cos_table() {
+    static bool is_filled = false;
+    if (is_filled) return;
+    for (int i = 0; i < SIN_COS_N_COUNT; i++) {
+        double theta = (2*M_PI*i)/SIN_COS_N_COUNT;
+        sin_vals[i] = sin(theta);
+        cos_vals[i] = cos(theta);
+    }
+    is_filled = true;
+}
+
 // naive Discrete Fourier Transform
 // input is real-valued
 // output is complex-valued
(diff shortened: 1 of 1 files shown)


bobqianic (approved):

bobqianic:
@AlexandrGraschenkov If you have some free time, could you do me a favor and test how much of an improvement there would be if the optimization in #1148 is combined with your optimization? This is because I've set it to skip FFT calculations and output the results directly when the FFT input is all zeros during the calculation of log Mel. If you have more time, please help me see how much the quality of speech-to-text has improved. Thank you!

bobqianic:
> > This is because I've set it to skip FFT calculations and output the results directly when the FFT input is all zeros during the calculation of log Mel.
> 
> How often do you observe this to be the case? I would guess it never happens for normal audio

After the log Mel correction, we should have `30` seconds of zeros, equating to `480,000` samples or approximately `2,998` frames. The presence of these zeros is due to the zero padding added to the end of the sample before performing calculations. With each FFT taking an estimated `25` microseconds, we anticipate a time saving of `75ms`. Additionally, by removing the if statement from the loop, we can reduce the number of branch predictions.

whisper.cpp
```diff
@@ -2347,22 +2347,40 @@ static std::string to_timestamp(int64_t t, bool comma = false) {
...
+static float cos_vals[SIN_COS_N_COUNT];
+
+// In FFT, we frequently use sine and cosine operations with the same values. 
+// We can use precalculated values to speed up the process.
+static void fill_sin_cos_table() {
+    static bool is_filled = false;
+    if (is_filled) return;
+    for (int i = 0; i < SIN_COS_N_COUNT; i++) {
+        double theta = (2*M_PI*i)/SIN_COS_N_COUNT;
+        sin_vals[i] = sin(theta);
```
bobqianic, line 2361:
```suggestion
        sin_vals[i] = sinf(theta);
```

whisper.cpp
```diff
@@ -2347,22 +2347,40 @@ static std::string to_timestamp(int64_t t, bool comma = false) {
...
+        cos_vals[i] = cos(theta);
```
bobqianic, line 2362:
```suggestion
        cos_vals[i] = cosf(theta);
```

whisper.cpp
bobqianic, line 2361:
```suggestion
        sin_vals[i] = sinf(theta);
```

bobqianic (changes requested):

ggerganov (approved):

PR #1144, opened 2023-07-30
Improvements to vim plugin and LSP server

AustinMroz:
https://github.com/ggerganov/whisper.cpp/assets/4284322/9e88083c-e531-45ea-9780-004740ec7c9e

While I think it'll need about another week of polish to fix a number of outstanding bugs, work is far enough along to make a fancy demonstration video.

The implementation is divided between a vim plugin and a server that communicate over stdout/stdin by way of Language Server Protocol. Other methods like libcall proved nonviable and this option increases the likelihood that future programs or plugins are able to reuse the work.

The server is able to perform both unguided transcription and guided transcription restricted to a client provided list of commands. Timestamps can be provided in client requests so that the user isn't forced to wait for the previous command to finish processing before providing the next, and I plan to work on an improved VAD to reduce the need for pronounced gaps/minimize reprocessing of voice detection on a sliding window.

Most vim actions built of single keys are functional and action names are able to change based on context (i becomes inside when used as part of a motion), but I've still observed some inconsistencies with less frequent commands such as the replace multiple of 'R'


ggerganov:
Very cool work! Looking forward to see this improved further

Mihaiii:
Haha, super cool! Love it!

AustinMroz:
https://github.com/ggerganov/whisper.cpp/assets/4284322/5bba3c76-2ac7-46ab-bc34-6bdde1ff30c6

Progress is reaching the point where it should be stable enough for other people to play with and the recently implemented user commands setting makes for another flashy demo.

Short of others having difficulty running it, I just want to take another pass at formatting, naming (the c++ code isn't an LSP, it's a language server that implements the protocol), figure out cmake, and add some documentation.

thiswillbeyourgithub:
Just chiming in to shout out a related project I heard about : CursorLess It's FOSS and allows to code using your voice.

Here's a video (skip to 1min09) : https://www.youtube.com/watch?v=0ZZb12Qp6-0
Here's a github link : https://github.com/cursorless-dev/cursorless

ggerganov:
@AustinMroz 

Let us know when it is ready for review

AustinMroz:
@ggerganov  I believe it's ready for review.

While the scope here is far greater than that of the llama plugin, I was able to install an older version of vim (that's one day off of yours) with nix and all my test cases are still passing on it.

ggerganov (approved):
I haven't played with this yet, but I think the implementation is very nice and it is a really cool example to add to `whisper.cpp`. Nice work!

yyy33:
Pretty, but it looks like there's some delay

PR #1154, opened 2023-08-05
Minor fixes

csukuangfj:


whisper.cpp
```diff
@@ -82,7 +82,7 @@ static void byteswap_tensor(ggml_tensor * tensor) {
     } while (0)
 #define BYTESWAP_TENSOR(t)       \
     do {                         \
-        byteswap_tensor(tensor); \
+        byteswap_tensor(t); \
```
bobqianic, line 85: Could you clarify what issues this is causing? And why do we need to switch from `tensor` to `t`?

csukuangfj: 1. The current implementation is misleading. The macro accepts an argument, but it is never used.

2. It has not caused errors so far because the place where the macro is used exists a variable with name 'tensor'. Had you changed the variable name, would the code be broken.

ggerganov (approved):

PR #1173, opened 2023-08-11
Fix OpenBLAS detection under Arch Linux

marmistrz:
/usr/include/openblas is used under ARch Linux


marmistrz:
@ggerganov ping

CMakeLists.txt
```diff
@@ -147,15 +147,16 @@ if (WHISPER_BLAS)
             message(WARNING "BLAS library was not found. Environment variable OPENBLAS_PATH not defined.")
         endif ()
     else ()
-        set(BLA_STATIC 1)
-        set(BLA_VENDOR ${WHISPER_BLAS_VENDOR})
+        set(BLAS_STATIC 1)
+        set(BLAS_VENDOR ${WHISPER_BLAS_VENDOR})
```
ggerganov, line 151: Does not look right

https://cmake.org/cmake/help/latest/module/FindBLAS.html

marmistrz: Thanks for catching that, reverted the offending lines.

marmistrz:
Actually, I need to remove `BLA_STATIC` and `BLA_INTEGER_SIZE 8` to have OpenBLAS found under Arch Linux.  The library is shipping only as an `.so` library. For some reason, CMake doesn't detect `/usr/lib/libopenblas_64.so`, only   `/usr/lib/libopenblas.so`,

sscotti:
___________________





Stephen D. Scotti, M.D.

Some Idiots Are Savants


***@***.***
   Mobile:  +1 651 313 0209

Address:  2508 Delaware St SE, Unit 163, Minneapolis, MN 55414







  https://http://www.linkedin.com/in/stephen-d-scotti

  https://github.com/sscotti

  https://idiots-are-savants.com (for fun). https://www.sias.dev (for business)



___________________

The content of this email is confidential and intended for the recipient specified in message only. It is strictly forbidden to share any part of this message with any third party, without a written consent of the sender. If you received this message by mistake, please reply to this message and follow with its deletion, so that we can ensure such a mistake does not occur in the future.












---- On Mon, 21 Aug 2023 02:17:44 +0200 Marcin Mielniczuk ***@***.***> wrote ---





Actually, I need to remove BLA_STATIC and BLA_INTEGER_SIZE 8 to have OpenBLAS found under Arch Linux.  The library is shipping only as an .so library. For some reason, CMake doesn't detect /usr/lib/libopenblas_64.so, only /usr/lib/libopenblas.so,

PR #1204, opened 2023-08-23
whisper.android : migrate from ndk-build to CMake

JunkFood02:
CMake is the default and recommended tool to build native library for an Android app now. Switching to CMake could also make the Android example aligned with the rest of the project.


ggerganov (approved):
Thanks! Waiting for the CI to pass and will merge

PR #1209, opened 2023-08-26
ROCm Port

ardfork:
Similar to https://github.com/ggerganov/llama.cpp/pull/1087 this is a simple HIP port of ggml-cuda.cu.

In the same way build with CMake with `CC=/opt/rocm/llvm/bin/clang CXX=/opt/rocm/llvm/bin/clang++ cmake -DWHISPER_HIPBLAS` or with the Makefile `WHISPER_HIPBLAS=1 make -j`.


ggerganov (approved):

mega-ice:
please check https://github.com/ggerganov/whisper.cpp/issues/1453

PR #1211, opened 2023-08-27
Detect SSSE3

przemoc:


przemoc:
First 2 commits most likely may be ported to `llama.cpp`, I may create PR tomorrow there too, but it will be without detection at make, as it seems to be organized differently there.

ggerganov (approved):

PR #1212, opened 2023-08-27
make : add support for building on DragonFlyBSD/NetBSD/OpenBSD

przemoc:


przemoc:
*BSDs do not have /proc/cpuinfo equivalent, so users of older machines have to remove or comment out locally in `Makefile`:

```
ifeq ($(UNAME_M),amd64)
    CFLAGS += -mavx -mavx2 -mfma -mf16c
endif
```

I plan to attempt addressing it in the future by writing small single source file cpuinfo app, that could be built early in Makefile for *BSDs and be used to guide Makefile what compiler flags can be used to utilize all available instruction sets instead of forcing all of them blindly.

przemoc:
But the plan I mentioned in my previous comment is for the future (other weekend) and different PR.
There is nothing preventing this PR to be merged as-is, as it's not making anything worse than currently is.

Example from running:
`./main -m models/ggml-tiny.en.bin -f samples/jfk.wav`
on 4 *BSD in VMs (`whisper.cpp` built with local Makefile modification as described in [my previous comment](https://github.com/ggerganov/whisper.cpp/pull/1212#issuecomment-1694632043))

image

ggerganov (approved):

PR #1218, opened 2023-08-28
Fix CoreML memleak (fixes #1202) 

### LINKED ISSUE #1202: CoreML + calls to whisper_full result in increased memory usage (apparent leak)
Calls to whisper_full when using CoreML appear to be leaking memory (~5.88mb per call on medium model). Transcription results seem to be fine. This behavior was observed calling the `stream` example. It was not observed when CoreML is disabled.

## Environment

whisper.cpp 21e8c67a4fb6bfc5f1af2a6c00db97b28a0f15b2
Mac Mini, M2, MacOS 13.4

```
$ g++ --version
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.5.0
Thread model: posix

$ coremlc version
1436.100.10 
```

Build/Run
```
make clean
WHISPER_COREML=1 make stream
./stream -m ./models/ggml-medium.bin
```

## Debug Info

Every call to whisper_full seems to be allocating a new MLMultiArray. It isn't reported as a leak, because apparently there is still someone holding a reference to it. Nevertheless, every successive call allocates a new one, without releasing the previous one.

whisper_debug

The screenshot above shows a debug session on `./stream -m ./models/ggml-medium.bin`. These 5.88mb allocations occur when calling `whisper_coreml_encode` and are never freed.

I wonder if this could be related to #797 or #910. Tried to find a solution myself but i have zero experience with ObjC. Any insight is appreciated!

denersc:
Fixes memleak that occurs when calling `whisper_full` with CoreML (#1202)

### DIFF
--- a/coreml/whisper-encoder.mm
+++ b/coreml/whisper-encoder.mm
@@ -53,9 +53,11 @@ void whisper_coreml_encode(
                                            error: nil
     ];
 
-    whisper_encoder_implOutput * outCoreML = [(__bridge id) ctx->data predictionFromLogmel_data:inMultiArray error:nil];
+    @autoreleasepool {
+        whisper_encoder_implOutput * outCoreML = [(__bridge id) ctx->data predictionFromLogmel_data:inMultiArray error:nil];
 
-    memcpy(out, outCoreML.output.dataPointer, outCoreML.output.count * sizeof(float));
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
This helped fix the memory leak in `llama.cpp` Metal, so I guess it works

PR #1220, opened 2023-08-29
ggml : sync (ggml-alloc, GPU, eps, etc.)

ggerganov:
This syncs the latest `ggml` library and will be needed for several new improvements in `whisper.cpp`:

- reduce memory usage via `ggml-alloc`
- add full GPU offloading with CUDA and Metal

Need to check carefully if this accidentally overrides some of the recent commits in `whisper.cpp` because the sync is manual and I could have make a mistake. But overall, there should be no change in the behavior of `whisper.cpp` with these changes


ggerganov:
Would appreciate some help with testing this branch and confirming that it does not change the behavior from `master`

adam-international:
These changes seem to be fine when using normal models, but quantized models produce broken output. For example, the jfk speech outputs only exclamation marks from my testing. I haven't changed any parameters.

PR #1238, opened 2023-09-01
make : improve cpuinfo handling on x86 hosts

przemoc:


przemoc:
Not having access to any macOS machine is such a pain, but I don't plan buying anyone soon.
If there is some kind soul reading this message accidentally and willing to give some macOS (Intel or ARM, ideally both) shell account/access free of charge to random stranger (myself), please let me know via mail. I promise to use it only for open-source related activities and not abuse it.

bobqianic:
> Not having access to any macOS machine is such a pain, but I don't plan buying anyone soon.
> If there is some kind soul reading this message accidentally and willing to give some macOS (Intel or ARM, ideally both) shell account/access free of charge to random stranger (myself), please let me know via mail. I promise to use it only for open-source related activities and not abuse it.

Using the college's electricity and internet to set up a M1 Mac Mini server is feasible, but this will have to wait until October when I return to college.

PR #1243, opened 2023-09-03
Faster `beam_search` sampling

bobqianic:
**PR (v1):** The KV cache update logic in the Master is quite simple and heavy-handed. First, a temporary cache is established, and all the KV cache is stored in this temporary cache. After the beam search is completed, the KV cache is updated based on the results of the beam search, pulling from the temporary cache to replace the old KV cache. If we assume the `beam_size` is `X`, then with the above method we need to read and write the KV cache `2X` times, which is very inefficient. In reality, we often don't need to update the decoder's KV cache (zero-copy), and even if we do need to update, we only need to update a portion and can directly copy without going through the temporary cache (one-copy). Lastly, we can also detect if there are read-write conflicts to reduce the number of times we write to the temporary cache and then copy (two-copy).

**PR (v2):** The calculation of `whisper_sample_token_topk` in Master is highly inefficient. First, it uses `logits_id.clear()` to clear all elements in the vector, and then it uses a loop to extract the logit value from logits. Along with the current index, a new `std::pair` is created and then pushed to the end of `logits_id` using `push_back`. Because the length of logits is determined by `n_vocab`, and `n_vocab` in whisper is `51,865`, each time we calculate `whisper_sample_token_topk` using the method in Master, we need to create (via `logits_id.push_back`) ~~and destroy (via `logits_id.clear`)~~ `51,865` `std::pair` objects, which is highly inefficient. In fact, we can use `logits_id.resize()` to limit the size of `logits_id`. Since the new size is less than or equal to the old capacity, there will be no memory allocation, and we can directly assign values to the old pairs using the index, thus avoiding creation and destruction.

**PR(v3):** Adopted ggerganov's suggestion to make ·kv_bufs· static, avoiding memory allocation and deallocation, thereby significantly improving performance.

**PR(v4):** Added the pointer swapping function, which slightly improved performance.

**PR(v5):** Updated the logic for determining `two-copy`, reducing unnecessary memory copying, slightly improving performance.

This PR optimizes both the KV cache update logic and the calculation of `whisper_sample_token_topk`, significantly reducing memory read-write operations, thereby saving a lot of time. 

The average sample time has decreased from **`1.56 ms/run`** to **`0.69 ms/run`**, reducing the computational time by approximately **`~55%`**. @ggerganov 

**Master:** `i7-12700H` `diffusion2023-07-03.wav` `ggml-model-whisper-base.bin`
| `-bs 1` `-bo 1` | `-bs 2` `-bo 2` | `-bs 3` `-bo 3` | `-bs 4` `-bo 4` | `-bs 5` `-bo 5` |
| :-----: | :----: | :----: | :----: | :----: |
| 0.57 ms/run | 1.56 ms/run | 1.55 ms/run | 1.58 ms/run | 1.55 ms/run |

<details> <summary>Older Versions</summary>

**This PR (v1):** `i7-12700H` `diffusion2023-07-03.wav` `ggml-model-whisper-base.bin`
| `-bs 1` `-bo 1` | `-bs 2` `-bo 2` | `-bs 3` `-bo 3` | `-bs 4` `-bo 4` | `-bs 5` `-bo 5` |
| :-----: | :----: | :----: | :----: | :----: |
| 0.57 ms/run | 0.96 ms/run | 1.01 ms/run | 1.07 ms/run | 1.14 ms/run |

**This PR (v2):** `i7-12700H` `diffusion2023-07-03.wav` `ggml-model-whisper-base.bin`
| `-bs 1` `-bo 1` | `-bs 2` `-bo 2` | `-bs 3` `-bo 3` | `-bs 4` `-bo 4` | `-bs 5` `-bo 5` |
| :-----: | :----: | :----: | :----: | :----: |
| 0.57 ms/run | 0.76 ms/run | 0.81 ms/run | 0.87 ms/run | 0.94 ms/run |

**This PR (v3):** `i7-12700H` `diffusion2023-07-03.wav` `ggml-model-whisper-base.bin`
| `-bs 1` `-bo 1` | `-bs 2` `-bo 2` | `-bs 3` `-bo 3` | `-bs 4` `-bo 4` | `-bs 5` `-bo 5` |
| :-----: | :----: | :----: | :----: | :----: |
| 0.57 ms/run | 0.69 ms/run | 0.72 ms/run | 0.74 ms/run | 0.77 ms/run |

**This PR (v4):** `i7-12700H` `diffusion2023-07-03.wav` `ggml-model-whisper-base.bin`
| `-bs 1` `-bo 1` | `-bs 2` `-bo 2` | `-bs 3` `-bo 3` | `-bs 4` `-bo 4` | `-bs 5` `-bo 5` |
| :-----: | :----: | :----: | :----: | :----: |
| 0.57 ms/run | 0.67 ms/run | 0.70 ms/run | 0.72 ms/run | 0.75 ms/run |

</details>

**This PR (v5):** `i7-12700H` `diffusion2023-07-03.wav` `ggml-model-whisper-base.bin`
| `-bs 1` `-bo 1` | `-bs 2` `-bo 2` | `-bs 3` `-bo 3` | `-bs 4` `-bo 4` | `-bs 5` `-bo 5` |
| :-----: | :----: | :----: | :----: | :----: |
| 0.57 ms/run | 0.64 ms/run | 0.68 ms/run | 0.70 ms/run | 0.73 ms/run |

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -4132,15 +4132,16 @@ static bool whisper_kv_swap_fast(
         // since we modify data sequentially, we only consider decoder indices after current index
         for (int j = i + 1; j < size; j++) {
             if (i == view[j]) {
-                two_copy.insert(i);
-                is_one_copy = false;
                 // detect symmetric diagram
                 if (j == view[i]) {
                     p_swap_set.insert(i);
                     p_swap_set.insert(j);
                     p_swap_vec.emplace_back(i, j);
+                } else {
+                    two_copy.insert(i);
+                    is_one_copy = false;
+                    break;
                 }
-                break;
             }
         }
         if (is_one_copy) {
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -4008,9 +4008,12 @@ static std::vector<whisper_token_data> whisper_sample_token_topk(
 
     {
         using pair_type = std::remove_reference<decltype(logits_id)>::type::value_type;
-        std::partial_sort(logits_id.begin(),logits_id.begin() + k, logits_id.end(),[](const pair_type & a, const pair_type & b) {
-                    return a.first > b.first;
-                });
+        std::partial_sort(
+                logits_id.begin(),
+                logits_id.begin() + k, logits_id.end(),
+                [](const pair_type & a, const pair_type & b) {
+            return a.first > b.first;
+        });
     }
 
     std::vector<whisper_token_data> result;
@@ -4110,7 +4113,8 @@ static bool whisper_kv_swap_fast(
                    std::vector<int> & view,
                     whisper_decoder   src[],
                 std::vector<kv_buf> & kv_swap_bufs,
-                          const int & size) {
+                          const int & n_decoders) {
+    WHISPER_PRINT_DEBUG("%s: n_decoders %d\n", __func__, n_decoders);
 
     // (decoder->buffer->decoder or decoder->buffer + decoder->decoder)
     std::set<int> two_copy; // decoder indices require two copies to safely modify KV caches
@@ -4121,16 +4125,18 @@ static bool whisper_kv_swap_fast(
     // (decoder<->decoder)
     std::set<int> p_swap_set; // decoder indices able to swap KV-cache pointers
     std::vector<whisper_pair<int, int>> p_swap_vec;
-    p_swap_vec.reserve(size);
+    p_swap_vec.reserve(n_decoders);
 
     // see https://github.com/ggerganov/whisper.cpp/wiki
-    for (int i = 0; i < size; i++) {
+    for (int i = 0; i < n_decoders; i++) {
         // zero-copy (no modification)
-        if (i == view[i] || view[i] < 0) {continue;}
+        if (i == view[i] || view[i] < 0) {
+            continue;
+        }
 
         bool is_one_copy = true;
         // since we modify data sequentially, we only consider decoder indices after current index
-        for (int j = i + 1; j < size; j++) {
+        for (int j = i + 1; j < n_decoders; j++) {
             if (i == view[j]) {
                 // detect symmetric diagram
                 if (j == view[i]) {
@@ -4149,15 +4155,16 @@ static bool whisper_kv_swap_fast(
         }
     }
 
-    kv_swap_bufs.resize(size);
+    kv_swap_bufs.resize(n_decoders);
 
-    for (int i = 0; i < size; i++) {
+    for (int i = 0; i < n_decoders; i++) {
         kv_swap_bufs[i].k.resize(ggml_nbytes(src[i].kv_self.k));
         kv_swap_bufs[i].v.resize(ggml_nbytes(src[i].kv_self.v));
     }
 
     for (auto & i : two_copy) {
         // make a copy of KV caches
+        WHISPER_PRINT_DEBUG("%s: store KV cache into swap: idx %d\n", __func__, i);
         memcpy(kv_swap_bufs[i].k.data(), src[i].kv_self.k->data, kv_swap_bufs[i].k.size());
         memcpy(kv_swap_bufs[i].v.data(), src[i].kv_self.v->data, kv_swap_bufs[i].v.size());
     }
@@ -4165,14 +4172,18 @@ static bool whisper_kv_swap_fast(
     // since two-copy decoder KV caches are protected by kv_swap_bufs, modify them first
     for (auto & i : two_copy) {
         // skip the decoder indices that require pointer swapping
-        if (p_swap_set.find(i) != p_swap_set.end()) {continue;}
+        if (p_swap_set.find(i) != p_swap_set.end()) {
+            continue;
+        }
 
         if (two_copy.find(view[i]) != two_copy.end()) {
             // modify KV caches of decoder using data from kv_swap_bufs
+            WHISPER_PRINT_DEBUG("%s: two-copy decoder using   swap buffers: swap[%d] -> %d\n", __func__, view[i], i);
             memcpy(src[i].kv_self.k->data, kv_swap_bufs[view[i]].k.data(), kv_swap_bufs[view[i]].k.size());
             memcpy(src[i].kv_self.v->data, kv_swap_bufs[view[i]].v.data(), kv_swap_bufs[view[i]].v.size());
         } else {
             // modify KV caches of decoder using data from correspond decoder KV caches directly
+            WHISPER_PRINT_DEBUG("%s: two-copy decoder without swap buffers:      %d  -> %d\n", __func__, view[i], i);
             memcpy(src[i].kv_self.k->data, src[view[i]].kv_self.k->data, ggml_nbytes(src[view[i]].kv_self.k));
             memcpy(src[i].kv_self.v->data, src[view[i]].kv_self.v->data, ggml_nbytes(src[view[i]].kv_self.v));
         }
@@ -4181,14 +4192,18 @@ static bool whisper_kv_swap_fast(
     // then modify one-copy decoder KV caches
     for (auto & i : one_copy) {
         // skip the decoder indices that require pointer swapping
-        if (p_swap_set.find(i) != p_swap_set.end()) {continue;}
+        if (p_swap_set.find(i) != p_swap_set.end()) {
+            continue;
+        }
 
         if (two_copy.find(view[i]) != two_copy.end()) {
             // modify KV caches of decoder using data from kv_swap_bufs
+            WHISPER_PRINT_DEBUG("%s: one-copy decoder using   swap buffers: swap[%d] -> %d\n", __func__, view[i], i);
             memcpy(src[i].kv_self.k->data, kv_swap_bufs[view[i]].k.data(), kv_swap_bufs[view[i]].k.size());
             memcpy(src[i].kv_self.v->data, kv_swap_bufs[view[i]].v.data(), kv_swap_bufs[view[i]].v.size());
         } else {
             // modify KV caches of decoder using data from correspond decoder KV caches directly
+            WHISPER_PRINT_DEBUG("%s: one-copy decoder without swap buffers:      %d  -> %d\n", __func__, view[i], i);
             memcpy(src[i].kv_self.k->data, src[view[i]].kv_self.k->data, ggml_nbytes(src[view[i]].kv_self.k));
             memcpy(src[i].kv_self.v->data, src[view[i]].kv_self.v->data, ggml_nbytes(src[view[i]].kv_self.v));
         }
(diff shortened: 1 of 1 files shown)


whisper.cpp
```diff
@@ -23,6 +23,8 @@
 #include <vector>
 #include <regex>
 #include <random>
+#include <unordered_set>
+#include <iostream>
```
bobqianic, line 27:
```suggestion
```

whisper.cpp
```diff
@@ -4082,6 +4097,79 @@ static void whisper_sequence_score(
     }
 }
 
+static bool whisper_fast_swap_KV(std::vector<int> & view, whisper_decoder src[], int size) {
+    // (decoder->buffer->decoder or decoder->buffer + decoder->decoder)
+    std::unordered_set<int> two_copy; // decoder indices requires two copies to safely modify KV caches
```
bobqianic, line 4102:
```suggestion
    std::unordered_set<int> two_copy; // decoder indices require two copies to safely modify KV caches
```

whisper.cpp
```diff
@@ -4082,6 +4097,79 @@ static void whisper_sequence_score(
...
+    two_copy.reserve(size);
+
+    // (buffer->decoder or decoder->decoder)
+    std::unordered_set<int> one_copy; // decoder indices requires one copies to safely modify KV caches
```
bobqianic, line 4106:
```suggestion
    std::unordered_set<int> one_copy; // decoder indices require one copy to safely modify KV caches
```

whisper.cpp
```diff
@@ -4082,6 +4096,79 @@ static void whisper_sequence_score(
...
+    }
+
+    // beam-search helpers
+    struct kv_buf {
+        std::vector<uint8_t> k;
+        std::vector<uint8_t> v;
+    };
+
+    std::vector<kv_buf> kv_bufs;
+    kv_bufs.resize(size);
```
ggerganov, line 4133: Can we add some simple "kv-swap-work-state" to avoid deallocating and allocating these temporary buffers?
If it makes the implementation more complicated, then we can leave it like this

whisper.cpp
```diff
@@ -4082,6 +4096,79 @@ static void whisper_sequence_score(
     }
 }
 
+static bool whisper_fast_swap_KV(std::vector<int> & view, whisper_decoder src[], int size) {
```
ggerganov, line 4099:
```suggestion
static bool whisper_kv_swap_fast(std::vector<int> & view, whisper_decoder src[], int size) {
```

whisper.cpp
```diff
@@ -4082,6 +4096,79 @@ static void whisper_sequence_score(
...
+    // beam-search helpers
+    struct kv_buf {
+        std::vector<uint8_t> k;
+        std::vector<uint8_t> v;
+    };
+
+    std::vector<kv_buf> kv_bufs;
+    kv_bufs.resize(size);
+
+    for (auto &i : two_copy) {
```
ggerganov, line 4135: Respect the coding style:

```suggestion
    for (auto & i : two_copy) {
```

Fix in other places too

whisper.cpp
```diff
@@ -4465,6 +4529,7 @@ int whisper_full_with_state(
                     });
 
                     uint32_t cur_c = 0;
+                    std::vector<int> decoder_index(n_decoders_cur, -1);
```
ggerganov, line 4532:
```suggestion
                    std::vector<int> decoder_idx(n_decoders_cur, -1);
```

whisper.cpp
```diff
@@ -4401,23 +4480,8 @@ int whisper_full_with_state(
...
-                        }
-
-                        kv_bufs[j].k.resize(ggml_nbytes(decoder.kv_self.k));
-                        kv_bufs[j].v.resize(ggml_nbytes(decoder.kv_self.v));
-
-                        memcpy(kv_bufs[j].k.data(), decoder.kv_self.k->data, kv_bufs[j].k.size());
-                        memcpy(kv_bufs[j].v.data(), decoder.kv_self.v->data, kv_bufs[j].v.size());
-                    }
-
+                    // clear beam candidates
```
ggerganov, line 4484: This comment is not needed

ggerganov (changes requested):
Clever implementation!

Some minor comments and then I will take a more deeper look + testing to make sure this works correctly. But I understand the idea and agree this would be faster

ggerganov:
> First, it uses logits_id.clear() to clear all elements in the vector, and then it uses a loop to extract the logit value from logits.

This is not slower than the new approach, because all that `clear()` does is change the internal elements counter to 0. The memory of the vector remains allocated. But anyway, the new implementation is also fine

bobqianic:
> > First, it uses logits_id.clear() to clear all elements in the vector, and then it uses a loop to extract the logit value from logits.
> 
> This is not slower than the new approach, because all that `clear()` does is change the internal elements counter to 0. The memory of the vector remains allocated. But anyway, the new implementation is also fine

You're right, that's my bad. I misunderstood what the `clear()` function does.

whisper.cpp
bobqianic: I ran some quick tests, and after adding the `static` keyword, the sample time for `beam_size` of `5` dropped to just 
`0.77 ms / run`.

bobqianic:
In certain scenarios, it might be more efficient to just swap the pointers instead of actually moving the memory around. For instance, let's say we only have two decoders engaged in beam search. If we need to switch the KV caches between Decoder A and Decoder B, a simple pointer swap would do the trick. Just like `std:swap`

bobqianic:
I've now visualized the logic for updating the KV cache across different versions, as shown in the figure. I've made the changes in the new version based on the suggestions you provided earlier. If there aren't any bugs, it's ready to be merged. Thanks. @ggerganov 
### Master:
image

### PR(v1):
image

### PR(v4):
image

whisper.cpp
```diff
@@ -4082,6 +4096,109 @@ static void whisper_sequence_score(
...
+    }
+
+    // beam-search helpers
+    struct kv_buf {
+        std::vector<uint8_t> k;
+        std::vector<uint8_t> v;
+    };
+
+    static std::vector<kv_buf> kv_bufs(size);
+    static bool is_resized = false;
```
ggerganov, line 4145: This is not OK - different whisper context will now share the same KV buffers, so this breaks parallel usage.
These buffers have to become part of `whisper_state` instead.

Also, no need for `is_resized` flag. Just always resize. The resize will be noop if the new size is small or equal to the old size.

bobqianic: > This is not OK - different whisper context will now share the same KV buffers, so this breaks parallel usage.

Could you clarify this a bit more, please? I don't fully understand what you mean.

> Also, no need for is_resized flag. Just always resize. The resize will be noop if the new size is small or equal to the old size.

Agree.

whisper.cpp
```diff
@@ -4082,6 +4096,109 @@ static void whisper_sequence_score(
...
+            memcpy(src[i].kv_self.k->data, src[view[i]].kv_self.k->data, ggml_nbytes(src[view[i]].kv_self.k));
+            memcpy(src[i].kv_self.v->data, src[view[i]].kv_self.v->data, ggml_nbytes(src[view[i]].kv_self.v));
+        }
+    }
+
+    // swap the pointers
+    for (auto & i : p_swap_vec) {
+        std::swap(src[i.first].kv_self.k->data, src[i.second].kv_self.k->data);
+        std::swap(src[i.first].kv_self.v->data, src[i.second].kv_self.v->data);
+    }
```
ggerganov, line 4197: Clever!

We should swap the entire `kv_self` struct so that things remain consistent:

```c++
std::swap(src[i.first].kv_self, src[i.second].kv_self);
```

Otherwise, we can end up in situation where `decoder[0].kv_self.k->data` points into `decoder[1].kv_self.buf`, which is confusing and error prone.

ggerganov:
The test program that you made is very useful. You should try to add it as a unit test in `tests`

It might be a good idea to create a wiki page with your idea and plots from this PR. Or at the very least, you can just link to this PR. But the goal is this information to remain available and easy to find in the future.

You can also add a wiki entry for your previous PR about matching OpenAI input. 
I've sent you a collaborator invite

whisper.cpp
```diff
@@ -601,7 +603,7 @@ struct whisper_sequence {
 
 // TAGS: WHISPER_DECODER_INIT
 struct whisper_decoder {
-    // each decoders keeps its own KV-cache
+    // each decoders keep its own KV-cache
```
bobqianic, line 606:
```suggestion
    // each decoder keeps its own KV-cache
```

bobqianic:
I apologize for the late update. I was very busy yesterday and didn't have time to handle it. I've just updated the code.

whisper.cpp
```diff
@@ -4082,6 +4106,101 @@ static void whisper_sequence_score(
...
+        }
+    }
+
+    // then modify one-copy decoder KV caches
+    for (auto & i : one_copy) {
+        // skip the decoder indices that require pointer swapping
+        if (p_swap_set.find(i) != p_swap_set.end()) {continue;}
+
+        if (two_copy.find(view[i]) != two_copy.end()) {
+            // modify KV caches of decoder using data from kv_bufs
```
bobqianic, line 4186:
```suggestion
            // modify KV caches of decoder using data from kv_swap_bufs
```

whisper.cpp
```diff
@@ -4082,6 +4106,101 @@ static void whisper_sequence_score(
...
+        memcpy(kv_swap_bufs[i].v.data(), src[i].kv_self.v->data, kv_swap_bufs[i].v.size());
+    }
+
+    // since two-copy decoder KV caches are protected by kv_bufs, modify them first
+    for (auto & i : two_copy) {
+        // skip the decoder indices that require pointer swapping
+        if (p_swap_set.find(i) != p_swap_set.end()) {continue;}
+
+        if (two_copy.find(view[i]) != two_copy.end()) {
+            // modify KV caches of decoder using data from kv_bufs
```
bobqianic, line 4170:
```suggestion
            // modify KV caches of decoder using data from kv_swap_bufs
```

whisper.cpp
```diff
@@ -4082,6 +4106,101 @@ static void whisper_sequence_score(
...
+        kv_swap_bufs[i].v.resize(ggml_nbytes(src[i].kv_self.v));
+    }
+
+    for (auto & i : two_copy) {
+        // make a copy of KV caches
+        memcpy(kv_swap_bufs[i].k.data(), src[i].kv_self.k->data, kv_swap_bufs[i].k.size());
+        memcpy(kv_swap_bufs[i].v.data(), src[i].kv_self.v->data, kv_swap_bufs[i].v.size());
+    }
+
+    // since two-copy decoder KV caches are protected by kv_bufs, modify them first
```
bobqianic, line 4164:
```suggestion
    // since two-copy decoder KV caches are protected by kv_swap_bufs, modify them first
```

ggerganov (approved):

PR #1261, opened 2023-09-08
Dynamic selection of extended instructions sets at runtime for x86 architecture

didzis:
This PR aims to solve the problem of distributing a single, highly performant binary executable to x86 target machines by dynamically identifying the available extended instruction sets at runtime. With this approach, there's no need to compromise between delivering optimal performance on each target machine (having to select only the common shared instruction sets) and ensuring broad distribution across as many target machines as possible without encountering illegal instruction errors.

Unfortunately, due to some limitations, the solution is not very clean. Specifically, there are copies of the same functionality for different instruction sets, as well as combinations of instruction sets, to achieve the best performance in the context of whisper.cpp.

However, I believe this might offer significant practical value for some. Therefore, I've submitted this PR for your consideration.


Macoron:
I didn't review the code implementation, but the feature is really important for shipping build apps for end-customers. Great work, hope it will get merged!

z11h:
Any update on getting this merged?

ngxson:
Hi, your implementation of `cpu_features.c` looks very good. Do you need help to finish this PR?

PR #1270, opened 2023-09-10
whisper : Metal and ggml-alloc support

ggerganov:
This PR adds Metal support for full GPU inference on Apple Silicon.
It also optimizes memory usage.

- Base model running on M2 Ultra using Metal:

https://github.com/ggerganov/whisper.cpp/assets/1991296/c82e8f86-60dc-49f2-b048-d2fdbd6b5225

- Medium model running on M2 Ultra using Metal:

https://github.com/ggerganov/whisper.cpp/assets/1991296/5073177f-1025-4ec1-8cd2-b6f8d1adb1ff

### Usage:

- Full Metal inference (all processing is on the GPU)

```bash
make clean
make -j && ./main -m ./models/ggml-base.en.bin -f ./samples/gb0.wav
```

| CPU | OS | Config | Model | Th | Load [ms] | Encode [ms] | Commit |
| --- | -- | ------ | ----- | ------- | --------- | ----------- | ------ |
| M2 Ultra | MacOS 13.5.1 |  Metal | tiny   |   8 |   16.69 |    1.42 |    2.92 | f365543 |
| M2 Ultra | MacOS 13.5.1 |  Metal | base   |   8 |   25.88 |    2.03 |    4.57 | f365543 |
| M2 Ultra | MacOS 13.5.1 |  Metal | small  |   8 |   62.30 |    4.01 |   11.80 | f365543 |
| M2 Ultra | MacOS 13.5.1 |  Metal | medium |   8 |  160.75 |    8.26 |   29.41 | f365543 |
| M2 Ultra | MacOS 13.5.1 |  Metal | large  |   8 |  268.36 |   12.55 |   52.22 | f365543 |

- CoreML Encoder + Metal Decoder

```bash
make clean
WHISPER_COREML=1 make -j && ./main -m ./models/ggml-base.en.bin -f ./samples/gb0.wav
```

| CPU | OS | Config | Model | Th | Load [ms] | Encode [ms] | Commit |
| --- | -- | ------ | ----- | ------- | --------- | ----------- | ------ |
| M2 Ultra | MacOS 13.5.1 | Core ML (ANE) | tiny   |    8 |   24.41 |    1.42 |    2.91 | f365543 |
| M2 Ultra | MacOS 13.5.1 | Core ML (ANE) | base   |    8 |   46.50 |    2.02 |    4.72 | f365543 |
| M2 Ultra | MacOS 13.5.1 | Core ML (ANE) | small  |    8 |  137.58 |    3.99 |   11.91 | f365543 |
| M2 Ultra | MacOS 13.5.1 | Core ML (ANE) | medium |    8 |  677.31 |    8.25 |   29.51 | f365543 |
| M2 Ultra | MacOS 13.5.1 | Core ML (ANE) | large  |    8 | 1823.94 |   12.74 |   52.40 | f365543 |

| CPU | OS | Config | Model | Th | Load [ms] | Encode [ms] | Commit |
| --- | -- | ------ | ----- | ------- | --------- | ----------- | ------ |
| M2 Ultra | MacOS 13.5.1 | Core ML (GPU) | tiny   |  8 |    6.72 |    1.42 |    3.15 | f365543 |
| M2 Ultra | MacOS 13.5.1 | Core ML (GPU) | base   |  8 |   11.31 |    2.03 |    4.42 | f365543 |
| M2 Ultra | MacOS 13.5.1 | Core ML (GPU) | small  |  8 |   32.81 |    4.00 |   11.79 | f365543 |
| M2 Ultra | MacOS 13.5.1 | Core ML (GPU) | medium |  8 |  103.47 |    8.22 |   29.38 | f365543 |
| M2 Ultra | MacOS 13.5.1 | Core ML (GPU) | large  |  8 |  185.71 |   12.48 |   52.30 | f365543 |

### TODOs

- [x] merge https://github.com/ggerganov/whisper.spm branch `ggml-alloc` when ready
- [x] fix CI
- [x] fix `ggml_nbytes()`
- [ ] Can't figure out how to make the CMake build look for `ggml-metal.metal` in the `bin` folder

### DIFF
--- a/ggml-alloc.c
+++ b/ggml-alloc.c
@@ -342,8 +342,8 @@ static void free_vmem(void * base_addr, size_t size) {
 
 // allocate uncommitted virtual memory to measure the size of the graph
 static void alloc_measure_vmem(void ** base_addr, size_t * size) {
-    // 1TB for 64-bit, 1GB for 32-bit
-    *size = sizeof(void *) == 4 ? 1ULL<<30 : 1ULL<<40;
+    // 128GB for 64-bit, 1GB for 32-bit
+    *size = sizeof(void *) == 4 ? 1ULL<<30 : 1ULL<<37;
     do {
         *base_addr = alloc_vmem(*size);
         if (*base_addr != NULL) {
--- a/ggml-alloc.c
+++ b/ggml-alloc.c
@@ -6,6 +6,26 @@
 #include <stdlib.h>
 #include <string.h>
 
+#ifdef __has_include
+    #if __has_include(<unistd.h>)
+        #include <unistd.h>
+        #if defined(_POSIX_MAPPED_FILES)
+            #include <sys/types.h>
+            #include <sys/mman.h>
+        #endif
+    #endif
+#endif
+
+#if defined(_WIN32)
+    #define WIN32_LEAN_AND_MEAN
+    #ifndef NOMINMAX
+        #define NOMINMAX
+    #endif
+    #include <windows.h>
+    #include <memoryapi.h>
+#endif
+
+
 #define UNUSED(x) (void)(x)
 #define MAX(a, b) ((a) > (b) ? (a) : (b))
 #define GGML_MAX_CONCUR (2*GGML_MAX_NODES)
@@ -99,15 +119,28 @@ static void remove_allocated_tensor(struct ggml_allocr * alloc, struct ggml_tens
 }
 #endif
 
-
-static size_t ggml_allocator_get_alloc_size(struct ggml_allocr * alloc, struct ggml_tensor * tensor) {
+static size_t ggml_allocr_get_alloc_size(struct ggml_allocr * alloc, struct ggml_tensor * tensor) {
     return ggml_nbytes(tensor);
 
     UNUSED(alloc);
 }
 
+// check if a tensor is allocated by this buffer
+static bool ggml_allocr_is_own(struct ggml_allocr * alloc, const struct ggml_tensor * tensor) {
+    void * ptr = tensor->data;
+    return ptr >= alloc->data && (char *)ptr < (char *)alloc->data + alloc->max_size;
+}
+
+static bool ggml_is_view(struct ggml_tensor * t) {
+    return t->view_src != NULL;
+}
+
 void ggml_allocr_alloc(struct ggml_allocr * alloc, struct ggml_tensor * tensor) {
-    size_t size = ggml_allocator_get_alloc_size(alloc, tensor);
+#ifdef GGML_ALLOCATOR_DEBUG
+    GGML_ASSERT(!ggml_is_view(tensor)); // views generally get data pointer from one of their sources
+    GGML_ASSERT(tensor->data == NULL); // avoid allocating tensor which already has memory allocated
+#endif
+    size_t size = ggml_allocr_get_alloc_size(alloc, tensor);
     size = aligned_offset(NULL, size, alloc->alignment);
 
     AT_PRINTF("%s: allocating %s (%zu bytes) - ", __func__, tensor->name, size);
@@ -131,14 +164,14 @@ void ggml_allocr_alloc(struct ggml_allocr * alloc, struct ggml_tensor * tensor) 
     if (best_fit_block == -1) {
         // the last block is our last resort
         struct free_block * block = &alloc->free_blocks[alloc->n_free_blocks - 1];
+        max_avail = MAX(max_avail, block->size);
         if (block->size >= size) {
             best_fit_block = alloc->n_free_blocks - 1;
-            max_avail = MAX(max_avail, block->size);
         } else {
             fprintf(stderr, "%s: not enough space in the buffer (needed %zu, largest block available %zu)\n",
                     __func__, size, max_avail);
             GGML_ASSERT(!"not enough space in the buffer");
-        return;
+            return;
         }
     }
     struct free_block * block = &alloc->free_blocks[best_fit_block];
@@ -173,17 +206,17 @@ void ggml_allocr_alloc(struct ggml_allocr * alloc, struct ggml_tensor * tensor) 
 }
 
 // this is a very naive implementation, but for our case the number of free blocks should be very small
-static void ggml_allocator_free_tensor(struct ggml_allocr * alloc, struct ggml_tensor * tensor) {
+static void ggml_allocr_free_tensor(struct ggml_allocr * alloc, struct ggml_tensor * tensor) {
     void * ptr = tensor->data;
 
-    if (ptr < alloc->data || (char*)ptr >= (char*)alloc->data + alloc->max_size) {
+    if (ggml_allocr_is_own(alloc, tensor) == false) {
         // the tensor was not allocated in this buffer
         // this can happen because the graph allocator will try to free weights and other tensors from different buffers
         // the easiest way to deal with this is just to ignore it
         return;
     }
 
-    size_t size = ggml_allocator_get_alloc_size(alloc, tensor);
+    size_t size = ggml_allocr_get_alloc_size(alloc, tensor);
     size = aligned_offset(NULL, size, alloc->alignment);
     AT_PRINTF("%s: freeing %s (%zu bytes) - n_free_blocks = %d\n", __func__, tensor->name, size, alloc->n_free_blocks);
 
@@ -277,17 +310,68 @@ struct ggml_allocr * ggml_allocr_new(void * data, size_t size, size_t alignment)
     return alloc;
 }
 
-// address and size of the buffer when measuring
-// it needs to be large enough to fit all the tensors, but it cannot overlap with other existing buffers
-static void * const MEASURE_BASE_ADDR = (void *) 0x1000;
-static const size_t MEASURE_MAX_SIZE  = 1ULL<<40; // 1 TB
+// OS specific functions to allocate and free uncommitted virtual memory
+static void * alloc_vmem(size_t size) {
+#if defined(_WIN32)
+    return VirtualAlloc(NULL, size, MEM_RESERVE, PAGE_NOACCESS);
+#elif defined(_POSIX_MAPPED_FILES)
+    void * ptr = mmap(NULL, size, PROT_NONE, MAP_PRIVATE | MAP_ANON, -1, 0);
+    if (ptr == MAP_FAILED) {
+        return NULL;
+    }
+    return ptr;
+#else
+    // use a fixed address for other platforms
+    uintptr_t base_addr = (uintptr_t)-size - 0x100;
+    return (void *)base_addr;
+#endif
+}
+
+static void free_vmem(void * base_addr, size_t size) {
+#if defined(_WIN32)
+    VirtualFree(base_addr, 0, MEM_RELEASE);
+    UNUSED(size);
+#elif defined(_POSIX_MAPPED_FILES)
+    munmap(base_addr, size);
+#else
+    // nothing to do
+    UNUSED(base_addr);
+    UNUSED(size);
+#endif
+}
+
+// allocate uncommitted virtual memory to measure the size of the graph
+static void alloc_measure_vmem(void ** base_addr, size_t * size) {
+    // 1TB for 64-bit, 1GB for 32-bit
+    *size = sizeof(void *) == 4 ? 1ULL<<30 : 1ULL<<40;
+    do {
+        *base_addr = alloc_vmem(*size);
+        if (*base_addr != NULL) {
+            AT_PRINTF("allocated %.2f GB of virtual memory for measure buffer at %p\n", *size / 1024.0 / 1024.0 / 1024.0, *base_addr);
+            return;
+        }
+        // try again with half the size
+        *size /= 2;
+    } while (*size > 0);
+
+    GGML_ASSERT(!"failed to allocate virtual memory for measure buffer");
+}
+
+static void free_measure_vmem(void * base_addr, size_t size) {
+    free_vmem(base_addr, size);
+}
 
 struct ggml_allocr * ggml_allocr_new_measure(size_t alignment) {
     struct ggml_allocr * alloc = (struct ggml_allocr *)malloc(sizeof(struct ggml_allocr) /* + n_free_blocks * sizeof(struct free_block) */);
 
+    void * base_addr;
+    size_t size;
+
+    alloc_measure_vmem(&base_addr, &size);
+
     *alloc = (struct ggml_allocr){
-        /*.data          = */ MEASURE_BASE_ADDR,
-        /*.size          = */ MEASURE_MAX_SIZE,
+        /*.data          = */ base_addr,
+        /*.size          = */ size,
         /*.alignment     = */ alignment,
         /*.n_free_blocks = */ 0,
         /*.free_blocks   = */ {{0}},
@@ -307,6 +391,9 @@ struct ggml_allocr * ggml_allocr_new_measure(size_t alignment) {
 }
 
 void ggml_allocr_free(struct ggml_allocr * alloc) {
+    if (alloc->measure) {
+        free_measure_vmem(alloc->data, alloc->size);
+    }
     free(alloc);
 }
 
@@ -315,11 +402,6 @@ bool ggml_allocr_is_measure(struct ggml_allocr * alloc) {
 }
 
 //////////// compute graph allocator
-
-static bool ggml_is_view(struct ggml_tensor * t) {
-    return t->op == GGML_OP_RESHAPE || t->op == GGML_OP_VIEW || t->op == GGML_OP_TRANSPOSE ||
-           t->op == GGML_OP_PERMUTE || t->op == GGML_OP_CPY;
-}
 
 static bool ggml_are_same_layout(const struct ggml_tensor * a, const struct ggml_tensor * b) {
     if (a->type != b->type) {
@@ -336,28 +418,6 @@ static bool ggml_are_same_layout(const struct ggml_tensor * a, const struct ggml
     return true;
 }
 
-static struct ggml_tensor * get_view_parent(struct ggml_tensor * t) {
-    switch (t->op) {
-        case GGML_OP_PERMUTE:
-        case GGML_OP_RESHAPE:
-        case GGML_OP_TRANSPOSE:
-        case GGML_OP_VIEW:
-            return t->src[0];
-        case GGML_OP_CPY:
-            return t->src[1];
-        default:
-            return NULL;
-    }
-}
-
-static struct ggml_tensor * get_view_source(struct ggml_tensor * t) {
-    struct ggml_tensor * parent = t;
-    do {
-        parent = get_view_parent(parent);
-    } while (ggml_is_view(parent));
-    return parent;
-}
-
 static bool ggml_op_can_inplace(enum ggml_op op) {
     switch (op) {
         case GGML_OP_SCALE:
@@ -365,7 +425,6 @@ static bool ggml_op_can_inplace(enum ggml_op op) {
         case GGML_OP_DIAG_MASK_INF:
         case GGML_OP_ADD:
         case GGML_OP_ADD1:
-        case GGML_OP_ACC:
         case GGML_OP_SUB:
         case GGML_OP_MUL:
         case GGML_OP_DIV:
@@ -375,10 +434,8 @@ static bool ggml_op_can_inplace(enum ggml_op op) {
         case GGML_OP_UNARY:
         case GGML_OP_ROPE:
         case GGML_OP_RMS_NORM:
-        case GGML_OP_SET:
         case GGML_OP_SOFT_MAX:
         case GGML_OP_CONT:
-        case GGML_OP_ADD_REL_POS:
             return true;
 
         default:
@@ -390,24 +447,8 @@ static void allocate_node(struct ggml_allocr * alloc, struct ggml_tensor * node)
     struct hash_node * ht = alloc->hash_table;
     if (node->data == NULL) {
         if (ggml_is_view(node)) {
-            size_t offset;
-            switch(node->op) {
-                case GGML_OP_VIEW:
-                    memcpy(&offset, node->op_params, sizeof(size_t));
-                    node->data = (char *) node->src[0]->data + offset;
-                    break;
-                case GGML_OP_PERMUTE:
-                case GGML_OP_RESHAPE:
-                case GGML_OP_TRANSPOSE:
-                    node->data = node->src[0]->data;
-                    break;
-                case GGML_OP_CPY:
-                    node->data = node->src[1]->data;
-                    break;
-                default:
-                    GGML_ASSERT(!"unknown view op");
-                    break;
-            }
+            assert(node->view_src->data != NULL);
+            node->data = (char *)node->view_src->data + node->view_offs;
         } else {
             // see if we can reuse a parent's buffer (inplace)
             if (ggml_op_can_inplace(node->op)) {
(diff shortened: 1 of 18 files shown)


whisper.cpp
```diff
@@ -1762,212 +1660,259 @@ static bool whisper_encode_internal(
...
-        cur->src[0] = nullptr;
-        cur->src[1] = nullptr;
+    ggml_cgraph * gf = ggml_new_graph(ctx0);
 
-        for (int il = 0; il < model.hparams.n_text_layer; ++il) {
-            auto& layer = model.layers_decoder[il];
+    ggml_allocr * alloc = wstate.alloc_encode_post;
 
-            wstate.use_buf(ctx0, 0);
+    struct ggml_tensor * cur = ggml_view_tensor(ctx0, wstate.embd_enc);
```
ggerganov, line 1771: @slaren

I ported `ggml-alloc` into `whisper.cpp` and everything seems to work great.

There was one tricky moment, which I'm not sure why occurs - here it is in case you get some idea:

In Whisper, there are three steps executed one after the other:

- `encoder`
- `encoder_post`
- `decoder`

The `encoder` step outputs it's data into a tensor that we keep a reference to into the `whisper_state->embd_enc`:

https://github.com/ggerganov/whisper.cpp/blob/fa672b46e69e9aabd5b6799e0002372c4f8586eb/whisper.cpp#L1727-L1731

In the next step `encoder_post`, we want to use this data.
Initially, I directly tried to use the tensor from the state - i.e. `embd_enc`:

```c++
        struct ggml_tensor * cur = wstate.embd_enc;
        cur->op = GGML_OP_NONE;
        cur->src[0] = nullptr;
        cur->src[1] = nullptr;

       ...
```

And this seemed to work with the CPU version. Notice that this is not super clean, because `embd_enc` was created in the `ggml_ctx` of the `encoder` and this context is already destroyed. However, the allocator is alive and I checked that the `embd_enc->data` points to valid memory with the correct data.

Next I updated the CoreML functionality to work in the new implementation. The main difference when CoreML is enabled is that the `encoder` graph becomes just a single leaf in which the CoreML writes data:

https://github.com/ggerganov/whisper.cpp/blob/fa672b46e69e9aabd5b6799e0002372c4f8586eb/whisper.cpp#L1708-L1715

And this leaf is used in the same way in `encoder_post`.
However, this fails somewhere during the evaluation of `encoder_post` because the output data of `encoder_post` is incorrect. I checked that `embd_enc` has correct data at the start.

I thought that maybe because `embd_enc` is a tensor from a destroyed `ggml_ctx`, there is some problem that occurs. So tried to make a view of `embd_enc` which now lives is in the `ggml_ctx` of `encoder_post` and points to the same data:

https://github.com/ggerganov/whisper.cpp/blob/fa672b46e69e9aabd5b6799e0002372c4f8586eb/whisper.cpp#L1770-L1772

This is much cleaner compared to the hack above and it works correctly.
But still, I'm not sure why the first version does not work.

slaren: I noticed that `whisper_build_graph_encoder` and `whisper_build_graph_encoder_post` seem to use the same buffer for the context, ie. `wstate.buf_compute`. So I think that what may be happening is that the tensor struct, which is allocated from this buffer, is overwritten in `whisper_build_graph_encoder_post`. When using `ggml_view_tensor`, a new tensor struct is allocated instead of trying to reuse the old one, so that fixes the issue. Could this be the cause?

whisper.cpp
```diff
@@ -2817,12 +2771,59 @@ struct whisper_state * whisper_init_state(whisper_context * ctx) {
...
-    state->buf_scratch[1].resize(MEM_REQ_SCRATCH1.at(ctx->model.type));
-    state->buf_scratch[2].resize(MEM_REQ_SCRATCH2.at(ctx->model.type));
-    state->buf_scratch[3].resize(MEM_REQ_SCRATCH3.at(ctx->model.type));
+    state->buf_compute.resize(ggml_tensor_overhead()*GGML_MAX_NODES + ggml_graph_overhead());
+
+    static const size_t tensor_alignment = 32;
+
+    state->alloc_encode      = ggml_allocr_new_measure(tensor_alignment);
+    state->alloc_encode_post = ggml_allocr_new_measure(tensor_alignment);
+    state->alloc_decode      = ggml_allocr_new_measure(tensor_alignment);
```
slaren, line 2781: There is a chance that this will not work in some systems with limited virtual memory, such as iOS, because each measure allocator reserves a large amount of virtual memory. It would be safer to allocate only one measure allocator at a time, I think that should be possible here.
It's definitely not ideal that ggml-alloc has this limitation, I expect to improve this and remove the use of virtual memory entirely with the common backends interface implementation.

whisper.cpp
ggerganov: Yup, creating a separate `buf_compute` for `encoder_post` fixes the problem with the initial hacky version.

ggml-alloc.c
```diff
@@ -277,17 +310,68 @@ struct ggml_allocr * ggml_allocr_new(void * data, size_t size, size_t alignment)
...
 
-// address and size of the buffer when measuring
-// it needs to be large enough to fit all the tensors, but it cannot overlap with other existing buffers
-static void * const MEASURE_BASE_ADDR = (void *) 0x1000;
-static const size_t MEASURE_MAX_SIZE  = 1ULL<<40; // 1 TB
+// OS specific functions to allocate and free uncommitted virtual memory
+static void * alloc_vmem(size_t size) {
+#if defined(_WIN32)
+    return VirtualAlloc(NULL, size, MEM_RESERVE, PAGE_NOACCESS);
+#elif defined(_POSIX_MAPPED_FILES)
```
slaren, line 317: If the emscripten build doesn't work, it can be excluded from using mmap here and in `free_vmem` by checking if `__EMSCRIPTEN__ ` is defined. I think this should do it:
```suggestion
#elif defined(_POSIX_MAPPED_FILES) && !defined(__EMSCRIPTEN__)
```

whisper.cpp
ggerganov: I think I reorganized the allocators as proposed, but it seems some OSes still fail - see `linux/arm64` and `linux/ppc64le` in the CI: https://github.com/ggerganov/whisper.cpp/actions/runs/6146272809/job/16675319319

slaren: I am a bit confused by this, it seems that the call to `mmap` is crashing the process instead of returning an error, because otherwise we should see the failed assert `GGML_ASSERT(!"failed to allocate virtual memory for measure buffer");`. I imagine that this is related to QEMU, I'll try to reproduce it locally.

slaren: I tried the exact same commands that the CI uses to run the arm64 version with docker and QEMU, and it works on my computer. So whatever is the issue, it only seems to happen on the github CI environment and I cannot reproduce it. Maybe it is hitting some memory usage limit.

<details>

```bash
$ sudo docker run --platform linux/arm64 --rm \
    -v /home/diego/code/whisper.cpp:/workspace \
    -w /workspace ubuntu:22.04 /bin/sh -c '
    apt update
    apt install -y build-essential cmake libsdl2-dev
    cmake . -DWHISPER_SUPPORT_SDL2=ON -DCMAKE_BUILD_TYPE=Release
    make
    ctest -L gh --output-on-failure'

[...]

-- The C compiler identification is GNU 11.4.0
-- The CXX compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Could NOT find Git (missing: GIT_EXECUTABLE)
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- CMAKE_SYSTEM_PROCESSOR: aarch64
-- ARM detected
-- Configuring done
-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    WHISPER_SUPPORT_SDL2


-- Build files have been written to: /workspace
[  7%] Building C object CMakeFiles/whisper.dir/ggml.c.o
[ 15%] Building C object CMakeFiles/whisper.dir/ggml-alloc.c.o
[ 23%] Building CXX object CMakeFiles/whisper.dir/whisper.cpp.o
[ 30%] Linking CXX shared library libwhisper.so
[ 30%] Built target whisper
[ 38%] Building CXX object examples/CMakeFiles/common.dir/common.cpp.o
[ 46%] Building CXX object examples/CMakeFiles/common.dir/common-ggml.cpp.o
[ 53%] Linking CXX static library libcommon.a
[ 53%] Built target common
[ 61%] Building CXX object examples/main/CMakeFiles/main.dir/main.cpp.o
[ 69%] Linking CXX executable ../../bin/main
[ 69%] Built target main
[ 76%] Building CXX object examples/bench/CMakeFiles/bench.dir/bench.cpp.o
[ 84%] Linking CXX executable ../../bin/bench
[ 84%] Built target bench
[ 92%] Building CXX object examples/quantize/CMakeFiles/quantize.dir/quantize.cpp.o
[100%] Linking CXX executable ../../bin/quantize
[100%] Built target quantize
Test project /workspace
    Start 1: test-main-tiny
1/2 Test #1: test-main-tiny ...................   Passed   82.79 sec
    Start 2: test-main-tiny.en
2/2 Test #2: test-main-tiny.en ................   Passed   83.46 sec

100% tests passed, 0 tests failed out of 2

Label Time Summary:
en      =  83.46 sec*proc (1 test)
gh      = 166.24 sec*proc (2 tests)
tiny    = 166.24 sec*proc (2 tests)
```

</details>

slaren: A possible workaround could be reducing the amount of virtual memory allocated here:
https://github.com/ggerganov/whisper.cpp/blob/d3b2dd4955e40172c2dec4a39a3f0c2874a905be/ggml-alloc.c#L345-L346

ggerganov: Ok, thanks for looking into this. I'll now continue working on this branch and try to find a solution

ggml.c
```diff
@@ -4303,10 +4303,43 @@ int64_t ggml_nrows(const struct ggml_tensor * tensor) {
...
+            }
+        }
+    }
+
+    size_t nbytes = (sne[0]/ggml_blck_size(tensor->type))*snb[0];
     for (int i = 1; i < GGML_MAX_DIMS; ++i) {
-        nbytes += (tensor->ne[i] - 1)*tensor->nb[i];
+        nbytes += (sne[i] - 1)*snb[i];
     }
+
```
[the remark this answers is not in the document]
slaren: So, the goal of my implementation was to calculate the address of the last element plus one. However, the implementation assumes that `nb[0] == type_size`, so it doesn't work with transposed tensors. This should fix it for `blck_size == 1`:
```C
size_t nbytes = ggml_type_size(tensor->type);
for (int i = 0; i < GGML_MAX_DIMS; ++i) { 
    nbytes += (tensor->ne[i] - 1)*tensor->nb[i]; 
}
```

However, this will not work with quantized types. A possible solution could be fall back to the previous implementation for `blck_size > 1`, but it would be nicer to have a single implementation.
```C
size_t ggml_nbytes(const struct ggml_tensor * tensor) { 
    size_t nbytes;
    size_t blck_size = ggml_blck_size(tensor->type);
    if (blck_size == 1) { 
        nbytes = ggml_type_size(tensor->type);
        for (int i = 0; i < GGML_MAX_DIMS; ++i) { 
            nbytes += (tensor->ne[i] - 1)*tensor->nb[i]; 
        }
    }
    else {
        nbytes = tensor->ne[0]*tensor->nb[0]/blck_size;
        for (int i = 1; i < GGML_MAX_DIMS; ++i) { 
            nbytes += (tensor->ne[i] - 1)*tensor->nb[i]; 
        }
    }
    return nbytes;
}
```

whisper.cpp
ggerganov: Reducing the size to 128GB fixes the CI: https://github.com/ggerganov/whisper.cpp/pull/1270/commits/b19888cfb48d135466a890bccb348ad4e0772b14

ggml-alloc.c
ggerganov: Good news - the Emscripten build looks to be working without any adjustments needed

nchudleigh:
Benchmark comparing to current master. Looking promising @ggerganov 

image

ggerganov:
@nchudleigh I just pushed some extra speed-up changes and the Encoder is significantly faster. It's now faster than Core ML for all model sizes.

nchudleigh:
@ggerganov uhh yeah wow, you did good...

image

lawwu:
@ggerganov This is amazing. I used this yesterday and it worked wonderfully. Does this only work for .wav files?

ggerganov:
@lawwu The `whisper.cpp` library works with raw PCM audio. The examples in the repo work only with `.wav` format as it is easier to parse. Projects using `whisper.cpp` can load any audio format and pass the PCM audio to `whisper.cpp` for transcription

ggerganov:
I think this is ready to merge - one last call for tests on macOS and iOS to make sure everything works smoothly

ggerganov:
@colinc Thanks! I wonder why in the 3rd run Core ML got about ~2 times faster in the Encoder. On my M2 Ultra, I get comparable Encoder speed between `master` and this PR when using Core ML.

> MLComputeUnitsCPUAndGPU

Ah, I actually didn't realize Core ML can run on the GPU as well. I believe I am running it on the ANE. How do I make it run on the GPU - I forgot the steps for this.

colinc:
@ggerganov Honestly. I don't understand what changed that had that speedup. The commits earlier this morning on the metal-and-alloc branch didn't have that speedup. I can figure out which one did it (maybe 257d794).

For guiding which compute units are used by Core ML you can do the following in _coreml/whisper-encoder.mm_

```
    MLModelConfiguration *config = [[MLModelConfiguration alloc] init];
    config.computeUnits = MLComputeUnitsCPUAndGPU;
    //config.computeUnits = MLComputeUnitsCPUAndNeuralEngine;
    //config.computeUnits = MLComputeUnitsAll;

    const void * data = CFBridgingRetain([[whisper_encoder_impl alloc] initWithContentsOfURL:url_model configuration:config error:nil]);
```

However, I've found on the M1 (no modifier) chips, this is actually slower. M2 (no modifier) chips get a slight bump (but it's negligible). Higher performance GPUs obviously get a decent performance gain.

colinc:
Ok. I just ran the metal-and-alloc branch (23277d2) without COREML enabled against 71 files comprised of several hours of audio, both with single speakers (34 files) and dual speakers (37 files) with diarization. I have some scripts that reformat the results so it eliminates arbitrary line breaks (when each speaker is talking) and makes results more easily comparable.

A signifiant amount of the new transcriptions are identical and all of the changes / differences that I reviewed are positive improvements. Either more accurate transcriptions or punctuation. Or correcting hallucinations (including a few micro hallucinations I had missed when manually reviewing these files previously).

So, from my perspective, this is a net improvement for accuracy with (effectively) the same performance as MLComputeUnitsCPUAndGPU. If there's a way to further improve the encoding time with Metal that would just be a bonus. 😁

ggerganov:
Ok, so running the Core ML on the CPU + GPU is indeed faster.

Also there is some strange behavior where the first Core ML Encoder run after starting the process is ~x2 times slower, but next runs are very fast. This happens only with CPU + GPU Core ML, it does not happen with ANE Core ML. I've updated the `bench` tool to take this into account and provide more accurate measurement, by running a "pre-heat" encoder pass that is not measured.

I suppose that Core ML does some extra optimizations on the first run. I.e. the first Core ML GPU run is similar in performance to my Metal implementation, but then it gets a x2 factor lead. Would be interesting if whatever optimization occurs can be done manually in the Metal code. Would be a great benefit with potential dramatic improvement in the Decoder where we cannot use Core ML, but we can use Metal.

nchudleigh:
The improvement over ANE is insanely impressive.  I wonder if it will enable me to actually run the larger models with CoreML.

jhen0409:
I was very curious how this would perform on iOS, so I did some testing on it:

| Device (CPU) | Model | Load w/ Metal [ms] | Full w/ Metal [ms] | Load w/ Core ML (ANE) [ms] | Full w/ Core ML (ANE) [ms] |
|---|---|---|---|---|---|
| iPhone 13 Pro Max (A15) | tiny.en | 153 | 574 | 1160 | 239 |
| iPhone 13 Pro Max (A15) | base.en | 151 | 736 | 1626 | 369 |
| iPhone 13 Pro Max (A15) | small.en | 482 | 1871 | 6371 | 962 |
| iPhone 13 Pro Max (A15) | medium.en | 1295 | 4275 | 21136 | 2790 |
| iPhone 13 Pro Max (A15) | large | (crash) | (crash) | (crash) | (crash) |
| iPad Air 5 (M1) | tiny.en | 105 | 327 | 4357 | 149 |
| iPad Air 5 (M1) | base.en | 139 | 533 | (stuck) | (unknown) |
| iPad Air 5 (M1) | small.en | 247 | 1280 | (stuck) | (unknown) |
| iPad Air 5 (M1) | medium.en | 564 | 2919 | (stuck) | (unknown) |
| iPad Air 5 (M1) | large | 5118 | 5275 | (stuck) | (unknown) |

>  iOS 16.6.1, use jkv.wav, the Full is second run

In https://github.com/mybigday/whisper.rn/pull/123, I use commit f408c64564edf2ba6fe30ae94ed71db6289d430a, the difference is that I used ARC in ggml-metal.m and removed some release code. The app archive will be more faster than the build, but currently I use Xcode build to easily test.

Compared with iOS Device, the number of GPU cores is relatively small, I think that's why it has a gap on iPhone. I don't know why CoreML is not working on my M1 iPad, I didn't enable it in Production before so I didn't find out before.

What I love about the Metal backend is that it uses less disk size & memory, this means we can use a larger models in real-world scenarios. (Maybe we could consider to support CoreML OPs for GGML instead of load a mlmodelc in the future)

ggerganov:
@jhen0409 Thank you very much for the results!

I've further improved the Metal inference and now it is as fast as Core ML (GPU). I still have one kernel that is not optimized (the convolution at the start of the Encoder) so this explains the remaining difference between Metal and Core ML (GPU). I'm now satisfied with the results and indeed, Core ML does not do any extra optimizations - it is just slower the first run because it probably initializes some internal things.

The optimization that I did is to pad matrix multiplications that have a row dimension not multiple of 32. It's a very simple code change that can provide significant benefits for Metal. It should be straightforward to apply to `llama.cpp` as well and gain some extra performance:

https://github.com/ggerganov/whisper.cpp/blob/2b4160af29821c63b70689c0d571dbb1dc140cdd/whisper.cpp#L138-L171

nchudleigh:
| Commit  | Model              | Thread | Processor Count | Load Time | Sample Time | Encode Time | Decode Time | Sample Time per Run | Encode Time per Run | Decode Time per Run | Total Time |
| ------- | ------------------ | ------ | --------------- | --------- | ----------- | ----------- | ----------- | ------------------- | ------------------- | ------------------- | ---------- |
| master  | ggml-tiny.en.bin   | 4      | 1               | 54.23     | 81.29       | 263.68      | 274.8       | 0.4234              | 131.84              | 1.4313              | 708.43     |
| master  | ggml-base.en.bin   | 4      | 1               | 77.24     | 79.6        | 518.94      | 501.9       | 0.4124              | 259.47              | 2.6005              | 1223.97    |
| master  | ggml-small.en.bin  | 4      | 1               | 232.09    | 81.2        | 2248.65     | 1424.05     | 0.4207              | 1124.325            | 7.3785              | 4034.77    |
| master  | ggml-medium.bin    | 4      | 1               | 618.91    | 82.61       | 7265.93     | 3383.07     | 0.4193              | 3632.965            | 17.1729             | 11418.43   |
| master  | ggml-medium.en.bin | 4      | 1               | 585.99    | 78.57       | 7404.29     | 3389.91     | 0.4389              | 3702.145            | 18.9380             | 11525.09   |
| master  | ggml-large.bin     | 4      | 1               | 1108.13   | 79.07       | 12743.14    | 5901.58     | 0.4251              | 6371.57             | 31.7289             | 19960.5    |
| 8e8daa8 | ggml-tiny.en.bin   | 4      | 1               | 51.78     | 77.05       | 143.27      | 355.16      | 0.4013              | 71.635              | 1.8498              | 675.18     |
| 8e8daa8 | ggml-base.en.bin   | 4      | 1               | 79.24     | 77.64       | 275.19      | 532.86      | 0.4023              | 137.595             | 2.7609              | 1027.55    |
| 8e8daa8 | ggml-small.en.bin  | 4      | 1               | 219.28    | 77.51       | 782.83      | 1204.21     | 0.4016              | 391.415             | 6.2394              | 2344.08    |
| 8e8daa8 | ggml-medium.en.bin | 4      | 1               | 577.15    | 74.32       | 2117.78     | 2538.31     | 0.4152              | 1058.89             | 14.1805             | 5384.62    |
| 8e8daa8 | ggml-medium.bin    | 4      | 1               | 543.54    | 79.18       | 2120.5      | 2834.63     | 0.4019              | 1060.25             | 14.3890             | 5653.88    |
| 8e8daa8 | ggml-large.bin     | 4      | 1               | 1755.19   | 76.37       | 3711.21     | 4318.31     | 0.4106              | 1855.605            | 23.2167             | 9984.53    |
| 2b4160a | ggml-tiny.en.bin   | 4      | 1               | 50.08     | 78.1        | 94.62       | 320.86      | 0.4068              | 47.31               | 1.6799              | 593.96     |
| 2b4160a | ggml-base.en.bin   | 4      | 1               | 85.11     | 79.58       | 172.43      | 486.06      | 0.4123              | 86.215              | 2.5316              | 1702.74    |
| 2b4160a | ggml-small.en.bin  | 4      | 1               | 212.96    | 77.66       | 496.48      | 1118.51     | 0.4024              | 248.24              | 5.8256              | 1995.0     |
| 2b4160a | ggml-medium.en.bin | 4      | 1               | 555.72    | 75.79       | 1338.75     | 2378.07     | 0.4234              | 669.375             | 13.3599             | 4503.9     |
| 2b4160a | ggml-medium.bin    | 4      | 1               | 593.97    | 80.49       | 1334.06     | 2598.74     | 0.4086              | 667.03              | 13.3269             | 4792.06    |
| 2b4160a | ggml-large.bin     | 4      | 1               | 1831.34   | 77.67       | 2477.62     | 4044.54     | 0.4176              | 1238.81             | 21.9812             | 8704.82    |

colinc:
@ggerganov This is **amazing**! Thank you so, so much!

On an M2-Ultra (76-core), I'm now seeing 25x realtime (so ~2 minutes to process a ~50 minute long audio, including diarization).
But, more impressively, on a basic M2 (10-core) I'm seeing ~9.4x realtime (so ~5 minutes to process a ~50 minute long audio, including diarization).

And, most importantly, really good transcription results so far (especially vs previous results).

pudepiedj:
These are the timings I got from what I think is the current **master** model versions for **./samples/gb0.wav** (127.4s) on an M2 MAX with 32GB RAM on MacOS 13.5.1 with 8 threads (all timings in [ms]); maybe I have not done something right because the encoding and decoding times seem quite slow compared with those @ggerganov posted for an M2 Ultra on **X(formerly Twitter)**:
|Model|Load|Encode|Decode|Total|
|-----:|----:|------:|------:|-----:|
|tiny.en|49|107|607|1022|
|base.en|65|193|948|1557|
|medium.en|427|1353|4980|7259|
|large|993|2372|7695|11735|

Issue #1179, opened 2023-08-14
GPU utilization rate is very low with `WHISPER_CUBLAS=1`


bobqianic:
It seems that the CPU is working most of the time while the GPU is resting. There's still a lot of room for optimization. 
**A 27.8-minute audio takes 62.7 minutes to transcribe...**

model: `ggml-model-largev2.bin`
parameters: `-bs 5` `-bo 5`
audio: `diffusion2023-07-03.wav` `27.8 mins`

> 

<img width="627" alt="image" src="https://github.com/ggerganov/whisper.cpp/assets/129547291/764013c4-8166-45ed-a7ff-fdaf90d6acd4">

> 

<img width="647" alt="image" src="https://github.com/ggerganov/whisper.cpp/assets/129547291/32980c83-0c46-4640-97c3-4b66c667f095">

> 

<img width="649" alt="image" src="https://github.com/ggerganov/whisper.cpp/assets/129547291/a5a5eca5-7c58-466b-a4da-1249d66a7809">

MichaelDays:
Could you paste the full text of the log output, startup to shutdown?

jbrough:
It's at least an order of magnitude slower than my M2 Air. I'm trying with cuda on g4dn.xlarge

```
ggml_init_cublas: found 1 CUDA devices:
  Device 0: Tesla T4
  ```

ggerganov:
Yes, current cuBLAS support is quite rudimentary as we constantly keep moving data between the CPU and the GPU.
Better implementation with tensor offloading as in `llama.cpp` is possible, but don't have enough time to implement it yet

dereklll:
> Yes, current cuBLAS support is quite rudimentary as we constantly keep moving data between the CPU and the GPU. Better implementation with tensor offloading as in `llama.cpp` is possible, but don't have enough time to implement it yet

This should leave a lot of room for optimization of the whole process and hopefully help speed it up

dereklll:
whisper_print_timings:     load time =  1177.86 ms
whisper_print_timings:     fallbacks =   3 p /   1 h
whisper_print_timings:      mel time =    75.22 ms
whisper_print_timings:   sample time =   127.27 ms /   142 runs (    0.90 ms per run)
whisper_print_timings:   **encode time =  4798.87 ms /     7 runs (  685.55 ms per run)**
whisper_print_timings:   decode time =   755.72 ms /   125 runs (    6.05 ms per run)
[...]

PR #1293, opened 2023-09-15
whisper : add context param for disable gpu

jhen0409:
Currently the Metal backend is using some SIMD operations, which it's only supported for Apple7+ family devices (ref: [Metal-Feature-Set-Tables.pdf](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf)). In order to allow most older devices to run whisper.cpp normally, we can provide a param like `use_gpu`.

I think it will be helpful for some whisper.cpp binding that enabled Metal in build, but need to support old devices, or just don't want to access the GPU resources in some cases.

In this PR, I've added a new `struct whisper_context_params { use_gpu = true }` for that, I think it will also used for param like `use_mmap` in the future.

@ggerganov please let me know if you think this is a good idea or not, if yes I will do update for other backends & all bindings.


bobqianic:
It seems that this modification could introduce some backward compatibility issues, which would necessitate refactoring many of the APIs. I recall the last time I attempted to introduce a `debug-mode` option in whisper.cpp. It proved challenging to implement without disrupting the existing API. Generally speaking, I believe **_the current API design lacks elegance_**. A well-designed API should ensure both forward and backward compatibility, so that minor changes don't inconvenience or disrupt the users.

jhen0409:
> It seems that this modification could introduce some backward compatibility issues, which would necessitate refactoring many of the APIs. I recall the last time I attempted to introduce a `debug-mode` option in whisper.cpp. It proved challenging to implement without disrupting the existing API. Generally speaking, I believe **_the current API design lacks elegance_**. A well-designed API should ensure both forward and backward compatibility, so that minor changes don't inconvenience or disrupt the users.

We can consider to add new methods and deprecate the old methods, it can retain some flexibility for users.

jhen0409:
> It seems that this modification could introduce some backward compatibility issues, which would necessitate refactoring many of the APIs. I recall the last time I attempted to introduce a `debug-mode` option in whisper.cpp. It proved challenging to implement without disrupting the existing API. Generally speaking, I believe **_the current API design lacks elegance_**. A well-designed API should ensure both forward and backward compatibility, so that minor changes don't inconvenience or disrupt the users.

We can consider to add new methods and mark deprecated for the old methods, it can retain some flexibility for users.

(There were some issues on Comments with my phone that made me delete the previous comment, sorry for the double notifications.)

bobqianic:
@ggerganov ping

ggerganov:
Will take a look this week - been travelling for a few days

ggerganov:
@jhen0409 Yes, this is a good idea

jhen0409:
I'm looking for a way to disable CUDA backend (and OpenCL) by the param, should we just check `GGML_BACKEND_GPU` or need add a param to ggml?

I found all tensors are not setup backend as `GGML_BACKEND_GPU`, and it only use `ggml_cuda_mul_mat` (during can_mul_mat(...) will be true). Compared to llama.cpp, I think it missing setup such like a call of `ggml_cuda_assign_buffers`, but I'm not still not familiar with it. Also, it seems like related to #1179.

ggml-metal.m
```diff
@@ -150,56 +150,46 @@ @implementation GGMLMetalClass
...
+        } else {
+            metal_printf("%s: default.metallib not found, loading from source\n", __func__);
+
+            NSString * sourcePath = [bundle pathForResource:@"ggml-metal" ofType:@"metal"];
+            metal_printf("%s: loading '%s'\n", __func__, [sourcePath UTF8String]);
+            NSString * src = [NSString stringWithContentsOfFile:sourcePath encoding:NSUTF8StringEncoding error:&error];
+            if (error) {
+                metal_printf("%s: error: %s\n", __func__, [[error description] UTF8String]);
+                return NULL;
+            }
```
jhen0409, line 178: I enabled metal for test whisper.swiftui, and the project required to load compiled `default.metallib`, so I made this change. @bachittle I think this should be also help for https://github.com/ggerganov/llama.cpp/issues/3284.

For `GGML_SWIFT` I use `SWIFTPM_MODULE_BUNDLE` instead and reuse the code of default.metallib load, I think it should be also work in llama.cpp swift package.

jhen0409:
Aside from the GPU backend questions, I think another things are ready to review.

paulocoutinhox:
It solve this https://github.com/ggerganov/whisper.cpp/issues/1386?

bobqianic:
> I found all tensors are not setup backend as GGML_BACKEND_GPU, and it only use ggml_cuda_mul_mat (during can_mul_mat(...) will be true). Compared to llama.cpp, I think it missing setup such like a call of ggml_cuda_assign_buffers, but I'm not still not familiar with it. Also, it seems like related to https://github.com/ggerganov/whisper.cpp/issues/1179.

I'm in a similar situation. I'm completely lost in the llama.cpp code. Can't figure out how to offload the other operations to the GPU.

ggerganov:
> Aside from the GPU backend questions, I think another things are ready to review.

I will implement full GPU offloading in the following days.
Turning off CUDA is currently discussed in https://github.com/ggerganov/llama.cpp/pull/3946

ggerganov (approved):
@jhen0409 This should be good to merge, correct?

jhen0409:
> @jhen0409 This should be good to merge, correct?

Yes, it should be ready.

Issue #1292, opened 2023-09-15
ggml_metal_init failure: loading kernel function on Intel based mac 


nchudleigh:
Using latest on master ([951a119](https://github.com/ggerganov/whisper.cpp/commit/951a1199265ff4424f938182542cf6bac9b36154)) with metal 

load pipeline error: Error Domain=CompilerError Code=2 "SC compilation failure

<img width="305" alt="image" src="https://github.com/ggerganov/whisper.cpp/assets/5580824/48985067-6798-429f-8a90-496a35e6d326">

```
❯ ./main -m models/ggml-tiny.en.bin -f samples/jfk.wav
whisper_init_from_file_no_state: loading model from 'models/ggml-tiny.en.bin'
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51864
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 384
whisper_model_load: n_audio_head  = 6
whisper_model_load: n_audio_layer = 4
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 384
whisper_model_load: n_text_head   = 6
whisper_model_load: n_text_layer  = 4
whisper_model_load: n_mels        = 80
whisper_model_load: ftype         = 1
whisper_model_load: qntvr         = 0
whisper_model_load: type          = 1
whisper_model_load: adding 1607 extra tokens
whisper_model_load: model ctx     =   73.62 MB
whisper_model_load: model size    =   73.54 MB
whisper_init_state: kv self size  =    2.62 MB
whisper_init_state: kv cross size =    8.79 MB
whisper_init_state: compute buffer (conv)   =   11.17 MB
whisper_init_state: compute buffer (encode) =   61.76 MB
whisper_init_state: compute buffer (cross)  =    3.67 MB
whisper_init_state: compute buffer (decode) =   18.82 MB
ggml_metal_init: allocating
ggml_metal_init: found device: Intel Iris Pro Graphics
ggml_metal_init: found device: AMD Radeon R9 M370X
ggml_metal_init: picking default device: AMD Radeon R9 M370X
ggml_metal_init: loading '/Users/neil/Development/whisper.cpp/ggml-metal.metal'
ggml_metal_init: loaded kernel_add                         0x7fa14b207ab0 | th_max = 1024 | th_width =   64
ggml_metal_init: loaded kernel_add_row                     0x7fa14b208200 | th_max = 1024 | th_width =   64
[...]

PR #1294, opened 2023-09-15
Skip mm_mul kernel functions additions if on Intel

nchudleigh:
Tempfix for issue detailed here #1292

### DIFF
--- a/ggml-metal.m
+++ b/ggml-metal.m
@@ -21,6 +21,14 @@
 #define UNUSED(x) (void)(x)
 
 #define GGML_MAX_CONCUR (2*GGML_MAX_NODES)
+#include <sys/utsname.h>
+
+NSString *GetMachineHardwareName(void) {
+    struct utsname sysinfo;
+    int retVal = uname(&sysinfo);
+    if (EXIT_SUCCESS != retVal) return nil;
+    return [NSString stringWithUTF8String:sysinfo.machine];
+}
 
 struct ggml_metal_buffer {
     const char * name;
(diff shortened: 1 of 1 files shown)


jhen0409:
It's also related to https://github.com/ggerganov/llama.cpp/issues/2965, there may be more related issues in llama.cpp.

bobqianic (approved):

bobqianic:
It looks good, but this is making encoding really slow.

nchudleigh:
Otherwise, doesnt seem to work with intel at all. I wonder if its possible to write patches for the kernels.

nchudleigh:
@ggerganov @bobqianic I dont feel confident in this merge, feels like the fix should be more elegant / actually fix the issue. What do you think?

bobqianic:
> I dont feel confident in this merge, feels like the fix should be more elegant / actually fix the issue.

Agree.

ggerganov:
I'm not sure what is the best approach here.

@nchudleigh Does using Metal provide significant speed-up compared to running on the CPU for Mac Intel?
I am considering disabling Metal all together for Mac Intel if it does not offer a significant benefit

nchudleigh:
@ggerganov I missed this entirely!  

Hard to say whether it does or not... because the functions are missing I am not sure if the speed is improved or not. Probably depends on the GPU that is available.

bobqianic:
I've reopened this PR after noticing that @ggerganov addressed a similar issue in [#3524](https://github.com/ggerganov/llama.cpp/pull/3524). Instead of introducing a new kernel, he chose to disable the buggy kernels on older devices as a fix. We have two choices moving forward:

1. Wait for the next ggml sync.
2. Merge this PR into master directly.

```metal
        if ([ctx->device supportsFamily:MTLGPUFamilyApple7]) {
            GGML_METAL_ADD_KERNEL(mul_mm_f32_f32);
            GGML_METAL_ADD_KERNEL(mul_mm_f16_f32);
            GGML_METAL_ADD_KERNEL(mul_mm_q4_0_f32);
            GGML_METAL_ADD_KERNEL(mul_mm_q8_0_f32);
            GGML_METAL_ADD_KERNEL(mul_mm_q4_1_f32);
            GGML_METAL_ADD_KERNEL(mul_mm_q2_K_f32);
            GGML_METAL_ADD_KERNEL(mul_mm_q3_K_f32);
            GGML_METAL_ADD_KERNEL(mul_mm_q4_K_f32);
            GGML_METAL_ADD_KERNEL(mul_mm_q5_K_f32);
            GGML_METAL_ADD_KERNEL(mul_mm_q6_K_f32);
        }
```

z11h:
@ggerganov any idea when the next GGML sync will be? thank you :)

ggerganov:
Yup, let's try to fix this as part of #1422

PR #1310, opened 2023-09-19
save the recorded audio to a file

litongjava:
save the recorded audio to a file

### DIFF
--- a/examples/stream/stream.cpp
+++ b/examples/stream/stream.cpp
@@ -133,7 +133,7 @@ bool whisper_params_parse(int argc, char ** argv, whisper_params & params) {
         else if (arg == "-m"   || arg == "--model")         { params.model         = argv[++i]; }
         else if (arg == "-f"   || arg == "--file")          { params.fname_out     = argv[++i]; }
         else if (arg == "-tdrz" || arg == "--tinydiarize")  { params.tinydiarize   = true; }
-        else if (arg == "-sc"  || arg == "--save-audio")    { params.save_audio    = true; }
+        else if (arg == "-sa"  || arg == "--save-audio")    { params.save_audio    = true; }
 
         else {
             fprintf(stderr, "error: unknown argument: %s\n", arg.c_str());
--- a/examples/stream/stream.cpp
+++ b/examples/stream/stream.cpp
@@ -174,9 +174,9 @@ void whisper_print_usage(int /*argc*/, char **argv, const whisper_params &params
     fprintf(stderr, "  -l LANG,  --language LANG [%-7s] spoken language\n", params.language.c_str());
     fprintf(stderr, "  -m FNAME, --model FNAME   [%-7s] model path\n", params.model.c_str());
     fprintf(stderr, "  -f FNAME, --file FNAME    [%-7s] text output file name\n", params.fname_out.c_str());
-    fprintf(stderr, "  -tdrz,     --tinydiarize  [%-7s] enable tinydiarize (requires a tdrz model)\n",
+    fprintf(stderr, "  -tdrz,    --tinydiarize   [%-7s] enable tinydiarize (requires a tdrz model)\n",
             params.tinydiarize ? "true" : "false");
-    fprintf(stderr, "  -sc, --save-audio,        [%-7s] save the recorded audio to a file\n",
+    fprintf(stderr, "  -sc,      --save-audio    [%-7s] save the recorded audio to a file\n",
             params.tinydiarize ? "true" : "false");
 
     fprintf(stderr, "\n");
--- a/examples/stream/stream.cpp
+++ b/examples/stream/stream.cpp
@@ -285,6 +285,7 @@ int main(int argc, char ** argv) {
         if (params.save_audio && wavWriter) {
             wavWriter->writeData(pcmf32_new.data(), pcmf32_new.size());
         }
+        // handle Ctrl + C
         is_running = sdl_poll_events();
 
         if (!is_running) {
--- a/examples/stream/stream.cpp
+++ b/examples/stream/stream.cpp
@@ -1,7 +1,7 @@
-// Real-time speech recognition of input from a microphone
 //
 // A very quick-n-dirty implementation serving mainly as a proof of concept.
 //
+#include <fstream>
 
 #include "common-sdl.h"
 #include "common.h"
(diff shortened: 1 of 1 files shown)


litongjava:
[Save the correct audio](https://github.com/litongjava/whisper.cpp/commit/463450ed04ffe569b3e7e33277ce82b057ab1437)

litongjava:
How can I rerun CI / freeBSD-latest

examples/stream/stream.cpp
```diff
@@ -82,6 +133,7 @@ bool whisper_params_parse(int argc, char ** argv, whisper_params & params) {
         else if (arg == "-m"   || arg == "--model")         { params.model         = argv[++i]; }
         else if (arg == "-f"   || arg == "--file")          { params.fname_out     = argv[++i]; }
         else if (arg == "-tdrz" || arg == "--tinydiarize")  { params.tinydiarize   = true; }
+        else if (arg == "-sc"  || arg == "--save-audio")    { params.save_audio    = true; }
```
bobqianic, line 136:
```suggestion
        else if (arg == "-sa"  || arg == "--save-audio")    { params.save_audio    = true; }
```

examples/stream/stream.cpp
```diff
@@ -211,16 +263,29 @@ int main(int argc, char ** argv) {
...
+    printf("[Start speaking]\n");
     fflush(stdout);
 
-          auto t_last  = std::chrono::high_resolution_clock::now();
+    auto t_last  = std::chrono::high_resolution_clock::now();
     const auto t_start = t_last;
 
     // main audio loop
     while (is_running) {
         // handle Ctrl + C
```
bobqianic, line 285:
```suggestion
```

examples/stream/stream.cpp
```diff
@@ -211,16 +263,29 @@ int main(int argc, char ** argv) {
...
+        if (params.save_audio && wavWriter) {
+            wavWriter->writeData(pcmf32_new.data(), pcmf32_new.size());
+        }
         is_running = sdl_poll_events();
```
bobqianic, line 289:
```suggestion
        // handle Ctrl + C
        is_running = sdl_poll_events();
```

bobqianic (changes requested):

examples/stream/stream.cpp
```diff
@@ -117,6 +169,7 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
     fprintf(stderr, "  -m FNAME, --model FNAME   [%-7s] model path\n",                                     params.model.c_str());
     fprintf(stderr, "  -f FNAME, --file FNAME    [%-7s] text output file name\n",                          params.fname_out.c_str());
     fprintf(stderr, "  -tdrz,     --tinydiarize  [%-7s] enable tinydiarize (requires a tdrz model)\n",     params.tinydiarize ? "true" : "false");
+    fprintf(stderr, "  -sa,      --save-audio    [%-7s] save the recorded audio to a file\n",              params.tinydiarize ? "true" : "false");
```
bobqianic, line 172:
```suggestion
    fprintf(stderr, "  -sa,      --save-audio    [%-7s] save the recorded audio to a file\n",              params.save_audio ? "true" : "false");
```

bobqianic (changes requested):

litongjava:
Thanks,You've made some changes. You can merge them.

examples/stream/stream.cpp
litongjava: What to do with this place? Do you have any suggestions?

bobqianic: > What to do with this place? Do you have any suggestions?

I first removed this comment, and then moved it down to the right place.

bobqianic (approved):

litongjava:
Can you merge now?

examples/stream/stream.cpp
```diff
@@ -13,7 +13,57 @@
...
 
+    ~SimpleWavWriter() {
+        if (file.is_open()) {
+            file.seekp(4, std::ios::beg);
+            int32_t fileSize = 36 + dataSize;
+            file.write(reinterpret_cast<char *>(&fileSize), 4);
+            file.seekp(40, std::ios::beg);
+            file.write(reinterpret_cast<char *>(&dataSize), 4);
+            file.close();
+        }
```
bobqianic, line 64:
```suggestion
        if (file.is_open()) {
            file.close();
        }
```

examples/stream/stream.cpp
```diff
@@ -13,7 +13,57 @@
...
+        file.write("data", 4);
+        file.write("\0\0\0\0", 4);    // Placeholder for data size
+    }
+
+    void writeData(const float *data, size_t length) {
+        for (size_t i = 0; i < length; ++i) {
+            int16_t intSample = static_cast<int16_t>(data[i] * 32767);
+            file.write(reinterpret_cast<char *>(&intSample), sizeof(int16_t));
+            dataSize += sizeof(int16_t);
+        }
```
bobqianic, line 53:
```suggestion
    void writeData(const float *data, size_t length) {
        for (size_t i = 0; i < length; ++i) {
            int16_t intSample = static_cast<int16_t>(data[i] * 32767);
            file.write(reinterpret_cast<char *>(&intSample), sizeof(int16_t));
            dataSize += sizeof(int16_t);
        }
        if (file.is_open()) {
            file.seekp(4, std::ios::beg);
            int32_t fileSize = 36 + dataSize;
            file.write(reinterpret_cast<char *>(&fileSize), 4);
            file.seekp(40, std::ios::beg);
            file.write(reinterpret_cast<char *>(&dataSize), 4);
            file.seekp(0, std::ios::end);
        }
    }
```

examples/stream/stream.cpp
```diff
@@ -407,5 +472,8 @@ int main(int argc, char ** argv) {
     whisper_print_timings(ctx);
     whisper_free(ctx);
 
+    if (wavWriter) {
+        delete wavWriter;
+    }
```
bobqianic, line 477:
```suggestion
```

bobqianic (changes requested):

examples/stream/stream.cpp
bobqianic: There's a bug here. Although I understand that you intend to write the audio data size information when this class is destructed, this can pose problems. For instance, in the event of a sudden system crash, or if I terminate the program using `Ctrl+C`, the program will be killed instantly, leaving no time to write the audio data size information. Consequently, the audio would be corrupted and unplayable.

examples/stream/stream.cpp
bobqianic: The destructor for an object is automatically called when the object goes out of scope.

examples/stream/stream.cpp
```diff
@@ -13,7 +13,61 @@
...
+        if (file.is_open()) {
+            file.seekp(4, std::ios::beg);
+            int32_t fileSize = 36 + dataSize;
+            file.write(reinterpret_cast<char *>(&fileSize), 4);
+            file.seekp(40, std::ios::beg);
+            file.write(reinterpret_cast<char *>(&dataSize), 4);
+            file.seekp(0, std::ios::end);
+        }
+    }
+    }
```
bobqianic, line 63:
```suggestion
```

bobqianic (approved):

examples/stream/stream.cpp
```diff
@@ -2,7 +2,7 @@
 //
 // A very quick-n-dirty implementation serving mainly as a proof of concept.
 //
-
+#include <fstream>
```
ggerganov, line 5: Already included below

examples/stream/stream.cpp
```diff
@@ -13,7 +13,60 @@
 #include <thread>
 #include <vector>
 #include <fstream>
+#include <ctime>
+
+class SimpleWavWriter {
```
ggerganov, line 18: - The naming convention is to use snake case, so this should be `class wav_writer`
- Move to `common.h / common.cpp`
- API should be: `open()`, `write()`, `close()`
- Allocate on the stack instead of on the heap with `new` and forgetting to deallocate it
- If `params.save_audio` is true, call `init()`

bobqianic: OK

PR #1335, opened 2023-09-30
Better abort callback

mkiol:
Current abort mechanism with `whisper_full_params.encoder_begin_callback` works but it is not super effective. Usually "encoding" takes the longest, so if abort is requested during encoding, abort execution is very delayed.

This PR exposes in whisper API, already existing ggml abort callback. This callback ends any ggml computation, therefore whisper aborting is much quicker.


ggerganov (approved):

mkiol:
Thank you for merging and for this amazing project.

PR #1349, opened 2023-10-07
Prefer pkg-config while looking for BLAS

marmistrz:
It turns out that BLAS is _still_ not working on Arch Linux and `FindBLAS.cmake` fails to find `libopenblas_64.so` provided by `openblas64` (note that Arch Linux does _not_ have separate `-dev` packages).

This fixes it on Arch Linux. Even though the library `libopenblas.so` (32-bit) is detected, the performance difference is negligible on my system and it's better than nothing at all.


marmistrz:
> > the performance difference is negligible on my system and it's better than nothing at all.
> 
> That's odd. I'd expect OpenBLAS to perform much better than ggml on a CPU. I'll review this PR soon. Can you provide more details about your hardware and system?

I mean the difference between openblas and openblas64, not openblas and ggml

bobqianic:
> I mean the difference between openblas and openblas64, not openblas and ggml

Ah, I see.

bobqianic (approved):

marmistrz:
Btw., the reason why we can't find openblas64 with CMake are these lines:
https://github.com/Kitware/CMake/blob/bc82d62b1bb484aabb1fdbb563c44f6e7cffb55f/Modules/FindBLAS.cmake#L724-L726

If the file is changed to 
```
  if(_blas_sizeof_integer EQUAL 8)
    string(APPEND _blas_flexiblas_lib "_64")
  endif()
```
then `/usr/lib/libopenblas_64.so` is correctly detected without having to use pkgconfig.

PR #1350, opened 2023-10-07
Abort build if a feature was requested and could not be configured

marmistrz:
This is important because bindings might request features and not realize that they were not enabled. Note that all of these features are disabled by default and have to be manually turned on.


bobqianic:
Agree. This is also beneficial for CI testing.

bobqianic (approved):

PR #1352, opened 2023-10-08
Add library versioning

kenneth-ge:
This pull request adds library versioning to whisper.cpp. 

This is important so applications and distributors can determine when the ABI has changed and see whether the application needs to be recompiled. That way, we can safely compile and use a shared library in other applications and make some guarantees about the shared library's stability. 

If library versioning were to be added and maintained (specifically the SOVERSION), this would allow work to start on integrating whisper.cpp into e.g. Gnome Desktop, where it is desperately needed to increase accessibility


ggerganov:
Thanks, just to make sure I understand: I have to bump the SOVERSION every time there is a backward incompatible change in `whisper.h`, is this correct?

petterreinholdtsen:
Yes, that is correct.  With VERSION 1.0.0 and SOVERSION 1, the library name will look like this:

```
lrwxrwxrwx 1 pere pere     15 feb.  22 09:24 libwhisper.so -> libwhisper.so.1
lrwxrwxrwx 1 pere pere     19 feb.  22 09:24 libwhisper.so.1 -> libwhisper.so.1.0.0
-rwxr-xr-x 1 pere pere 985104 feb.  22 09:24 libwhisper.so.1.0.0
```

The binaries linking with -lwhisper will list "libwhisper.1" in their binary, and when the library become backwards incompatible and switch to major number 2, newly created binaries will list "libwhisper.2" and both sets of binaries will keep working.

petterreinholdtsen:
https://cmake.org/cmake/help/latest/prop_tgt/SOVERSION.html and https://www.netfort.gr.jp/~dancer/column/libpkg-guide/libpkg-guide.html have more information than you need about handling shared libraries.

kenneth-ge:
> Thanks, just to make sure I understand: I have to bump the SOVERSION every time there is a backward incompatible change in `whisper.h`, is this correct?

Yep! That is correct

ggerganov (approved):

Issue #1354, opened 2023-10-08
Converting whisper model to openvino fails


sp00ktober:
Morning, im trying to follow the steps from the Readme to convert the whisper model to openvino (to try it on my rPI4). Im doing this in a docker container on my Windows PC, installed the `openvino-conversion-requirements.txt`

Running `python convert-whisper-to-openvino.py --model large` with the `large` model produces the following error after a while:

```
100%|█████████████████████████████████████| 2.87G/2.87G [19:57<00:00, 2.58MiB/s]
/models/whisper.cpp/models/openvino_conv_env/lib/python3.10/site-packages/whisper/model.py:166: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  assert x.shape[1:] == self.positional_embedding.shape, "incorrect audio shape"
Killed
```

Is the large model not supported? Am i doing something wrong?
This also happens on the `medium` model.

sp00ktober:
Maybe as addition, running the same commands on my rPI4 4GB (with insane 12GB swap, booted from a USB3 ssd, ubuntu server 2023) results in `Segmentation fault (core dumped)`.
Not sure why, i thought maybe because of OOM thats why i increased the swap file but im not sure if increasing it further would solve this.

bobqianic:
> convert the whisper model to openvino (to try it on my rPI4).

Wait, does the Raspberry Pi 4 support OpenVINO? Isn't it exclusive to the Intel x86 platform?

sp00ktober:
![grafik](https://github.com/ggerganov/whisper.cpp/assets/79335616/d2f25cd2-444e-49ff-bbb3-46704f30f9f3)
At least Pi 4 Model B is listed [here](https://github.com/openvinotoolkit/openvino) and i succeeded in compiling OpenVINO from sources on it, altho it took more or less a day to complete.

bobqianic:
```console
/models/whisper.cpp/models/openvino_conv_env/lib/python3.10/site-packages/whisper/model.py:166: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  assert x.shape[1:] == self.positional_embedding.shape, "incorrect audio shape"
```

[...]

PR #1356, opened 2023-10-10
Use absolute paths for the converted OpenVINO model

bobqianic:
This PR fixed the problem mentioned in #1354


ggerganov (approved):
Thank you @bobqianic for taking care of the repo.
I plan to invest some time in the following week to make a new release and add quite a few improvements to the project

PR #1358, opened 2023-10-12
JSON: allow outputting per-token data too

akx:
... for a better karaoke experience.

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -83,6 +83,7 @@ struct whisper_params {
     bool output_wts      = false;
     bool output_csv      = false;
     bool output_jsn      = false;
+    bool output_jsn_full = false;
     bool output_lrc      = false;
     bool print_special   = false;
     bool print_colors    = false;
@@ -151,6 +152,7 @@ bool whisper_params_parse(int argc, char ** argv, whisper_params & params) {
         else if (arg == "-fp"   || arg == "--font-path")       { params.font_path       = argv[++i]; }
         else if (arg == "-ocsv" || arg == "--output-csv")      { params.output_csv      = true; }
         else if (arg == "-oj"   || arg == "--output-json")     { params.output_jsn      = true; }
+        else if (arg == "-ojf"  || arg == "--output-json-full"){ params.output_jsn_full = params.output_jsn = true; }
         else if (arg == "-of"   || arg == "--output-file")     { params.fname_out.emplace_back(argv[++i]); }
         else if (arg == "-ps"   || arg == "--print-special")   { params.print_special   = true; }
         else if (arg == "-pc"   || arg == "--print-colors")    { params.print_colors    = true; }
@@ -206,6 +208,7 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
     fprintf(stderr, "  -fp,       --font-path         [%-7s] path to a monospace font for karaoke video\n",     params.font_path.c_str());
     fprintf(stderr, "  -ocsv,     --output-csv        [%-7s] output result in a CSV file\n",                    params.output_csv ? "true" : "false");
     fprintf(stderr, "  -oj,       --output-json       [%-7s] output result in a JSON file\n",                   params.output_jsn ? "true" : "false");
+    fprintf(stderr, "  -ojf,      --output-json-full  [%-7s] include more information in the JSON file\n",      params.output_jsn_full ? "true" : "false");
     fprintf(stderr, "  -of FNAME, --output-file FNAME [%-7s] output file path (without file extension)\n",      "");
     fprintf(stderr, "  -ps,       --print-special     [%-7s] print special tokens\n",                           params.print_special ? "true" : "false");
     fprintf(stderr, "  -pc,       --print-colors      [%-7s] print colors\n",                                   params.print_colors ? "true" : "false");
@@ -511,7 +514,12 @@ bool output_score(struct whisper_context * ctx, const char * fname, const whispe
     return true;
 }
 
-bool output_json(struct whisper_context * ctx, const char * fname, const whisper_params & params, std::vector<std::vector<float>> pcmf32s) {
+bool output_json(
+             struct whisper_context * ctx,
+                         const char * fname,
+               const whisper_params & params,
+    std::vector<std::vector<float>>   pcmf32s,
+                               bool   full) {
     std::ofstream fout(fname);
     int indent = 0;
 
@@ -528,7 +536,7 @@ bool output_json(struct whisper_context * ctx, const char * fname, const whisper
     auto end_arr = [&](bool end) {
         indent--;
         doindent();
-        fout << (end ? "]\n" : "},\n");
+        fout << (end ? "]\n" : "],\n");
     };
 
     auto start_obj = [&](const char *name) {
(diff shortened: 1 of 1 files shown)


examples/main/main.cpp
```diff
@@ -628,7 +634,23 @@ bool output_json(struct whisper_context * ctx, const char * fname, const whisper
...
+                        start_arr("tokens");
+                        const int n = whisper_full_n_tokens(ctx, i);
+                        for (int j = 0; j < n; ++j) {
+                            auto token = whisper_full_get_token_data(ctx, i, j);
+                            start_obj(nullptr);
+                                value_s("text", whisper_token_to_str(ctx, token.id), false);
+                                value_i("from", token.t0 * 10, false);
+                                value_i("to", token.t1 * 10, false);
+                                value_i("id", token.id, false);
+                                value_i("p", token.p, true);
```
bobqianic, line 649: `value_i` accepts only integers as input, while `token.p` is a `float`. Therefore, to write a `float` into JSON, you'll need to create a new function named `value_f`.

https://github.com/ggerganov/whisper.cpp/blob/6b4f5f4488f0575db69c2d1de4f4009d4f73f5fd/whisper.h#L74-L90

examples/main/main.cpp
```diff
@@ -628,7 +634,23 @@ bool output_json(struct whisper_context * ctx, const char * fname, const whisper
...
+                            end_obj(j == (n - 1));
+                        }
+                        end_arr(false);
```
bobqianic, line 652:
```suggestion
                        end_arr(true);
```

Otherwise, you'll end up with an extra comma `,` at the end.

examples/main/main.cpp
```diff
@@ -1012,7 +1034,7 @@ int main(int argc, char ** argv) {
             // output to JSON file
             if (params.output_jsn) {
                 const auto fname_jsn = fname_out + ".json";
-                output_json(ctx, fname_jsn.c_str(), params, pcmf32s);
+                output_json(ctx, fname_jsn.c_str(), params, pcmf32s, true);
```
bobqianic, line 1037: Could you offer an alternative option, allowing users to decide if they require token-level data?

akx:
@bobqianic Thanks, great catches. Added `-ojf` (that needs to be used with `-oj`) for full JSON output.

examples/main/main.cpp
```diff
@@ -1012,7 +1042,7 @@ int main(int argc, char ** argv) {
             // output to JSON file
             if (params.output_jsn) {
```
bobqianic, line 1043:
```suggestion
            if (params.output_jsn || params.output_jsn_full) {
```

bobqianic (changes requested):

examples/main/main.cpp
```diff
@@ -511,7 +513,13 @@ bool output_score(struct whisper_context * ctx, const char * fname, const whispe
...
 }
 
-bool output_json(struct whisper_context * ctx, const char * fname, const whisper_params & params, std::vector<std::vector<float>> pcmf32s) {
+bool output_json(
+    struct whisper_context * ctx,
+    const char * fname,
+    const whisper_params & params,
+    std::vector<std::vector<float>> pcmf32s,
+    bool token_timing
+) {
```
bobqianic, line 522:
```suggestion
bool output_json(
             struct whisper_context * ctx,
                         const char * fname,
               const whisper_params & params,
    std::vector<std::vector<float>>   pcmf32s,
                  bool token_timing
) {
```

bobqianic (commented):
Code style

piotr-sikora-v:
There is any option to add also to json-full information about how good whisper recognise words?

bobqianic:
> EDIT: there's still something off about the results...

How's it going?

polvoazul:
I did a similar thing but for realtime mode!

akx:
@bobqianic Okay, ready to go again. The "something off" was that the per-token `t0` and `t1` were -10 because per-token timestamps would only be computed when `wts` output was _also_ requested. Now they're elided if they're `-1`, but requesting full JSON output will request per-token timestamp computation too.

examples/main/main.cpp
akx: Made `ojf` imply `output_jsn` in argument parsing instead :)

bobqianic:
> @bobqianic Okay, ready to go again. The "something off" was that the per-token `t0` and `t1` were -10 because per-token timestamps would only be computed when `wts` output was _also_ requested. Now they're elided if they're `-1`, but requesting full JSON output will request per-token timestamp computation too.

OK, I see. I will test your code tonight.

examples/main/main.cpp
```diff
@@ -628,7 +643,27 @@ bool output_json(struct whisper_context * ctx, const char * fname, const whisper
...
+
+                    if (full) {
+                        start_arr("tokens");
+                        const int n = whisper_full_n_tokens(ctx, i);
+                        for (int j = 0; j < n; ++j) {
+                            auto token = whisper_full_get_token_data(ctx, i, j);
+                            start_obj(nullptr);
+                                value_s("text", whisper_token_to_str(ctx, token.id), false);
+                                if(token.t0 > -1) {
+                                    value_i("from", token.t0 * 10, false);
```
bobqianic, line 656:
```suggestion
                                    value_i("from", to_timestamp(token.t0, true).c_str(), false);
```

examples/main/main.cpp
```diff
@@ -628,7 +643,27 @@ bool output_json(struct whisper_context * ctx, const char * fname, const whisper
...
+                                }
+                                if(token.t1 > -1) {
+                                    value_i("to", token.t1 * 10, false);
```
bobqianic, line 659:
```suggestion
                                    value_i("to", to_timestamp(token.t1, true).c_str(), false);
```

bobqianic (changes requested):

bobqianic:
image

bobqianic (approved):

examples/main/main.cpp
```diff
@@ -511,7 +514,13 @@ bool output_score(struct whisper_context * ctx, const char * fname, const whispe
...
 }
 
-bool output_json(struct whisper_context * ctx, const char * fname, const whisper_params & params, std::vector<std::vector<float>> pcmf32s) {
+bool output_json(
+    struct whisper_context * ctx,
+    const char * fname,
+    const whisper_params & params,
+    std::vector<std::vector<float>> pcmf32s,
+    bool full
+) {
```
bobqianic, line 523:
```suggestion
bool output_json(
             struct whisper_context * ctx,
                         const char * fname,
               const whisper_params & params,
    std::vector<std::vector<float>>   pcmf32s,
                               bool   full) {
```

bobqianic (changes requested):
Code Style

akx:
@bobqianic using `to_timestamp` will make this programmatic format much harder to parse by, well, programs. I disagree with that choice...

bobqianic:
> using to_timestamp will make this programmatic format much harder to parse by, well, programs.

Most programming languages, like C++, Python, Java, and Go, have built-in libraries that make parsing strings into time formats quite straightforward. It's not significantly more challenging. I believe that since one of the main benefits of using JSON over binary is its **human-readable** nature, utilizing `to_timestamp` would enhance the readability of timestamps.

akx:
@bobqianic Okay, let's split the difference and do like the top-level data does and have both the numeric offsets and the human-readable timestamps in? :)

bobqianic (approved):

PR #1364, opened 2023-10-13
ISSUE-1329: replace " with ' so it doesn't try to execute code in backticks

spullara:
This should prevent escape.

### DIFF
--- a/examples/talk-llama/talk-llama.cpp
+++ b/examples/talk-llama/talk-llama.cpp
@@ -661,7 +661,7 @@ int main(int argc, char ** argv) {
                     }
                 }
 
-                text_to_speak = ::replace(text_to_speak, "'", "");
+                text_to_speak = ::replace(text_to_speak, "'", "'\"'\"'");
                 int ret = system((params.speak + " " + std::to_string(voice_id) + " '" + text_to_speak + "'").c_str());
                 if (ret != 0) {
                     fprintf(stderr, "%s: failed to speak\n", __func__);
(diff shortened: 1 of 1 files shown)


jxy:
I have been using this

```patch
@@ -661,8 +660,8 @@ int main(int argc, char ** argv) {
                     }
                 }
 
-                text_to_speak = ::replace(text_to_speak, "\"", "");
-                int ret = system((params.speak + " " + std::to_string(voice_id) + " \"" + text_to_speak + "\"").c_str());
+                text_to_speak = ::replace(text_to_speak, "'", "'\"'\"'");
+                int ret = system((params.speak + " " + std::to_string(voice_id) + " '" + text_to_speak + "'").c_str());
                 if (ret != 0) {
                     fprintf(stderr, "%s: failed to speak\n", __func__);
                 }
```

so single quotes and possessive cases would still be there.

ggerganov:
> I have been using this
> 
> ```diff
> @@ -661,8 +660,8 @@ int main(int argc, char ** argv) {
>                      }
>                  }
>  
> -                text_to_speak = ::replace(text_to_speak, "\"", "");
> -                int ret = system((params.speak + " " + std::to_string(voice_id) + " \"" + text_to_speak + "\"").c_str());
> +                text_to_speak = ::replace(text_to_speak, "'", "'\"'\"'");
> +                int ret = system((params.speak + " " + std::to_string(voice_id) + " '" + text_to_speak + "'").c_str());
>                  if (ret != 0) {
>                      fprintf(stderr, "%s: failed to speak\n", __func__);
>                  }
> ```
> 
> so single quotes and possessive cases would still be there.

Yeah, the proposed PR will break for single quotes like `you're` `what's`, etc. so it's replacing one problem with another.
Need a better solution

spullara:
Updated with the change suggested.

ggerganov (approved):

PR #1370, opened 2023-10-16
Add support for Swift Package Manager

sindresorhus:
Not ready. Depends on syncing with llama.cpp to get latest `ggml-metal` (also `ggml-backend` and `k_quants`).


ggerganov:
Bringing all `llama.cpp` stuff with #1422 - after that, will work on improving the build across platforms, new features and some performance improvements. Hopefully will be ready in the following week.

sindresorhus:
@ggerganov I have synced it from llama.cpp again. It now builds successfully in Xcode for me. I added it as a local dependency to my app.

---

I'm also seeing https://github.com/ggerganov/whisper.spm/blob/9752de41007fc54df5d6f98945929a54bc628a58/Package.swift#L23-L26, but I don't think that should stop it from being included here as it does build in Xcode, which is the main use-case. We could maybe document the caveat for now?

ggerganov:
Since it works on your end, I think we can assume I was doing something wrong, so we should proceed.
Just to make sure I understand: merging this will technically obsolete the [whisper.spm](https://github.com/ggerganov/whisper.spm) and people can simply use the `whisper.cpp` repo in their Swift projects, correct?

sindresorhus:
> Just to make sure I understand: merging this will technically obsolete the [whisper.spm](https://github.com/ggerganov/whisper.spm) and people can simply use the whisper.cpp repo in their Swift projects, correct?

Correct

sindresorhus:
I have updated the SwiftUI example app to use the package: https://github.com/ggerganov/whisper.cpp/pull/1370/commits/bacf34dd346ccd8a63cd73bb202a8df8b29df475

Issue #1369, opened 2023-10-16
Server example?


Azeirah:
I'm working on a voice-controlled application and I want to run small .wav files through whisper fairly often.

What I noticed is that it takes almost 50% of the total time just to load the model every single time I run `./main -m ... "my-short-spoken-command.wav"`

I think it'd be nice if like in llama.cpp this project includes a server example so the model only has to be loaded once and stays in memory after loading.

FSSRepo:
Maybe, when I finish working on optimizing stable-diffusion.cpp and adding a server to it, I could create a server example for whisper.cpp.

Azeirah:
I posted the code as a PR https://github.com/ggerganov/whisper.cpp/pull/1375

ggerganov:
Hey all, I notice several server examples being proposed. This is super cool!

I'm planning to do a major update to `whisper.cpp` in the following few days, bringing some new features and performance improvements. This will be the highest priority, so for the sake of less distractions, the server examples will have to wait until we finish the new release. Sorry for the delay

ggerganov:
Hi again! I think we should restart the server efforts now that `v1.5.0` is released.

I like both #1375 and #1380, so not sure how to decide which one to integrate. 
@Azeirah and @felrock (and others): do you have any opinion on this.

Also, I think we should aim to support the OpenAI Audio API for speech to text: https://platform.openai.com/docs/api-reference/audio

The approach in #1418 is also interesting, so it can be merged as an alternative solution to the REST-based server example.

felrock:
Hello! I'm keen on fixing and merging my changes for the server. I've seen that the use for server in llama.cpp has enabled projects such as ollama and others. So i think it's an important application to have, for users to easily create interfaces against.

I have also started to create a similar server solution for [bark.cpp](https://github.com/PABannier/bark.cpp) because in my use case I would like to have some sort of voice(a bit more granular than espeak). Which would complete the full llm robot, a brain(llama.cpp), ears(whisper.cpp) and voice(bark.cpp).

ggerganov:
Yup, I agree that a server can find many interesting applications.

> Which would complete the full llm robot, a brain(llama.cpp), ears(whisper.cpp) and voice(bark.cpp).

Yes! Great idea - we are getting close :)

colinator:
[...]

PR #1375, opened 2023-10-18
Draft: Basic example server

Azeirah:
A server example for whisper.cpp, based on the LLaMa.cpp server example.
This is still a work-in-progress, but the very basics work.

Why would you want to run whisper as a server? If you need to process a short clip often, for example on a voice-controlled application, the model you're using is loaded into RAM every time `./main` is invoked. For very short voice snippets of about 3-5 seconds, about 50% of the time is spent loading the base-en model every time.

With this you can simply send a http request to this web-server and your wav file gets converted immediately.

## Progress

- [x] Supports basic (model selection & performance & quality) parameters for whisper
- [x] Supports same server parameters as LLaMa.cpp server example
- [x] Supports `/convert` endpoint, send a .wav filename and get transcription + metadata back as JSON
- [ ] Sends error messages in JSON when something goes wrong
- [ ] Supports diarization
- [ ] Supports language detection
- [ ] Server is documented in README.md

Note that it's been a while since I used C++, I made most of this happen with frankenstein copy-pasting. If I'm doing something stupid please let me know.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -23,7 +23,7 @@ add_library(${TARGET} STATIC
     common.cpp
     common-ggml.h
     common-ggml.cpp
-    )
+)
 
 include(DefaultTargetOptions)
 
@@ -67,6 +67,7 @@ else()
     add_subdirectory(command)
     add_subdirectory(bench)
     add_subdirectory(quantize)
+    add_subdirectory(server)
     add_subdirectory(talk)
     add_subdirectory(talk-llama)
     add_subdirectory(lsp)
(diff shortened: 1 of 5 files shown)


litongjava:
When do you plan to merge this PR, I want to use this feature as soon as possible.

Azeirah:
> When do you plan to merge this PR, I want to use this feature as soon as possible.

Whenever ggerganov is done with his changes. I'll make sure a reasonably feature complete server example will be merged.

https://github.com/ggerganov/whisper.cpp/issues/1369#issuecomment-1792050527

PR #1381, opened 2023-10-20
Fix download coreml script with zipfile

yanyaoer:
Fix download url with zipfile and extra


bobqianic:
I think these two parts also need some changes.

image
image

pkreissel:
I tried your commit locally. It did not work.
First it provided a zip-file, maybe the script could unpack that automatically?
The corelmc required a file called "ggml-medium.mlmodel" so I renamed the unpacked file to that.
Then corelmc command then complained that there was no Manifest.json in the model. 
I tried renaming the metadata.json to manifest.json but it did not work. 

I hope this is helpful in improving the script, also I would be happy if someone could help me with this.

Issue #1336, opened 2023-09-30
'common-sdl.h' file not found — MacOS Sonoma


joshfarrant:
When trying to `make` any of the examples requiring `common-sdl.h` I run into

```
command.cpp:9:10: fatal error: 'common-sdl.h' file not found
```

I've installed `sdl2` using brew, and I've even tried building from source, but no joy.

Any suggestions appreciated, thanks

DissonanceTK:
Same issue - MacOS Sonoma. Macbook Air M1.
`make stream
c++     stream.cpp   -o stream
stream.cpp:5:10: fatal error: 'common-sdl.h' file not found
#include "common-sdl.h"
         ^~~~~~~~~~~~~~
1 error generated.
make: *** [stream] Error 1`

ai-at-home:
It seems likely that this error comes from not compiling in the root directory.  I am on Linux however if you attempt to compile from within the directory this is the error you get:

```
whisper.cpp/examples/stream$ make stream
g++     stream.cpp   -o stream
stream.cpp:6:10: fatal error: common/sdl.h: No such file or directory
    6 | #include "common/sdl.h"
      |          ^~~~~~~~~~~~~~
compilation terminated.
make: *** [<builtin>: stream] Error 1
```

The solution is simply to `cd ../../` to go to the root dir then compile.

As such I don't think its a bug so much as perhaps a clarification needed in the docs.

joyyang1215:
"The solution is simply to cd ../../ to go to the root dir then compile." --- it works

PR #1400, opened 2023-10-29
Clarify doc about where to compile from

ai-at-home:
I was confused for a bit as to where some libs were coming from.  I simply had the wrong dir I was attempting to make from.  It [appears others have the same issue](https://github.com/ggerganov/whisper.cpp/issues/1336) so proposing to update the doc to clarify what dir you need to be in and why.

### DIFF
--- a/examples/stream/README.md
+++ b/examples/stream/README.md
@@ -40,10 +40,18 @@ make stream
 ```
 
 Ensure you are at the root of the repo when running `make stream`.  Not within the `examples/stream` dir
-
 as the libraries needed like `common-sdl.h` are located within `examples`.  Attempting to compile within
-
 `examples/steam` means your compiler cannot find them and it gives an error it cannot find the file.
+
+```bash
+whisper.cpp/examples/stream$ make stream
+g++     stream.cpp   -o stream
+stream.cpp:6:10: fatal error: common/sdl.h: No such file or directory
+    6 | #include "common/sdl.h"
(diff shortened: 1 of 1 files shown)


bobqianic:
Yes, you're right. If anyone wants to run `make stream`, they should do it in the root folder of this repo.

https://github.com/ggerganov/whisper.cpp/blob/54c978c3a31dac51bb6fab06d5ac5b11b336ae30/Makefile#L343-L370

examples/stream/README.md
```diff
@@ -39,6 +39,10 @@ brew install sdl2
 make stream
 ```
 
+Ensure you are at the root of the repo when running `make stream`.  Not within the examples/stream dir
```
bobqianic, line 42:
```suggestion
Ensure you are at the root of the repo when running `make stream`.  Not within the `examples/stream` dir

```

examples/stream/README.md
```diff
@@ -39,6 +39,10 @@ brew install sdl2
 make stream
 ```
 
+Ensure you are at the root of the repo when running `make stream`.  Not within the examples/stream dir
+as the libraries needed like `common-sdl.h` are located within examples.  Attempting to compile within
```
bobqianic, line 43:
```suggestion
as the libraries needed like `common-sdl.h` are located within `examples`.  Attempting to compile within

```

bobqianic (changes requested):

bobqianic (approved):

PR #1404, opened 2023-10-30
Faster download for models on windows using BitTransfer

WhiteOlivierus:
This was very frustrating, it takes so long to download using `Invoke-WebRequest`

### DIFF
--- a/models/download-ggml-model.cmd
+++ b/models/download-ggml-model.cmd
@@ -40,7 +40,7 @@ if exist "ggml-%model%.bin" (
   goto :eof
 )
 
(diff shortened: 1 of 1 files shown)


bobqianic:
> This was very frustrating, it takes so long to download using Invoke-WebRequest

`Invoke-WebRequest` is slow due to a bug in Windows Powershell. 
["Although this has already been fixed in PowerShell, Windows PowerShell most likely will never get that fix."](https://github.com/PowerShell/PowerShell/issues/2138#issuecomment-1264124324)

bobqianic (approved):

PR #1415, opened 2023-11-02
Allow env variable to override metal resource path

codesoda:
Implementation for https://github.com/ggerganov/whisper.cpp/issues/1397

If an ENV variable called `WHISPER_METAL_PATH_RESOURCES` is available, then the `ggml-metal.metal` file is loaded from that path instead of using the `bundle#pathForResource` method.

---

### Testing

🟥  Run main and expect a load (null) error.

Use `make` to compile the `main` binary and move it to a folder outside the project root.

```
$ ./main -m models/ggml-base.bin test.wav
...
ggml_metal_init: allocating
ggml_metal_init: found device: Apple M1
ggml_metal_init: picking default device: Apple M1
ggml_metal_init: loading '(null)'
ggml_metal_init: error: Error Domain=NSCocoaErrorDomain Code=258 "The file name is invalid."
whisper_init_state: ggml_metal_init() failed
error: failed to initialize whisper context
```

🟩  Launch again, passing the project root path, where the `ggml-metal.metal` file exists. Success!

```
$ WHISPER_METAL_PATH_RESOURCES=/Users/codesoda/projects/whisper.cpp ./main  -m models/ggml-base.bin test.wav
...
ggml_metal_init: allocating
ggml_metal_init: found device: Apple M1
ggml_metal_init: picking default device: Apple M1
ggml_metal_init: loading '/Users/codesoda/projects/whisper.cpp/ggml-metal.metal'
ggml_metal_init: loaded kernel_add                            0x155707730 | th_max = 1024 | th_width =   32
...
```

Issue #1397, opened 2023-10-28
Request: Keep ggml-metal.* files in separate directory from main binary


dukeeagle:
Hi everyone! (And I hope you've been doing well since AIG @ggerganov!!)

I'm running whisper.cpp as a [Tauri binary sidecar](https://tauri.app/v1/guides/building/sidecar/). This requires that all secondary files in a separate directory in my Tauri app called `./resources`.

To run whisper.cpp with Metal, it seems like all `ggml-metal.*` files need to have the **same path as the main whisper.cpp binary**. This is incompatible with how Tauri invokes sidecar binaries.

Is that correct? Is there a command line argument that changes this? If not, I'd like to request / contribute to a feature that lets you specify a custom path for your metal files (specifically `ggml-metal.h`, `ggml-metal.m`, `ggml-metal.metal`)

ggerganov:
Hi Luke!

Yes,  currently it requires the `ggml-metal.metal` file to be in the same path as the binary. I've struggled with this logic and it definitely needs some improvement.

I think the easiest change to support your case would be to check an environment variable for a user-specified path if it fails to locate the `ggml-metal.metal` file with the default settings. The var can be called `WHISPER_METAL_PATH_RESOURCES` and should point to the folder that contains `ggml-metal.metal`

Feel free to open a PR. I soon plan to bring `whisper.cpp` up-to-date with latest developments of the `ggml` core library and improving Metal builds and Swift package will be one of the priorities.

codesoda:
Something like #1415 @ggerganov ?

PR #1420, opened 2023-11-03
cmake: Fix cmake builds on macOS.

### LINKED ISSUE #1419: CMake on macOS M2: `The file name is invalid`
Currently, when building with CMake on a M series processors give the following error:
```
ggml_metal_init: error: Error Domain=NSCocoaErrorDomain Code=258 "The file name is invalid."
```

Doing this fixes the error but I'm sure it'll break `make` builds.

```diff
diff --git a/ggml-metal.m b/ggml-metal.m
index 1139ee3..63ac00a 100644
--- a/ggml-metal.m
+++ b/ggml-metal.m
@@ -178,7 +178,7 @@ @implementation GGMLMetalClass
 
         //NSString * path = [[NSBundle mainBundle] pathForResource:@"../../examples/metal/metal" ofType:@"metal"];
         NSBundle * bundle = [NSBundle bundleForClass:[GGMLMetalClass class]];
-        NSString * path   = [bundle pathForResource:@"ggml-metal" ofType:@"metal"];
+        NSString * path   = [bundle pathForResource:@"bin/ggml-metal" ofType:@"metal"];
         metal_printf("%s: loading '%s'\n", __func__, [path UTF8String]);
 
         NSString * src  = [NSString stringWithContentsOfFile:path encoding:NSUTF8StringEncoding error:&error];
```

etra0:
If we have a null path, we then do a last check in `bin/ggml-metal` because that's where the metal file is located in the case of cmake builds.

Fixes #1419.

### DIFF
--- a/ggml-metal.m
+++ b/ggml-metal.m
@@ -179,6 +179,11 @@ struct ggml_metal_context * ggml_metal_init(int n_cb) {
         //NSString * path = [[NSBundle mainBundle] pathForResource:@"../../examples/metal/metal" ofType:@"metal"];
         NSBundle * bundle = [NSBundle bundleForClass:[GGMLMetalClass class]];
         NSString * path   = [bundle pathForResource:@"ggml-metal" ofType:@"metal"];
+        // In cases when we build with CMake, the actual path is
+        // bin/ggml-metal.
+        if (!path) {
+            path = [bundle pathForResource:@"bin/ggml-metal" ofType:@"metal"];
+        }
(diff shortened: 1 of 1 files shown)


codesoda:
@etra0 Have you also had issues with #1397 / #1415 and being unable to find `ggml-metal.metal` on disk in release builds?

ggerganov:
Yup, the proper fix would be #1415 
Hardcoding "bin/..." is a hack on top of an existing hack. With the env variable it should be a bit more generic

etra0:
I agree.

PR #1422, opened 2023-11-03
sync : ggml (backend v2, k-quants, CUDA opts, Metal opts, etc.)

ggerganov:
Bringing back many improvements from `llama.cpp` and `ggml`

These will be used to prepare a new release that will include:

- Efficient Beam Search
- Full CUDA GPU offloading
- K-quants support
- Grammar support (#1229)


nchudleigh:
Testing this on 2015 MBP with i7

PR #1432, opened 2023-11-05
ggml : try to fix the abort mechanism

ggerganov:


mkiol:
With this patch, I'm observing following SIGSEGV when triggering abort via `abort_callback`:

```
==145449==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000058 (pc 0x559badd13f9b bp 0x000000000000 sp 0x7f55ee458c50 T11)
==145449==The signal is caused by a READ memory access.
==145449==Hint: address points to the zero page.
    #0 0x559badd13f9b in ggml_get_n_tasks /home/mkiol/dev/whisper.cpp-abort-fix/ggml.c:16414
    #1 0x559badd36660 in ggml_graph_compute_thread /home/mkiol/dev/whisper.cpp-abort-fix/ggml.c:16756
```

PR #1445, opened 2023-11-07
Decouple Android example into a library and app module

tobrun:
I needed to integrate whisper.cpp project into another project. The current Android example structure didn't allow for this. In this PR, I'm decoupling the Android example into 2 modules:
 - lib: the minimum code to get whisper.cpp integrated into a project (kotlin+JNI+build files)
 - app: sample UI code, Recorder and RiffWaveHelper

Building the lib module will result in creating an AAR binary that can be used in other projects. The old setup of creating an APK or AAB didn't allow for this. 

Happy to follow up with any change request, feel free to close if this isn't a useful addition.


ggerganov:
Thanks! Merging without testing, since the CI seems good

PR #1457, opened 2023-11-08
models : Fix `n_mel` mismatch in convert-whisper-to-coreml.py

bobqianic:
Because `whisper-large-v3` uses `n_mel=128`, we cannot hardcode the `input_shape = (1, 80, 3000)` into the script.

PR #1459, opened 2023-11-08
models : Fix `n_mel` mismatch in convert-whisper-to-openvino.py 

bobqianic:
Because `whisper-large-v3` uses `n_mel=128`, we cannot hardcode the `input_shape = (1, 80, 3000)` into the script.

Same as #1457 #1458

### DIFF
--- a/models/convert-whisper-to-openvino.py
+++ b/models/convert-whisper-to-openvino.py
@@ -9,7 +9,7 @@ import shutil
 def convert_encoder(hparams, encoder, mname):
     encoder.eval()
 
-    mel = torch.zeros((1, 80, 3000))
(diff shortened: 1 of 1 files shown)


ggerganov:
Looks like `whisper_openvino_encode` needs similar update as in #1458

bobqianic:
> Looks like `whisper_openvino_encode` needs similar update as in #1458

I don’t think so.

https://github.com/ggerganov/whisper.cpp/blob/0de8582f65bde052b08b1844c39fed8808f75619/openvino/whisper-openvino-encoder.cpp#L81-L88

ggerganov (approved):

PR #1462, opened 2023-11-09
Set default `temperature_inc` to 0.2f

bobqianic:
See #1461 for details


bobqianic:
Since it has been implemented in #1486, it can now be closed.

PR #1466, opened 2023-11-09
whisper : migrating to ggml-backend

ggerganov:
WIP

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -715,6 +715,11 @@ struct whisper_state {
     int32_t n_fail_p = 0; // number of logprob threshold failures
     int32_t n_fail_h = 0; // number of entropy threshold failures
 
+    // TODO: free this
+    struct ggml_context * ctx;
+
+    ggml_backend_buffer_t buffer;
+
     // cross-attention KV cache for the decoders
     // shared between all decoders
     whisper_kv_cache kv_cross;
@@ -725,9 +730,6 @@ struct whisper_state {
     // buffer for swapping KV caches between decoders during beam-search
     std::vector<kv_buf> kv_swap_bufs;
 
-    // reusable buffer for `struct ggml_graph_plan.work_data`
-    std::vector<uint8_t> work_buffer;
-
     // backend schedulers
     ggml_backend_sched_t sched_conv;
     ggml_backend_sched_t sched_encode;
@@ -735,8 +737,10 @@ struct whisper_state {
     ggml_backend_sched_t sched_decode;
 
     // result of the encoder
-    struct ggml_tensor * embd_conv = nullptr;
-    struct ggml_tensor * embd_enc  = nullptr;
+    struct ggml_tensor * inp_mel      = nullptr;
+    struct ggml_tensor * inp_kq_scale = nullptr;
+    struct ggml_tensor * out_conv     = nullptr;
+    struct ggml_tensor * out_enc      = nullptr;
 
     // decode output (2-dimensional array: [n_tokens][n_vocab])
     std::vector<float> logits;
@@ -1546,7 +1550,8 @@ static struct ggml_cgraph * whisper_build_graph_conv(
 
     ggml_cgraph * gf = ggml_new_graph(ctx0);
 
-    struct ggml_tensor * mel = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, 2*n_ctx, n_mels);
+    //struct ggml_tensor * mel = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, 2*n_ctx, n_mels);
+    struct ggml_tensor * mel = wstate.inp_mel;
     //ggml_allocr_alloc(alloc, mel);
 
     //assert(mel->type == GGML_TYPE_F32);
@@ -1588,9 +1593,8 @@ static struct ggml_cgraph * whisper_build_graph_conv(
                     cur);
 
             cur = ggml_gelu(ctx0, cur);
+            ggml_set_name(cur, "out_conv");
         }
-
-        wstate.embd_conv = cur;
     } else {
 #ifdef WHISPER_USE_COREML
         cur = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, n_state, n_ctx);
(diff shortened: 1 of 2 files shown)


whisper.cpp
```diff
@@ -1636,24 +1630,25 @@ static struct ggml_cgraph * whisper_build_graph_encoder(
...
+    //ggml_allocr_alloc(alloc, KQscale);
 
-    if (!ggml_allocr_is_measure(alloc)) {
-        ggml_set_f32(KQscale, 1.0f/sqrt(float(n_state)/n_head));
-    }
+    //if (!ggml_allocr_is_measure(alloc)) {
+    //    ggml_set_f32(KQscale, 1.0f/sqrt(float(n_state)/n_head));
+    //}
 
     struct ggml_tensor * cur = ggml_view_tensor(ctx0, wstate.embd_conv);
```
ggerganov, line 1653: @slaren

I'm migrating `whisper.cpp` to use the new scheduler and backend API.

Here, if I want to use the result of a previous graph (`wstate.embd_conv`) as an input, what would be the proper way to do it? If I allocated `wstate.embd_conv` in a dedicated input buffer, then do I need to change the following:

https://github.com/ggerganov/whisper.cpp/blob/005b8ccbf0288975985ef15082579b34cadbf36a/whisper.cpp#L1592-L1593

to `ggml_cpy(ctx0, cur, wstate.embd_conv);`? If yes, is there a way to avoid the copy?

slaren: It is fine to use a pre-allocated tensor, I don't think that you need to make a copy here. Conceptually this should be the same as using a weight pre-allocated in another buffer, right? Maybe I am missing something.

slaren: To elaborate more: when using `ggml-backend` directly without `ggml-backend-sched`, all the tensors must be allocated in the same backend. If you wanted to evaluate a graph in a backend, and then use the output of the graph as in input of a different graph evaluated on a different backend, you would need to use `ggml_backend_tensor_copy` to copy the tensor to the other backend. When using `ggml-backend-sched`, these copies should be done automatically.

slaren:
The messages such as `!!!!!!! node_0 has no backend` indicate a failure while splitting the graph. There are some cases that may not be handled correctly by `ggml-backend-sched` yet. Do you intend to add support for partial offloading? otherwise, it may be simpler to just use `ggml-backend` without the scheduler.

whisper.cpp
ggerganov: >  Conceptually this should be the same as using a weight pre-allocated in another buffer, right?

Yes. Currently it errors that `wstate.embd_conv` does not have a backend, but I think I see the problem. Will try to fix it

Another question: here I am creating a separate scheduler for each separate graph:

https://github.com/ggerganov/whisper.cpp/blob/005b8ccbf0288975985ef15082579b34cadbf36a/whisper.cpp#L2865-L2899

Is there a way to do this with a single scheduler?

ggerganov:
There will be no support for partial offloading, but I thought it might be useful to exercise the new functionality. I think the `has no backend` error currently is because I'm not pre-allocating the input `mel` tensor for the `conv` graph, but still not sure. Will give this a few more tries, and might go back to `ggml-backend` if I can't make it work

whisper.cpp
slaren: No, as it is currently, each graph must use a different scheduler. Maybe this is something that we could look into adding in the future, but currently all the allocations happen during the call to `ggml_backend_sched_init_measure`, and there is no way to add other graphs.

ggerganov:
This version produces addr sanitizer from time to time during the schedulers initialization:

```bash
$ ▶ make -j && ./bin/main -m ../models/ggml-base.en.bin -f ../samples/jfk.wav -ng
[ 43%] Built target whisper
[ 56%] Built target bench
[ 75%] Built target common
[100%] Built target main
[100%] Built target quantize
whisper_init_from_file_with_params_no_state: loading model from '../models/ggml-base.en.bin'
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51864
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 512
whisper_model_load: n_audio_head  = 8
whisper_model_load: n_audio_layer = 6
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 512
whisper_model_load: n_text_head   = 8
whisper_model_load: n_text_layer  = 6
whisper_model_load: n_mels        = 80
whisper_model_load: ftype         = 1
whisper_model_load: qntvr         = 0
whisper_model_load: type          = 2 (base)
whisper_model_load: adding 1607 extra tokens
whisper_model_load: n_langs       = 99
whisper_model_load:      CPU buffer size =   140.66 MB
whisper_model_load: model size    =  140.54 MB
whisper_init_state: kv self size  =    5.25 MB
whisper_init_state: kv cross size =   17.58 MB
whisper_init_state: backend_in = CPU (7105540 bytes)
ggml_backend_sched size: 1468 KB
whisper_init_state: compute buffer (conv)   =   16.02 MB
ggml_backend_sched size: 1468 KB
whisper_init_state: compute buffer (encode) =   80.38 MB
ggml_backend_sched size: 1468 KB
whisper_init_state: compute buffer (encode) =    2.93 MB
ggml_backend_sched size: 1468 KB
/Users/ggerganov/development/github/whisper.cpp/ggml-backend.c:521:25: runtime error: index 12289 out of bounds for type 'char[12288][128]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/ggerganov/development/github/whisper.cpp/ggml-backend.c:521:25 in 
whisper_init_state: compute buffer (encode) =   23.13 MB

system_info: n_threads = 4 / 24 | AVX = 0 | AVX2 = 0 | AVX512 = 0 | FMA = 0 | NEON = 1 | ARM_FMA = 1 | METAL = 1 | F16C = 0 | FP16_VA = 1 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 0 | SSSE3 = 0 | VSX = 0 | COREML = 0 | OPENVINO = 0 | 

main: processing '../samples/jfk.wav' (176000 samples, 11.0 sec), 4 threads, 1 processors, lang = en, task = transcribe, timestamps = 1 ...
```

Will leave it for now and try without the scheduler in a separate branch.

slaren:
The out of bounds access is in the `causes` buffer which is just for debug, and might be too small for the whisper graph. The size can be increased as much as needed, it should be removed soon. But if you don't need to support partial offloading there is nothing to gain by using the scheduler anyway.

PR #1467, opened 2023-11-09
Add '-l auto' to talk-llama example

kubaracek:
Adds the possibility to select 'auto' as `lang` option for `talk-llama` example

### DIFF
--- a/examples/talk-llama/talk-llama.cpp
+++ b/examples/talk-llama/talk-llama.cpp
@@ -248,7 +248,7 @@ int main(int argc, char ** argv) {
         return 1;
     }
 
-    if (whisper_lang_id(params.language.c_str()) == -1) {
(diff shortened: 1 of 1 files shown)


examples/talk-llama/talk-llama.cpp
```diff
@@ -248,7 +248,7 @@ int main(int argc, char ** argv) {
         return 1;
     }
 
-    if (whisper_lang_id(params.language.c_str()) == -1) {
+    if (params.language != "auto" && whisper_lang_id(params.language.c_str()) == -1){
```
ggerganov, line 251:
```suggestion
    if (params.language != "auto" && whisper_lang_id(params.language.c_str()) == -1) {
```

ggerganov (approved):

PR #1472, opened 2023-11-10
whisper : add full CUDA and Metal offloading

ggerganov:
Build with:

```bash
# make
WHISPER_CUBLAS=1 make -j

# cmake
cmake -DWHISPER_CUBLAS=1 ../
```

Also, the convolution ops are now offloaded both with CUDA and Metal resulting in speed-up in the Encoder (#1473)
Credits and huge thanks to @FSSRepo: https://github.com/ggerganov/ggml/pull/564

If you want to have some fun, try this:

```bash
# get the models:
./models/download-ggml-model.sh medium.en
wget https://huggingface.co/TheBloke/Llama-2-13B-GGUF/resolve/main/llama-2-13b.Q4_K_M.gguf -O ./models/llama-13b-v2-q4_k_m.gguf

---

# NVIDIA (16GB VRAM required)
WHISPER_CUBLAS=1 make -j talk-llama
./talk-llama -mw ./models/ggml-medium.en.bin -ml ./models/llama-13b-v2-q4_k_m.gguf -p "John" -t 8

---

# Apple (Metal)
make -j talk-llama
./talk-llama -mw ./models/ggml-medium.en.bin -ml ./models/llama-13b-v2-q4_k_m.gguf -p "John" -t 8

---

# Apple (CoreML + Metal)
./models/generate-coreml-model.sh medium.en
WHISPER_COREML=1 make -j talk-llama
./talk-llama -mw ./models/ggml-medium.en.bin -ml ./models/llama-13b-v2-q4_k_m.gguf -p "John" -t 8
```

---

Bench on V100 and M2 Ultra

```bash
./extra/bench-all.sh 
Usage: ./bench.sh [n_threads] [encoder-only]

Running memcpy benchmark

memcpy: 9.55 GB/s (1 thread)
sum:    -536869898.000000

Running ggml_mul_mat benchmark with 4 threads

ggml_init_cublas: GGML_CUDA_FORCE_MMQ:   no
ggml_init_cublas: CUDA_USE_TENSOR_CORES: yes
ggml_init_cublas: found 1 CUDA devices:
  Device 0: Tesla V100-PCIE-16GB, compute capability 7.0
  64 x   64: Q4_0     2.6 GFLOPS (128 runs) | Q4_1     2.6 GFLOPS (128 runs)
  64 x   64: Q5_0     2.7 GFLOPS (128 runs) | Q5_1     2.7 GFLOPS (128 runs) | Q8_0     2.7 GFLOPS (128 runs)
[...]

PR #1479, opened 2023-11-11
whisper : try to fix the parallel whisper_state functionality

ggerganov:
ref #1472 

```bash
./main -m ./models/ggml-base.en.bin -f ./samples/gb0.wav -p 2
```

- [x] CPU
- [x] Metal
- [x] CUDA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .DS_Store
 
 build/
+build-coreml/
 build-em/
 build-debug/
 build-release/
--- a/ggml-metal.h
+++ b/ggml-metal.h
@@ -26,7 +26,7 @@
 #include <stdbool.h>
 
 // max memory buffers that can be mapped to the device
-#define GGML_METAL_MAX_BUFFERS 16
+#define GGML_METAL_MAX_BUFFERS 64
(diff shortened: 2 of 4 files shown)


slaren:
I will change ggml-backend to decouple the buffers from backend instances, but that is still going to require changing the metal backend such that the list of buffers is available to all the instances of the metal backend. In the long term, I think it would be better to get rid of the list of buffers entirely by making the `ggml_backend_buffer` of metal more an abstraction of the metal buffer object (I think it is `MTLBuffer`). That may also help fix issues that people with intel macs with discrete GPUs have with the metal backend, since it wouldn't assume unified memory.

ggerganov:
> In the long term, I think it would be better to get rid of the list of buffers entirely by making the ggml_backend_buffer of metal more an abstraction of the metal buffer object (I think it is MTLBuffer).

Yes, will probably reimplement this when I start working on https://github.com/ggerganov/llama.cpp/issues/3229. Planning to do various improvements in the Metal backend in addition to that.

For now, I've implemented a workaround where in `ggml_metal_get_buffer` we check if the tensor belongs to a backend buffer and if it does, we search the list of buffers for that backend. Otherwise, we fallback to the old logic. This seems to resolve the issue for now.

Issue #375, opened 2023-01-05
Implement word-level timestamps approach proposed by OpenAI


ggerganov:
See notebook, section "Word-level timestamps using attention weights":

https://github.com/openai/whisper/blob/main/notebooks/Multilingual_ASR.ipynb

djthorpe:
This came up in my "explore" feed as a way to implement accurate word-level timestamps:
https://github.com/m-bain/whisperX

kamranjon:
There is a functioning implementation of the attention weights approach here: https://github.com/linto-ai/whisper-timestamped which might be a useful reference for implementing in `whisper.cpp` eventually.

HaujetZhao:
The whisper python module itself provided a time-stamp output option, which could be a reference, and I tested it, the command is: 

```
python -m whisper --model tiny --language en --word_timestamps True --output_dir "test_out" "test.wav"
```

it generated 5 files in the test_out folder: 

```
test.json
test.txt
test.srt
test.vtt
test.tsv
```

In the test.json file, the content is: 

```json
[...]

PR #1485, opened 2023-11-13
[DRAFT] Token level timestamps with DTW (#375) 

denersc:
Tries to solve #375 

Attempt to implement DTW-based token level timestamps, as seen in [OpenAI Whisper](https://github.com/openai/whisper/blob/1cea4357687b676b293cb5473e1ade25f5b1cef7/whisper/timing.py#L163). 

This first commit implements the DTW algorithm on whisper.cpp and provides tests that compare the output with the output of OpenAI's implementation. Tests are done calling whisper.cpp from Python and comparing DTW output with OpenAI's `dtw_cpu` function.

An outline of remaining work is commented on `whisper_exp_compute_token_level_timestamps_dtw` in whisper.cpp. Help/insights are very appreciated, specially concerning how to cache/retrieve the output of MHA layers that are used as input for DTW.

In OpenAI's implementation, token-level timestamps are used with further heuristics to determine a supposed start/end time for words. In this PR, my intention is to implement token-level only as a first step that can be used to implement word timestamps in the future. 

### TODO

  - [x] Implement/test DTW algorithm
  - [x] Import [index of alignment heads ](https://github.com/openai/whisper/blob/1cea4357687b676b293cb5473e1ade25f5b1cef7/whisper/__init__.py#L34) from original whisper. 
  - [x] Find a way to cache ~~the output of the MHA layers~~ QKs from cross-attention layers from alignment heads (perhaps in whisper_state?) and retrieve them in `whisper_exp_compute_token_level_timestamps_dtw`
  - [x] Use GGML to pre-process QKs that are passed to DTW. Includes stacking cached QKs, scaling, clipping, normalizing. See `whisper_exp_compute_token_level_timestamps_dtw`
  - [x] Plug `whisper_exp_compute_token_level_timestamps_dtw` into `whisper_full` and use results to place timestamps on each inferred token
  - [x] Plumbing in general, enable/disable as whisper_full_param, etc
  - [x] Implement `N_TOP_MOST` alignment heads 
  - [x] Implement `CUSTOM` alignment heads, decide comfortable API for setting custom alignment heads
  - [x] Find a way to make `whisper_build_graph_decoder` to only save QK copies if requested, so there is no additional overhead when running decoder for other reasons than timestamps.
  - [ ] Avoid memory allocations on `whisper_exp_compute_token_level_timestamps_dtw` (probably allocate a buffer for the used tensors on init)
  - [ ] (Maybe) check if some operations on `whisper_exp_compute_token_level_timestamps_dtw`  that are currently done with manual for loops can benefit from ggml functions.


bmurray:
This is awesome. I think the big question is where the alignment heads actually reside in GGML.

denersc:
> This is awesome. I think the big question is where the alignment heads actually reside in GGML.

Yes, that is the most critical point at the moment. I suspect that we would need to save this tensor:

```c++
// Inside whisper_build_graph_decoder
2448             // K * Q
2449             struct ggml_tensor * KQ = ggml_mul_mat(ctx0, Kcross, Q);   // THIS ONE
2450 
2451             //struct ggml_tensor * KQ_scaled =
2452             //    ggml_scale(ctx0,
2453             //            KQ,
2454             //            ggml_new_f32(ctx0, 1.0f/sqrt(float(n_state)/n_head))
2455             //            );
2456 
2457             // no masking for cross-attention
2458             //struct ggml_tensor * KQ_masked = ggml_diag_mask_inf(ctx0, KQ_scaled, n_past);
2459 
2460             struct ggml_tensor * KQ_soft_max = ggml_soft_max(ctx0, KQ); // OR THIS ONE
```

In OpenAI implementation, they do a second pass through the model passing the tokens generated on the first pass to be able to retrieve the correct weights. I think if we are able to that second pass just like in openAI impl, it should just be a matter of saving those KQs and selecting the ones useful for timestamping. 

Apparently [lintoai whisper-timestamped](https://github.com/linto-ai/whisper-timestamped) seems to be able to do in one pass, only when there is no temperature fallback, greedy bestof = 1 and no beam search. If any of those conditions are not met, then it resorts to doing a second pass like on OpenAI impl.

In what concerns of selecting heads that are useful for timestamping, from what i have seen, a accepted default would be the heads of the top most half of decoder layers (the ones closer to model output). A more optimal selection, which is provided on OpenAIs whisper individually by model, was [apparently obtained by manual inspection  ](https://github.com/openai/whisper/discussions/1733). Note that alignment heads differ even for models with same dimensions (e.g. [medium and medium.en have different alignment heads specified](https://github.com/openai/whisper/blob/1cea4357687b676b293cb5473e1ade25f5b1cef7/whisper/__init__.py#L34))

We could still use these indexes for the usual pre-trained models but offer the option to use the n top-most decoder layers on some custom index i guess.

linmi:
Looking forward to the progress 🎉

denersc:
Haven't been able to work with this past week, but making some progress now!

Trying to get a very poorly implemented end-to-end POC for the `base.en` model. Having some trouble figuring out if i'm correctly generating/retrieving the needed cross-attention QKs

Validated most of the pre-processing operations done on QKs before passing then to DTW to get timestamps. So, taking the OpenAI implementation on [timing.py](https://github.com/openai/whisper/blob/e58f28804528831904c3b6f2c0e473f346223433/whisper/timing.py#L163): (with added comments by me)

```python

    # Implemented on whisper.cpp, seems ok
    tokens = torch.tensor(
        [
            *tokenizer.sot_sequence,
            tokenizer.no_timestamps,
            *text_tokens,
            tokenizer.eot,
        ]
    ).to(model.device)

    # I believe that i am making a mistake somewhere in this block
    # install hooks on the cross attention layers to retrieve the attention weights
    QKs = [None] * model.dims.n_text_layer
    hooks = [
        block.cross_attn.register_forward_hook(
            lambda _, ins, outs, index=i: QKs.__setitem__(index, outs[-1][0])
        )
        for i, block in enumerate(model.decoder.blocks)
    ]
    with torch.no_grad():
        logits = model(mel.unsqueeze(0), tokens.unsqueeze(0))[0]
        sampled_logits = logits[len(tokenizer.sot_sequence) :, : tokenizer.eot]
        token_probs = sampled_logits.softmax(dim=-1)
        text_token_probs = token_probs[np.arange(len(text_tokens)), text_tokens]
        text_token_probs = text_token_probs.tolist()

    for hook in hooks:
        hook.remove()

    # Implemented poorly in whisper.cpp (only using base.en alignment heads for now)
    # heads * tokens * frames
    weights = torch.stack([QKs[_l][_h] for _l, _h in model.alignment_heads.indices().T])
    weights = weights[:, :, : num_frames // 2]
    weights = (weights * qk_scale).softmax(dim=-1)

    # Implemented and validated on whisper.cpp 
    std, mean = torch.std_mean(weights, dim=-2, keepdim=True, unbiased=False)
    weights = (weights - mean) / std
    weights = median_filter(weights, medfilt_width)
    matrix = weights.mean(axis=0)
    matrix = matrix[len(tokenizer.sot_sequence) : -1]
    text_indices, time_indices = dtw(-matrix)
```

Now, to get the whisper.cpp QKs, i temporarily added a `decoder_save_cross_QKs` bool to `whisper_state`. If enabled, it will save QKs during decoding. So:

```c
// Inside whisper_build_graph_decoder
struct ggml_tensor * KQ_soft_max = ggml_soft_max(ctx0, KQ); // THIS TENSOR HERE!
if (wstate.decoder_save_cross_QKs) {
    wstate.cross_QKs.push_back(KQ_soft_max);
}


// After inference, when computing timestamps 
state->decoder_save_cross_QKs = true;
if (whisper_decode_with_state(ctx, state, tokens.data(), tokens.size(), 0, n_threads) != 0) { 
    WHISPER_LOG_INFO("DECODER FAILED\n");
}    
state->decoder_save_cross_QKs = false;
```

Doing this, i get a set of QKs to work with inside my timestamping function, and the amount of tensors retrieved and their dimensions are in line with what's retrieved in the openAI impl.

Now, since the output i got so far for timestamps is complete garbage, and most of the operations outside retrieval of QKs seem correct, i imagine something is wrong when i call `whisper_decode_with_state`. I am unsure if i need to do additional setup before calling `whisper_decode_with_state` or if the `ggml_tensor` pointers i saved are invalid or do not contain what i expect.

Any insight into how to correctly retrieve these QKs from decoder cross attention layers is very welcome! I'm currently on working on validating all operations other than QK retrieval to be absolutely sure this is where my mistake is.

denersc:
Currently stuck on retrieving the attention weights from the decoder. In all my attempts, i either get tensors with floats > 1 (which indicates that they are not the output of the softmax layer i'm trying to retrieve), null tensors, or tensors with null "data" pointer.

What i have tried so far (inside `whisper_build_graph_decoder`):

- Directly pushing `KQ_soft_max` into a vector on the state yields a tensor that does not seem like the output of softmax (floats > 1)

```c
// Inside whisper_build_graph_decoder
struct ggml_tensor * KQ_soft_max = ggml_soft_max(ctx0, KQ);
wstate.cross_QKs.push_back(KQ_soft_max);


// Later on, e.g. on timestamping function
// Many values are > 1, indicating that they are not the output of softmax
float v = ggml_get_f32_nd(state.cross_QKs[0], 0, 0, 0, 0);
```

- Using ggml_cpy on KQ_soft_max hoping that a copy of it would be later retrievable. This yields a tensor whose `data` field points to NULL

```c
// Inside whisper_build_graph_decoder
struct ggml_tensor * KQ_soft_max = ggml_soft_max(ctx0, KQ);
struct ggml_tensor * KQ_copy = ggml_cpy(ctx0, KQ_soft_max, ggml_new_tensor_3d(ctx0, GGML_TYPE_F32, KQ_soft_max->ne[0], KQ_soft_max->ne[1], KQ_soft_max->ne[2]));
wstate.cross_QKs.push_back(KQ_copy);

// Later on, e.g. on timestamping function
if (state.cross_QKs[0]->data == NULL)
  // this is true
```

- Permutations of the the above, but instead of pushing these tensors to a vector, setting a name on them and trying to retrieve them by name right after `ggml_graph_compute_helper` inside `whisper_decode_internal`. This yields either a tensor that is clearly not softmax output (values > 1) if done directly on KQ_soft_max or a a null pointer if done on the copy:

```c
// Inside whisper_build_graph_decoder
struct ggml_tensor * KQ_soft_max = ggml_soft_max(ctx0, KQ);

struct ggml_tensor * KQ_copy = ggml_cpy(ctx0, KQ_soft_max, ggml_new_tensor_3d(ctx0, GGML_TYPE_F32, KQ_soft_max->ne[0], KQ_soft_max->ne[1], KQ_soft_max->ne[2]));
char name[256];
snprintf(name, 256, "cross_QK_%d", il);
ggml_set_name(KQ_soft_max, name);
snprintf(name, 256, "cross_QK_copy_%d", il);
ggml_set_name(KQ_copy, name);


// Later on, right after ggml_graph_compute_helper inside whisper_decode_internal
ggml_graph_compute_helper(wstate.backend, gf, n_threads);

struct ggml_tensor * cross_QK = ggml_graph_get_tensor(gf, "cross_QK_0");
struct ggml_tensor * cross_copy = ggml_graph_get_tensor(gf, "cross_QK_copy_0");
float v = ggml_get_f32_nd(cross_QK, 0, 0, 0, 0) // v > 1, not softmax output
if (cross_copy == NULL)
  // this is true
```

@ggerganov would you be able to nudge me in right direction in how would i go about saving the values of the `KQ_soft_max` cross-attention tensor inside the `whisper_build_graph_decoder` function so i can re-use them for timestamping? I've been reading ggml source code but so far haven't been able to figure it out. Thanks!

denersc:
I've committed these recent results into the fork. Currently, the timestamps are only being placed if you run with `params.single_segment=true`, as there is some more work to get everything aligned in the case a single inference run produces more than one segment. 

To run with DTW timestamps, you need to enable then in params and select a collection of alignment heads. I've imported those on OpenAI impl for each model, but have yet to implement setting custom alignment heads or using the N top most heads for alignment.

So, e.g. for `main.cpp`, you need to set

```diff
diff --git a/examples/main/main.cpp b/examples/main/main.cpp
index 9699802..c101434 100644
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -942,6 +942,10 @@ int main(int argc, char ** argv) {
 
             wparams.strategy = params.beam_size > 1 ? WHISPER_SAMPLING_BEAM_SEARCH : WHISPER_SAMPLING_GREEDY;
 
+            wparams.dtw_token_timestamps = true;
+            wparams.dtw_ah_preset = WHISPER_AHEADS_BASE_EN;
+            wparams.single_segment = true;
+
             wparams.print_realtime   = false;
             wparams.print_progress   = params.print_progress;
             wparams.print_timestamps = !params.no_timestamps;
```

and timestamps should be placed on the `token_data` struct for each token, in the field `t_dtw`. Chose the `dtw_ah_preset` field to correctly match the model you are using.

```cpp
auto token_data =  whisper_full_get_token_data(ctx, i_segment, i_token);
float timestamp_seconds = (float)token_data.t_dtw/100;
```

I've left several FIXMEs on `whisper.cpp` and `whisper.h` outlining some of the work needed to conclude this implementation. I'll be updating the TODO list on this PR soon to it reflects the current issues.

denersc:
So, i've updated the TODO list and pushed some fixes:

- Setting `single_segment` is no longer needed. Timestamps should work correctly if an inference run hast multiple segments.
- Fixed compile error related to `whisper_alignment_heads_preset` on whisper_full_params that occurred in some compilers.
- Fixed a incorrect assertion that would cause DTW timestamp function to break if you didn't set `params.audio_ctx` to some values
- Fixed a detail in timestamping final loop that would cause gradual misalignment of timestamps on longer segments.

Any help in the current TODO list is very welcome, specially the last 3 items.

I've made some tests with the `a13.wav` audio, will continue validating with other cases. Any testing is highly appreciated!

denersc:
Just an update, haven't been able to work the last few issues lately and probably won't be completing in December. Nevertheless, I've been doing some tests with whats implemented so far and timestamps seem to be working as expected.

About the last couple of issues, what I'm having most trouble figuring it out is "Avoid memory allocations on whisper_exp_compute_token_level_timestamps_dtw".

I'm not very confident on how to retrieve the total memory needed in each case, since some operations are done "manually" over ggml tensors instead of using a ggml graph.

Also, the memory needed can vary greatly - like from <16MB to >2GB - depending on factors such as how many tokens the model yielded this run, audio_ctx size and the number of alignment heads (which is limited by numbers of attention heads in total). This can all be taken into account when allocating to try to allocate the minimum necessary for the worst case.

Any help on these final issues would be very appreciated!

denersc:
Hey @bmurray, thanks for testing! I'll try to run this audio here as well. May i ask which model size and alignment head preset were you using?

About that first token with wrong timestamp, i will try to test it on openAI impl as well to see their DTW output. It might be the case that they get the same output and fix that on the subsequent heuristics after running DTW. If you check their implementation, what we have implemented [stops about here](https://github.com/openai/whisper/blob/ba3f3cd54b0e5b8ce1ab3de13e32122d0d5f98ab/whisper/timing.py#L214C15-L214C15). After returning from that function, [they still do some stuff that seems to be especially aimed to improving timestamps on bounds ](https://github.com/openai/whisper/blob/ba3f3cd54b0e5b8ce1ab3de13e32122d0d5f98ab/whisper/timing.py#L298)

About model auto-detect, I'm not sure how we can implement that. Since alignment heads are different for models with same size (e.g. large-v1 and large-v2 have different alignment heads), we might need user input to guarantee we are using the correct alignment heads. Not sure if I'm mistaken though. 

Just a PS, I'll be out after this week and probably return to this only start/mid January, but I'll happily address any comments when i return.

whisper.cpp
```diff
@@ -6610,6 +6689,417 @@ static void whisper_exp_compute_token_level_timestamps(
...
+    struct ggml_tensor * cost = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, N + 1, M + 1);
+    struct ggml_tensor * trace = ggml_new_tensor_2d(ctx, GGML_TYPE_I32, N + 1, M + 1);
+
+    cost = ggml_set_f32(cost, INFINITY);
+    trace = ggml_set_f32(trace, -1);
+    ggml_set_f32_nd(cost, 0, 0, 0, 0, 0.0);
+
+    // dtw
+    // supposedly can be optmized by computing diagonals in parallel ?
+    // Not sure it is worth it since x will be GENERATED_TOKENS*1500 size at most.
```
bobqianic, line 6713: How long does this step take?

denersc:
Hey, got back this week and had to catch up to other things, but getting back to this now.

@bmurray i had a go at the "Queen 'We will meet again' speech" audio. Here are the timestamps i got for the first segment:

base.en:
```
|[_BEG_]|(-0.01) | I|(11.16) | am|(12.56) | speaking|(12.80) | to|(13.18) | you|(13.38) | at|(13.64) | what|(13.76) | I|(13.98) | know|(14.20) | is|(14.62) | an|(14.74) | increasingly|(15.18) | challenging|(16.00) | time|(16.48) |,|(17.78) | a|(18.02) | time|(18.30) | of|(18.48) | disruption|(18.86) |[_TT_958]|(-0.01) 
```

medium:
```
|[_BEG_]|(-0.01) | I|(12.42) | am|(12.68) | speaking|(13.04) | to|(13.32) | you|(13.58) | at|(13.78) | what|(13.92) | I|(14.18) | know|(14.50) | is|(14.74) | an|(15.06) | increasingly|(15.56) | challenging|(16.22) | time|(16.80) |,|(18.04) | a|(18.24) | time|(18.44) | of|(18.74) | disruption|(19.18) |[_TT_959]|(-0.01) 
```

large-v2:
```
|[_BEG_]|(-0.01) | I|(12.44) | am|(12.58) | speaking|(12.98) | to|(13.24) | you|(13.48) | at|(13.70) | what|(13.86) | I|(14.06) | know|(14.40) | is|(14.68) | an|(14.84) | increasingly|(15.42) | challenging|(16.04) | time|(16.62) |.|(17.74) | A|(18.08) | time|(18.40) | of|(18.58) | disruption|(19.04) |[_TT_958]|(-0.01)     
```

I ran with:

```
./main -f queen.wav -m <model>
```

In each case, i added these to whisper params, changing `dtw_ah_preset` accordingly:

```diff
+        wparams.dtw_token_timestamps = true;
+        wparams.dtw_ah_preset = WHISPER_AHEADS_BASE_EN; // Match correctly with the model you are using.
```

Notably there is difference between timestamps retrieved in each case. Which model were you running? Also, i synced my fork with all whisper.cpp latest commits. Maybe something has changed for better?

whisper.cpp
denersc: So, i've just made some measurements and it is quite rather insignificant compared to the rest of the function. 

Reporting from a Apple M2 and Metal enabled.

Just giving a very rough estimate, the total time in `whisper_exp_compute_token_level_timestamps_dtw`  seems to be dominated by the call to `whisper_decode_with_state`, followed by the call to `median_filter` and much less time on the rest.

On `medium` model, in average total function time for each 30 seconds of transcription is ~110mS. About 80mS goes to `whisper_decode_with_state`, 20mS to `median_filter` and 1mS to `dtw_and_backtrace`. 

On `large-v2` model, in average total function time for each 30 seconds of transcription is  ~230mS. About 130mS goes to `whisper_decode_with_state`, 85mS to `median_filter`, 1mS on `dtw_and_backtrace`.

Time in `dtw_and_backtrace` is practically governed by the number of tokens produced in a inference run and is independent of model size, so even in a extreme case where a run produced a lot of tokens, i don't think it would scale much compared to other times.

`whisper_decode_with_state` scales with model size. 

Most other operations in `compute_token_level_timestamps_dtw`  scale by a factor of N_TOKENS*N_ALIGNMENT_HEADS. Regarding alignment heads, the preset with most defined is the preset for large_v2, to it can be a good worst case benchmark. Nevertheless, a user setting custom alignment may use much more and it would surely heavily influence execution time.

denersc:
Regarding general time taken to timestamp each 30 second inference run, reporting from a Apple M2 and Metal enabled:

Just giving a very rough estimate, the total time in `whisper_exp_compute_token_level_timestamps_dtw`  seems to be dominated by the call to `whisper_decode_with_state`, followed by the call to `median_filter` and much less time on the rest.

On `medium` model, in average total function time for each 30 seconds of transcription is ~110mS. About 80mS goes to `whisper_decode_with_state`, 20mS to `median_filter` and 1mS to `dtw_and_backtrace`. 

On `large-v2` model, in average total function time for each 30 seconds of transcription is  ~230mS. About 130mS goes to `whisper_decode_with_state`, 85mS to `median_filter`, 1mS on `dtw_and_backtrace`.

Time in `dtw_and_backtrace` is practically governed by the number of tokens produced in a inference run and is independent of model size, so even in a extreme case where a run produced a lot of tokens, i don't think it would scale much compared to other times.

`whisper_decode_with_state` scales with model size. 

Most other operations in `compute_token_level_timestamps_dtw`  scale by a factor of N_TOKENS*N_ALIGNMENT_HEADS. Regarding alignment heads, the preset with most heads defined is the preset for large_v2, so it can be a reasonable "worst case" benchmark. Nevertheless, a user setting custom alignment heads or n_top_most may use much more heads and it would surely heavily influence execution time.

These times reported will vary depending on how much speech and how fast speech is in a certain audio piece. Very fast speech will produce more tokens thus increase function time generally. Inverse for very slow speech.

mahendra0203:
@denersc I would like to work on this. What can I do to help ?

denersc:
Hey @mahendra0203, cool. 

If you have any suggestions regarding the TODO list on the top post, I'm all ears.

There is a some high-level plumbing work that would facilitate users to use this new feature, such as including an option for it in `main.cpp` command line or on other examples. Also, updating language bindings to enable the use of DTW timestamps. **But i don't think we should do this immediately** since i highly suspect that the user-facing interface to this feature will still change. So i think it would be more efficient to do this later on;

My other suggestion, but this would be more lengthy/challenging: What i implemented only returns a token-level timestamp that more or less corresponds to the "end" time of that token. OpenAI implementation uses this information + some heuristics to determine actual start/end of each word. Implementing something like what is done by openAI would surely make this feature more useful to users. If you would like to tackle that, it would involve basically implementing something like [the rest of this function](https://github.com/openai/whisper/blob/ba3f3cd54b0e5b8ce1ab3de13e32122d0d5f98ab/whisper/timing.py#L214)  and then [the rest of this one](https://github.com/openai/whisper/blob/ba3f3cd54b0e5b8ce1ab3de13e32122d0d5f98ab/whisper/timing.py#L299). This could be done on a new function called right after the call to `whisper_exp_compute_token_level_timestamps_dtw`

If this is the case, I'm unsure how to organize this, but i think it may be better if we merge this MR first and then create a new one for this second step.

denersc:
Regarding these last tasks:

> - Find a way to make whisper_build_graph_decoder to only save QK copies if requested, so there is no additional overhead when running decoder for other reasons than timestamps.
> - Avoid memory allocations on whisper_exp_compute_token_level_timestamps_dtw (probably allocate a buffer for the used tensors on init)

I've given it some though and i think this is only possible if DTW enable/disable and alignment head setting occur before running `whisper_init_...` . So probably these settings should be available in `whisper_context_params` instead of `whisper_full_params`.

Since enabling DTW incurs in additional memory allocation and a small change in the decoder graph, it has to be done within this `init`.

Regarding how much memory to pre-allocate, i haven't figured out how to get the exact number, but I'm certain i can get a reasonable upper-limit and scale it in relation to the significant parameters (number of alignment heads being the main factor). 

This will solve the memory and graph issue, but comes with two caveats:

- We will probably allocate more memory than needed (hopefully not much)
- If future changes occur on `whisper_exp_compute_token_level_timestamps_dtw` that cause new tensors/operations to be allocated, the size of the pre-allocation has to be manually adjusted again.

So, although not ideal, i haven't been able to figure out another path for these issues. @bobqianic do you find this reasonable?

whisper.cpp
```diff
@@ -2539,6 +2574,13 @@ static struct ggml_cgraph * whisper_build_graph_decoder(
 
     struct ggml_tensor * logits = ggml_mul_mat(ctx0, model.d_te, cur);
 
+    // FIXME: maybe this additional node should by part of pipeline if requested,
+    // since it is only needed for dtw timestamps and useless on all other decoding
+    // calls. Should be counted when allocating memory for the decoder graph though.
+    // [EXPERIMENTAL] Token-level timestamps with DTW
+    for (int il = 0; il < n_layer; ++il)
+        ggml_build_forward_expand(gf, wstate.cross_QKs[il]);
```
bobqianic, line 2582: We can do this https://github.com/ggerganov/ggml/issues/700#issuecomment-1897621084

> If you want to keep some of the intermediate results, the recommended approach would be pre-allocate some tensors in a different buffer and use ggml_cpy to copy the result there.

slaren: Not sure if this changes anything for this case, but I plan to add (relatively soon) a way to specify graph outputs by setting a flag in `ggml_tensor`. Setting the "output" flag would prevent it from being overwritten by future operations when using ggml-alloc.

denersc:
I'll try to implement the changes discussed above by next week.

On other topic, had a run-in with a bug today. During `whisper_exp_compute_token_level_timestamps_dtw`, the call to `whisper_decode_with_state` failed. 

Inside `whisper_decode_internal`, the call to find a KV cache slot failed:

```c
// inside whisper_decode_internal
if (!whisper_kv_cache_find_slot(kv_self, batch)) {
    return false; // returns here
}

// inside whisper_kv_cache_find_slot
if (n_tested >= n_ctx) {
    //WHISPER_LOG_ERROR("%s: failed to find a slot for %d tokens\n", __func__, n_tokens);
    return false; // returns here
}

```

This failure seems to be fixed if a place a call to `whisper_kv_cache_clear` before my call to `whisper_decode_with_state` on DTW timestamping function.

```c
// Get result tokens, pass then along to decoder to get cross attention QKs
// used in timestamping
// Each QK is audio_ctx*N_TOKENS*N_HEADS_PER_LAYER
whisper_kv_cache_clear(state->kv_self);
if (whisper_decode_with_state(ctx, state, tokens.data(), tokens.size(), 0, n_threads) != 0) {
    WHISPER_LOG_INFO("DECODER FAILED\n");
    WHISPER_ASSERT(0);
}
```

I have a very poor if not nonexistent understanding of the KV cache, so i 'm unsure if this has any unintended consequences. The only analysis i could readily make was:

- `whisper_exp_compute_token_level_timestamps_dtw` is called by the end of an inference iteration.
- `whisper_kv_cache_clear(state->kv_self)` is also called in the inference loop right before starting decoding on a next iteration, so the call on `timestamps_dtw` probably have no effect on next iterations.

Any input in this would be appreciated!

erturkdotgg:
Any updates about this?

denersc:
> Any updates about this?

I've been very swamped the last couple of weeks and couldn't work on this, but i have the intent to see this through the end. I'll update when more progress comes up.

denersc:
So, i recently synced my fork cause it had been a while. Recent-ish (at least after 22/01) whisper.cpp commits made my additions to `whisper_build_graph_decoder` use a lot more memory than before:

```c++
// Inside whisper_build_graph_decoder
// For each layer...
            struct ggml_tensor * KQ_copy = ggml_cpy(ctx0, KQ_soft_max, ggml_new_tensor_3d(ctx0, GGML_TYPE_F32, KQ_soft_max->ne[0], KQ_soft_max->ne[1], KQ_soft_max->ne[2]));
            wstate.cross_QKs.push_back(KQ_copy);

// Later by end of function            
    for (int il = 0; il < n_layer; ++il)
        ggml_build_forward_expand(gf, wstate.cross_QKs[il]);
```

Saving all cross QKs already consumed a lot of memory (about 1.6gb on large-v2). Synced with latest whisper.cpp, saving cross QKs is consuming practically double that (about 3077gb on large-v2). 

```bash
# All large-v2

# whisper.cpp - recent - 31891db2e3cc5ef1887235d5af18c9e728f3eb12
whisper_init_state: compute buffer (decode) =  209.26 MB

# whisper.cpp  - recent - 31891db2e3cc5ef1887235d5af18c9e728f3eb12 + my changes 
whisper_init_state: compute buffer (decode) = 3434.86 MB

# whisper.cpp  - from 22/01 - 1cf679dec4eca99aeaed4fe09a8092803bdecfc1 + my changes
whisper_init_state: compute buffer (decode) = 1878.02 MB
```

Although unfortunate, it points to the fact that the saving all the cross QKs is a huge waste of memory, since generally we want something that i think is about less than 10% of all data in those QKs.

Given this, i think it would be very much worth to, on calls to the decoder, already clip only the attention heads we want instead of doing it later. Saving cross QKs should also be optional (used only when calling for DTW timestamping) to avoid any overhead when normal inference is taking place.

That said, I'm unsure how to proceed. Ideally, when whisper_decode (and consequentely whisper_buld_graph_decoder) is called to obtain cross QKs:

- For each text layer, check if it contains one or more attention heads that were specified during init.
- If it does, copy only cross QKs from those heads (heads are indexed by the third dimension of tensor), instead of the whole tensor. If not, don't copy anything.

Does anyone with more knowledge of the ggml API have a suggestion on how to implement this?

ggerganov:
With latest `ggml` you should be able to now mark tensors as "outputs":

```c
ggml_set_output(KQ_soft_max);
wstate.cross_QKs.push_back(KQ_soft_max);
```

Output tensors will not be overwritten by following results. This way you can avoid the copies

denersc:
Some changes from last commits:

- DTW on/off switch and presets are now set on `whisper_context_params`, not on `whisper_params`
- Decoder only saves cross QKs if requested (changed `whisper_decode_internal` signature to have a bool that enables/disables saving cross QKs)
- Significant decrease of memory usage when using DTW timestamps. 
    - On older commits, we saved all cross QKs when decoding and later on `whisper_exp_compute_token_level_timestamps_dtw` we merged the ones we actually wanted (i.e. the ones from alignment heads). 
    - Now, we already select the desired ones inside the decoder and avoid saving all of them. This is done by multiplying the QKs to some masks initialized during whisper_init_state, which will correctly drop the unnecessary heads. Remaining heads are concatenated and saved in `whisper_state` to be used when timestamping.
    - This saves a huge amount of memory. E.g. in large-v2, saving the cross QKs set the decoder compute_buffer to 3434.86 MB. After changes, this is now 742.28 MB
- Median filter is now called with `ggml_map_custom1` and is computed when computing the graph.

Because of changes to cross QK saving, I need to reimplement DTW_AHEADS_CUSTOM and DTW_AHEADS_N_TOP_MOST, but that should be relatively easy. Other presets are working.

What is currently missing is not having dynamic allocations inside `whisper_exp_compute_token_level_timestamps_dtw`. Currently, inside this function, we allocate 6 tensors with `ggml_new_tensor_...` (2 + 4 inside `dtw_and_backtrace`) + our compute graph. Allocation is hard-coded to max 128mb.

@slaren @bobqianic would you be able to advice me how to deal with this last detail? Would it would be easier to pre-compute needed memory if all operations occurred inside a graph? If so, i think two manual operations could be replaced by `ggml_view_...`, although i haven't been able to figure out strides without segfaulting. 

```c++
7028     // IN: Tensor with N_TOKENS*audio_ctx*N_ALIGNMENT_HEADS dims
7029     // OUT:Tensor with N_TOKENS*N_AUDIO_TOKENS*N_ALIGNMENT_HEADS dims
7030     // FIXME: can probably be replaced with a view. I failed to get right strides
7031     ggml_tensor * w = ggml_new_tensor_3d(gctx, GGML_TYPE_F32, n_tokens, n_audio_tokens, n_heads);
7032     for (int i = 0; i < n_tokens; ++i) {
7033         for (int j = 0; j < n_audio_tokens; ++j) {
7034             for (int k = 0; k < n_heads; ++k) {
7035                 const float v = ggml_get_f32_nd(state->aheads_cross_QKs, i, j, k, 0);
7036                 ggml_set_f32_nd(w, i, j, k, 0, v);
7037             }
7038         }
7039     }
```

and

```c++
7071     // Remove SOT sequence and EOT
7072     // Out dimension is (N_TOKENS-sot_sequence_length-1)*N_AUDIO_TOKENS
7073     // FIXME: can probably be replaced with a view. I failed to get right strides
7074     ggml_tensor * matrix = ggml_new_tensor_2d(gctx, GGML_TYPE_F32, w->ne[0] - sot_sequence_length - 1, w->ne[1]);
7075     for (int64_t i = 0; i < matrix->ne[0]; ++i) {
7076         for (int64_t j = 0; j < matrix->ne[1]; ++j) {
7077             float v = ggml_get_f32_nd(w, i + sot_sequence_length, j, 0, 0);
7078             ggml_set_f32_nd(matrix, i, j, 0, 0, v);
7079         }
7080     }
```

The last 4 manual operations (inside `dtw_and_backtrace`) could probably be done over  a reserved std::vector. This way, all computation involving `ggml_tensor` would be within a graph.

bobqianic:
> What is currently missing is not having dynamic allocations inside whisper_exp_compute_token_level_timestamps_dtw. Currently, inside this function, we allocate 6 tensors with ggml_new_tensor_... (2 + 4 inside dtw_and_backtrace) + our compute graph. Allocation is hard-coded to max 128mb.

I'm not sure how to accomplish this. Let's discuss it with @slaren

slaren:
I don't understand what issue you are having allocating the tensors, can you elaborate? If you add any tensors to the graph with `ggml_build_forward_expand`, they will be automaticaly allocated when the graph is allocated.

Regarding the code that you copied, the functions `ggml_get_f32_nd` and `ggml_set_f32_nd` are not compatible with ggml-backend, since these functions only work with tensors allocated in the CPU. Currently, the only way to access and modify the data of a tensor when using ggml-backend is with the functions `ggml_backend_tensor_get` and `ggml_backend_tensor_set`.

denersc:
Hey @slaren sure.

> I don't understand what issue you are having allocating the tensors, can you elaborate? If you add any tensors to the graph with ggml_build_forward_expand, they will be automaticaly allocated when the graph is allocated.

So, currently, inside `whisper_exp_compute_token_level_timestamps_dtw` - which is called on every inference iteration - i create a `ggml_context` with `no_alloc=false`. I understand this makes it so every time i call that function, the 128MB i chose on params will be re-allocated, correct?

```c++
// inside whisper.cpp
7019     struct ggml_init_params gparams = {
7020         /*.mem_size   =*/ 128*1024*1024,
7021         /*.mem_buffer =*/ NULL,
7022         /*.no_alloc   =*/ false,
7023     };
7024     struct ggml_context * gctx = ggml_init(gparams);

...

7164     ggml_free(gctx);
```

There is no guarantee 128MB will be enough, since it depends on how many alignment heads a user chose to use and how many tokens were produced during inference.

So, what i wish i could do, in spirit with the rest of whisper.cpp, is avoid this constant reallocation and properly compute how much memory i need beforehand. To do this, i would need to compute the memory i need during `whisper_init_state`, pre-allocate the backend buffer, and during my function start my `ggml_context` with `no_alloc=true`. 

From what i have gathered from other times this is done (e.g. building encoder, decoder graphs), the best way to do this would be to build the graph and ask ggml to tell me how much memory it needs.

Problem is, there are two operations inside  `whisper_exp_compute_token_level_timestamps_dtw`  that i could not figure out how to perform using ggml functions (the two code pieces i posted above). In the first one, I am basically removing the last X rows from a tensor, and in the second one, removing the first X columns from a tensor.

So, circling back:

- If i were able to perform those using ggml funcs and so all my computation would be inside a graph, i would be able to properly get how much memory i need to preallocate, correct?
- If so, do you have any recommendations on that ggml functions i could use to perform those operations? I though i could have done it `ggml_view_...`, but always got a segfault. I imagine i might be using wrong strides/offsets.

> Regarding the code that you copied, the functions ggml_get_f32_nd and ggml_set_f32_nd are not compatible with ggml-backend, since these functions only work with tensors allocated in the CPU. Currently, the only way to access and modify the data of a tensor when using ggml-backend is with the functions ggml_backend_tensor_get and ggml_backend_tensor_set.

I was also unaware of that. I might need to adjust to that or just use CPU tensors. My `median_filter` operations also uses `ggml_get/set_f32_nd`

slaren:
If you want to allocate a list of static tensors, the best way to do that is with `ggml_backend_alloc_ctx_tensors`. To do this, create the tensors in a `no_alloc` context, with just enough space for the number of tensors that you need to allocate, and then call `ggml_backend_alloc_ctx_tensors` to allocate them in a backend buffer. You can calculate the size of the context with `ggml_tensor_overhead(). For example, look at the code that allocates the KV tensors:
```C++
    struct ggml_init_params params = {
        /*.mem_size   =*/ 2*ggml_tensor_overhead(),
        /*.mem_buffer =*/ nullptr,
        /*.no_alloc   =*/ true,
    };

    cache.ctx = ggml_init(params);

    if (!cache.ctx) {
        WHISPER_LOG_ERROR("%s: failed to allocate memory for the kv cache context\n", __func__);
        return false;
    }

    cache.k = ggml_new_tensor_1d(cache.ctx, wtype, n_elements);
    cache.v = ggml_new_tensor_1d(cache.ctx, wtype, n_elements);

    cache.buffer = ggml_backend_alloc_ctx_tensors(cache.ctx, backend);
```

Regarding the operations in the code, I think these views would be equivalent to the operation that you are performing there:

```C
ggml_tensor * w = ggml_view_3d(ctx, state->aheads_cross_QKs, n_tokens, n_audio_tokens, n_heads,
        state->aheads_cross_QKs->nb[1], state->aheads_cross_QKs->nb[2],
        0);

ggml_tensor * matrix = ggml_view_2d(ctx, w, w->ne[0] - sot_sequence_length - 1, w->ne[1],  
        w->nb[1],
        sot_sequence_length * w->nb[0]);
```

denersc:
Thanks for all the help!

> Regarding the code that you copied, the functions ggml_get_f32_nd and ggml_set_f32_nd are not compatible with ggml-backend, since these functions only work with tensors allocated in the CPU. Currently, the only way to access and modify the data of a tensor when using ggml-backend is with the functions ggml_backend_tensor_get and ggml_backend_tensor_set.

I guess i missed that fact since I'm running with Metal, and in my case GPU and CPU memory were unified, so i was able to retrieve values using those functions on the tensors i was saving from the decoder (`tensor->backend` was CPU even with Metal backend). My code probably wouldn't have worked if someone tried to use it with other GPU acceleration. 

I've reworked a portion of the code to use `ggml_backend_tensor_get` instead and now i believe it should work for other GPUs.

I tried reworking the code to pre-allocate the memory needed for DTW, but i felt like it would take me more effort than i am able to give right now.

In the current state, i believe the code in this PR is usable, but not ideal and comes with many caveats. If anybody would like to give a go on finishing this, be my guest. I'll be open to chat and discuss what i figured so far. I'm unsure when i will be able to put some effort again, although I'll probably keep the fork sync with whisper.cpp. 

This is the current state of affairs:

- Everytime `whisper_exp_compute_token_level_timestamps_dtw`, we allocate (and free by the end) a `ggml_context` with 128MB by default. This memory can be inc/dec on startup changing context params, changing the field `dtw_mem_size`.
- During my tests, 128MB have were sufficient when using `large-v2`, which is the preset that should need more memory. Nevertheless, I'm not sure if it's enough on the worst possible scenario.
- All processing done over `whisper_exp_compute_token_level_timestamps_dtw` occurs on CPU. Some things, like the median filter and DTW/backtrace algorithms would need to be adjusted to work on other backends, since i rely on `ggml_get/set_f32_nd`. Another option would be to do all other operations on other backends and do filter and DTW work on CPU, not necessarily inside a tensor. DTW is not particularly compute-heavy compared to other parts of the pipeline.

I think those are the more critical points. Less critically, i believe that the way i save cross QKs might not be very optimal and maybe there are other more memory-efficient ways to do it. 

If anyone would like to try running this fork, just change your `whisper_context_params` accordingly:

```c++
struct whisper_context_params cparams = whisper_context_default_params();
cparams.dtw_token_timestamps = true;
cparams.dtw_aheads_preset = WHISPER_AHEADS_LARGE_V2;     // Choose a preset that matches your model
//cparams.dtw_mem_size = 128*1024*1024                   // Defaults to 128MB. Increase if needed
struct whisper_context * ctx = whisper_init_from_file_with_params(params.model.c_str(), cparams);
```

Timestamps should be available after inference on `whisper_token_data.t_dtw`

federico-terzi:
Hey @denersc, thank you for the amazing work! I've tested your fork on Windows and can confirm it works correctly there as well (without an external GPU).

For what is worth, I'd love to see this merged upstream, as the potential applications are endless (I'm not saying this to pressure you in any way, just to remark how valuable your work so far has been! so thank you for that)

whisper.h
```diff
@@ -84,9 +84,48 @@ extern "C" {
...
+
+    typedef struct whisper_ahead {
+        int n_text_layer;
+        int n_head;
+    } whisper_ahead;
+
+    typedef struct whisper_aheads {
+        size_t n_heads;
+        const whisper_ahead * heads;
+    } whisper_aheads;
```
ggerganov, line 112: These look like an implementation detail - probably better to hide them in `whisper_context`

ggerganov (commented):
Thank you for the great effort! The implementation is quite good and well separated, so I think we can proceed with merging

I'll try to dedicate some time in the following days to play with this. Even if the implementation is not perfect and there are things to polish out, we can do this from `master`

In the meantime, if you want to wrap or cleanup anything before the merge you can do so. Just mark this PR as ready for review when you are ready. Also, if people give this a try and report results and write usage instructions - it would be great!

denersc:
Awesome!

I'll wrap it up, clean some code and mark for review soon.

ggerganov:
The DTW timestamps can now be generated with `main` using the `-ojf` CLI argument:

```bash
./main -m models/ggml-small.bin -f samples/gb0.wav -dtw small -ojf
```

```json
{
			"timestamps": {
				"from": "00:02:02,960",
				"to": "00:02:05,520"
			},
			"offsets": {
				"from": 122960,
				"to": 125520
			},
			"text": " by the will of the people.",
			"tokens": [
				{
					"text": " by",
					"timestamps": {
						"from": "00:02:02,960",
						"to": "00:02:03,180"
					},
					"offsets": {
						"from": 122960,
						"to": 123180
					},
					"id": 538,
					"p": 0.999897,
					"t_dtw": 12312
				},
				{
					"text": " the",
					"timestamps": {
						"from": "00:02:03,180",
						"to": "00:02:03,510"
					},
					"offsets": {
						"from": 123180,
						"to": 123510
					},
					"id": 264,
					"p": 0.999729,
					"t_dtw": 12328
				},
				{
					"text": " will",
					"timestamps": {
						"from": "00:02:03,510",
						"to": "00:02:03,590"
					},
					"offsets": {
						"from": 123510,
						"to": 123590
					},
					"id": 486,
					"p": 0.997792,
					"t_dtw": 12354
				},
				{
					"text": " of",
					"timestamps": {
						"from": "00:02:04,140",
						"to": "00:02:04,170"
					},
					"offsets": {
						"from": 124140,
						"to": 124170
					},
					"id": 295,
					"p": 0.999649,
					"t_dtw": 12430
				},
				{
					"text": " the",
					"timestamps": {
						"from": "00:02:04,170",
						"to": "00:02:04,500"
					},
					"offsets": {
						"from": 124170,
						"to": 124500
					},
					"id": 264,
					"p": 0.999611,
					"t_dtw": 12440
				},
				{
					"text": " people",
					"timestamps": {
						"from": "00:02:04,500",
						"to": "00:02:05,090"
					},
					"offsets": {
						"from": 124500,
						"to": 125090
					},
					"id": 561,
					"p": 0.999641,
					"t_dtw": 12482
				},
				{
					"text": ".",
					"timestamps": {
						"from": "00:02:05,200",
						"to": "00:02:05,440"
					},
					"offsets": {
						"from": 125200,
						"to": 125440
					},
					"id": 13,
					"p": 0.998121,
					"t_dtw": 12512
				},
				{
					"text": "[_TT_416]",
					"timestamps": {
						"from": "00:02:05,520",
						"to": "00:02:05,520"
					},
					"offsets": {
						"from": 125520,
						"to": 125520
					},
					"id": 50780,
					"p": 0.280601,
					"t_dtw": -1
				}
			]
		},
```

ggerganov (approved):

ggerganov (approved):

hlevring:
I am looking to derive word level timestamps from DTW timestamps. Some tokens need to be joined to form a complete word, so as a start I figured to just concatenate tokens until a space or punctuation mark is encountered, and from just accumulate the timestamps from the tokens to determine the start and end times for each word. 

I am pretty sure there must be way more to consider.  @denersc it would be great if you can comment on this. - and hey super thanks for your work on the DTW timestamps.

eschmidbauer:
not sure if anyone has noticed this too but the DTW timestamps seem to be completely inaccurate in some segments of the transcript. while other segments, it's very precise.

hlevring:
I was experimenting yesterday and saw forward shifts in token timestamps prior to silence periods. (Say pauses in speech for 1-2 sec).  In my tests I did have a good bit of background noise that might have been a factor. I will try to redo some experiments in the next few days.

denersc:
Hey @hlevring and @eschmidbauer, thanks for trying it out!

So, I'll try to address what you guys said, but unfortunately i don't think i can provide perfect answers.

First, when thinking about DTW timestamps, i crudely rationalize about them as "A estimate of the moment the model decided to output a certain token".

So, in common speech flow, like between equal paced words in a sentence, it is very likely that the DTW timestamp of the last token in a word will be very close to the actual time of the end of the word. 

Nevertheless, it is not completely unusual that the model will the output some token long after it actually occurred in audio. In that case, DTW timestamp will likely be incorrect. The most common example of this is the period (.) token. It can outputted by the model after some time of silence preceded by a sequence of words. So the DTW timestamp for the period will be long after sentence end, and should probably be ignored.

Although period and punctuation in general are the most common occurrence of this, I don't doubt this kind of thing can happen with words, e.g. model outputting 3 words almost simultaneously because only when it understood the third word could it actually infer all three. In that case, the first 2 words will have very imprecise timestamps. Although i think this may be possible, it does not seem to be very likely, at least in the sense that i haven't observed it directly.

All of this to say, DTW timestamps are a imperfect source of information, and should be used  with some caution and combined with other data to provide good word timestamp estimates. [OpenAI tries to address some of theses issues](https://github.com/openai/whisper/blob/ba3f3cd54b0e5b8ce1ab3de13e32122d0d5f98ab/whisper/timing.py#L298) 

Of course, it may also be the case that my implementation is incorrect on some point. Maybe on the step of selecting and saving alignment heads or when performing some matrix operations. I think a good starting point to check that would be to compare the DTW timestamps given by openAI impl with the ones in my implementation for a variety of audios and see if there are any large discrepancies. Some very small variance is bound to happen probably because of different matrix operation implementations.

Finally, I'm more on the developer side than on the ML research side, my math understanding is reasonably shallow. So take all i said with a grain of salt :grimacing:

denersc:
Also, forgot to say, make sure whisper.cpp version you guys are using is after MR #2012. That bug likely caused incorrect alignment head selection, beyond the observed memory error.

Cool @eschmidbauer. On whisper.cpp, you can [uncomment these lines](https://github.com/ggerganov/whisper.cpp/blob/58210d6a7634ea1e42e0a2dab611f4a0518731dc/whisper.cpp#L7184) if you want to print DTW timestamps.

You might need to change the code on the OpenAI package to get the actual raw DTW timestamps, since they don't provide them to the final user. They do a lot of additional processing before giving it back, so those will be different for sure. You'll probably have to add some code to [retrieve these](https://github.com/openai/whisper/blob/ba3f3cd54b0e5b8ce1ab3de13e32122d0d5f98ab/whisper/timing.py#L214) and do some sort of loop equivalent to what i did [here](https://github.com/ggerganov/whisper.cpp/blob/58210d6a7634ea1e42e0a2dab611f4a0518731dc/whisper.cpp#L7158) to get timestamps for each token. These will be the raw timestamps which are comparable to the ones i made available on whisper.cpp.

Issue #1498, opened 2023-11-16
CUDA: the provided PTX was compiled with an unsupported toolchain


Linux13524:
Hello. Currently Im trying to build Whisper.cpp with CUDA support, however I get an error when trying to init Whisper.cpp. I run the test on a different device with different CUDA minor version than I compile on.

### How to reproduce

Test code:

```cpp
TEST_CASE("whisper-cpp: can init", "[whisper-cpp]") {
    whisper_context *context = whisper_init_from_file_with_params("res/models/for-tests-ggml-base.en.bin", whisper_context_default_params());
    REQUIRE(context != nullptr);
    whisper_free(context);
}
```

Building with `WHISPER_CUBLAS=ON`.

Output:

```
whisper_init_from_file_with_params_no_state: loading model from 'res/models/for-tests-ggml-base.en.bin'
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51864
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 512
whisper_model_load: n_audio_head  = 8
whisper_model_load: n_audio_layer = 6
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 512
whisper_model_load: n_text_head   = 8
whisper_model_load: n_text_layer  = 6
whisper_model_load: n_mels        = 80
whisper_model_load: ftype         = 1
whisper_model_load: qntvr         = 0
whisper_model_load: type          = 2 (base)
whisper_model_load: adding 1607 extra tokens
whisper_model_load: n_langs       = 99
whisper_model_load: model ctx     =  140.66 MB
ggml_init_cublas: GGML_CUDA_FORCE_MMQ:   no
ggml_init_cublas: CUDA_USE_TENSOR_CORES: yes
ggml_init_cublas: found 8 CUDA devices:
  Device 0: NVIDIA RTX A5000, compute capability 8.6
  Device 1: NVIDIA RTX A5000, compute capability 8.6
  Device 2: NVIDIA RTX A4000, compute capability 8.6
  Device 3: NVIDIA RTX A4000, compute capability 8.6
  Device 4: NVIDIA RTX A4000, compute capability 8.6
  Device 5: NVIDIA RTX A4000, compute capability 8.6
  Device 6: NVIDIA RTX A4000, compute capability 8.6
  Device 7: NVIDIA RTX A4000, compute capability 8.6

[...]

PR #1499, opened 2023-11-16
Fix CUDA architecture configuration

Linux13524:
This will fix the problem reported in #1498.


bobqianic:
Try `-DGGML_CUDA_ARCHITECTURES=all`

PR #1505, opened 2023-11-17
add recorder and simplest

litongjava:
an example of record audio to file
an example of recognize audio file


litongjava:
I add to examples. Please review code

PR #1519, opened 2023-11-18
Fixed with_state methods, to use the correct state

sandrohanea:
I wanted to upgrade https://github.com/sandrohanea/whisper.net to latest whisper.cpp runtime (1.5.0) but discovered that some functions are not working anymore using external provided state.

This is because the context-based state was used in these methods.

If the methods are used from a state call (for parallel transformation using multiple states and same context), errors will arise.

Fix is relatively, simple. 

Tested the new build and working as expected with state and searched for all `ctx->state` usages in the whisper.cpp and all of them seems correct now.


ggerganov (approved):
Thank you!

PR #1521, opened 2023-11-19
whisper : reuse whisper_decode_with_state

ggerganov:


ggerganov:
@sandrohanea PTAL

sandrohanea:
I missed that, but yes, I think it makes sense to reuse the _with_state function.

PR #1535, opened 2023-11-21
Change temp file name for server application

felrock:
Avoid issue of removing file if it exists in the current working directory.

### DIFF
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -453,21 +453,23 @@ int main(int argc, char ** argv) {
         std::vector<float> pcmf32;               // mono-channel F32 PCM
         std::vector<std::vector<float>> pcmf32s; // stereo-channel F32 PCM
 
-        // write file to temporary file
-        std::ofstream temp_file{filename, std::ios::binary};
+        // write to temporary file
+        const std::string temp_filename = "whisper_server_temp_file.wav";
+        std::ofstream temp_file{temp_filename, std::ios::binary};
         temp_file << audio_file.content;
         temp_file.close();
 
         // read wav content into pcmf32
-        if (!::read_wav(filename, pcmf32, pcmf32s, params.diarize)) {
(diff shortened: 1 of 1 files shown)


felrock:
Sorry for adding these changes one by one!! Not very good practice of me, will try to gather them next time.

examples/server/server.cpp
```diff
@@ -453,21 +453,22 @@ int main(int argc, char ** argv) {
...
         // read wav content into pcmf32
-        if (!::read_wav(filename, pcmf32, pcmf32s, params.diarize)) {
-            fprintf(stderr, "error: failed to read WAV file '%s'\n", filename.c_str());
+        if (!::read_wav(temp_filename, pcmf32, pcmf32s, params.diarize)) {
+            fprintf(stderr, "error: failed to read WAV file '%s'\n", temp_filename.c_str());
             const std::string error_resp = "{\"error\":\"failed to read WAV file\"}";
             res.set_content(error_resp, "application/json");
             whisper_mutex.unlock();
             return;
         }
```
bobqianic, line 469: What if the file saved on the disk is corrupt, causing errors when reading the WAV file? In this scenario, the corrupt file isn't deleted, which could lead to further issues.

felrock: Yeah, we should remove it also if the read fails, added

bobqianic:
I believe we can just tweak the `read_wav()` function in `common.cpp` to enable it to directly read WAV files from RAM, bypassing the need to write them to disk.

https://github.com/ggerganov/whisper.cpp/blob/146169ec380ca6976dce2ea7c5fb31ac27b3095f/examples/dr_wav.h#L875-L878

felrock:
Yes, I tried it for a while with no success dr_wav is not a very versatile lib imo. For the server to scale that is a really nice improvement, but for just running it locally the time to read/save file is marginal to the actual time of the inference.

bobqianic (approved):

PR #1539, opened 2023-11-22
automatically convert audio on the server

sapoepsilon:
- added argument `--convert`
- send execute ffmpeg conversion command if the `--convert ` argument is present
whisper-converter model

### DIFF
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -448,7 +448,7 @@ int main(int argc, char ** argv) {
     }
 
     if (sparams.ffmpeg_converter) {
-        check_FFmpeg_availibility();
+        check_ffmpeg_availibility();
     }
     // whisper init
     struct whisper_context_params cparams;
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -119,7 +119,7 @@ bool is_file_exist(const char *fileName)
 }
 
 void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & params,
-                        const server_params& sparams) {
+                         const server_params& sparams) {
     fprintf(stderr, "\n");
     fprintf(stderr, "usage: %s [options] \n", argv[0]);
     fprintf(stderr, "\n");
@@ -506,6 +506,7 @@ int main(int argc, char ** argv) {
         temp_file.close();
 
         // if file is not wav, convert to wav
+        
         if(sparams.ffmpeg_converter) {
             std::string error_resp = "{\"error\":\"Failed to execute ffmpeg command.\"}";
             bool isConverted = convert_to_WAV(temp_filename, error_resp);
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -509,7 +509,7 @@ int main(int argc, char ** argv) {
         
         if (sparams.ffmpeg_converter) {
             std::string error_resp = "{\"error\":\"Failed to execute ffmpeg command.\"}";
-            const bool is_converted = convert_to_WAV(temp_filename, error_resp);
+            const bool is_converted = convert_to_wav(temp_filename, error_resp);
             if (!is_converted) {
                 res.set_content(error_resp, "application/json");
                 whisper_mutex.unlock();
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -505,6 +505,7 @@ int main(int argc, char ** argv) {
             fprintf(stderr, "error: failed to read WAV file '%s'\n", temp_filename.c_str());
             const std::string error_resp = "{\"error\":\"failed to read WAV file\"}";
             res.set_content(error_resp, "application/json");
+            std::remove(temp_filename.c_str());
             whisper_mutex.unlock();
             return;
         }
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -43,6 +43,7 @@ struct server_params
     int32_t port          = 8080;
     int32_t read_timeout  = 600;
     int32_t write_timeout = 600;
+    
     bool ffmpeg_converter = false;
 };
 
@@ -156,8 +157,7 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
     fprintf(stderr, "  --host HOST,                   [%-7s] Hostname/ip-adress for the server\n", sparams.hostname.c_str());
     fprintf(stderr, "  --port PORT,                   [%-7d] Port number for the server\n", sparams.port);
     fprintf(stderr, "  --public PATH,                 [%-7s] Path to the public folder\n", sparams.public_path.c_str());
-    fprintf(stderr, "  --convert,                     [%-7s] Convert audio to WAV, requires ffmpeg on the server",
-            sparams.ffmpeg_converter ? "true" : "false");
+    fprintf(stderr, "  --convert,                     [%-7s] Convert audio to WAV, requires ffmpeg on the server", sparams.ffmpeg_converter ? "true" : "false");
     fprintf(stderr, "\n");
 }
 
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -221,7 +221,7 @@ struct whisper_print_user_data {
     int progress_prev;
 };
 
-bool convert_to_WAV( std::string& temp_filename, std::string &error_resp) {
+bool convert_to_WAV(const std::string& temp_filename, std::string &error_resp) {
     std::ostringstream cmd_stream;
     std::string converted_filename_temp = temp_filename + "_temp.wav";
     cmd_stream << "ffmpeg -i \"" << temp_filename << "\" -ar 16000 -ac 1 -c:a pcm_s16le \"" << converted_filename_temp << "\" 2>&1";
@@ -483,8 +483,8 @@ int main(int argc, char ** argv) {
         std::vector<float> pcmf32;               // mono-channel F32 PCM
         std::vector<std::vector<float>> pcmf32s; // stereo-channel F32 PCM
 
-        // write file to temporary file
-        std::string temp_filename = "whisper_server_temp_file.wav";
+        // write to temporary file
+        const std::string temp_filename = "whisper_server_temp_file.wav";
         std::ofstream temp_file{temp_filename, std::ios::binary};
         temp_file << audio_file.content;
         temp_file.close();
@@ -493,7 +493,6 @@ int main(int argc, char ** argv) {
         if(sparams.ffmpeg_converter) {
             std::string error_resp = "{\"error\":\"Failed to execute ffmpeg command.\"}";
             bool isConverted = convert_to_WAV(temp_filename, error_resp);
-            // if isConverted is false, return error response and unlock mutex and print error
             if(!isConverted) {
                 res.set_content(error_resp, "application/json");
                 whisper_mutex.unlock();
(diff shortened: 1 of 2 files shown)


examples/server/server.cpp
```diff
@@ -453,23 +487,33 @@ int main(int argc, char ** argv) {
         std::vector<float> pcmf32;               // mono-channel F32 PCM
         std::vector<std::vector<float>> pcmf32s; // stereo-channel F32 PCM
 
-        // write to temporary file
-        const std::string temp_filename = "whisper_server_temp_file.wav";
-        std::ofstream temp_file{temp_filename, std::ios::binary};
+        // write file to temporary file
+        std::ofstream temp_file{filename, std::ios::binary};
```
felrock, line 491: We can't create a file with the same name as the file we recieve, if the request/file is the in the same current working directory it will remove the file completely. Please keep the filename.

examples/server/server.cpp
```diff
@@ -453,23 +487,33 @@ int main(int argc, char ** argv) {
...
+            filename = converted_filename;
+        }
+
         // read wav content into pcmf32
-        if (!::read_wav(temp_filename, pcmf32, pcmf32s, params.diarize)) {
-            fprintf(stderr, "error: failed to read WAV file '%s'\n", temp_filename.c_str());
+        if (!::read_wav(filename, pcmf32, pcmf32s, params.diarize)) {
+            fprintf(stderr, "error: failed to read WAV file '%s'\n", filename.c_str());
             const std::string error_resp = "{\"error\":\"failed to read WAV file\"}";
             res.set_content(error_resp, "application/json");
```
felrock, line 511: Keep file remove here.

examples/server/server.cpp
```diff
@@ -217,6 +221,36 @@ struct whisper_print_user_data {
     int progress_prev;
 };
 
+std::string executeConvertToWAV(const std::string& cmd) {
```
felrock, line 224: function name should be camel_case

examples/server/server.cpp
```diff
@@ -453,23 +487,33 @@ int main(int argc, char ** argv) {
...
+        // write file to temporary file
+        std::ofstream temp_file{filename, std::ios::binary};
         temp_file << audio_file.content;
         temp_file.close();
 
+        // if file is not wav, convert to wav
+        if(sparams.ffmpeg_converter) {
+            std::string converted_filename = "converted_" + filename;
+            // ffmpeg command
+            std::string ffmpeg_cmd = "ffmpeg -i " + filename + " -ar 16000 -ac 1 -c:a pcm_s16le " + converted_filename + " 2>&1"; // Redirect stderr to stdout
```
felrock, line 499: Would be cleaner to put this cmd in the executeConvertToWAV function, since it's not used anywhere else here. Just pass the filename to the function.

examples/server/server.cpp
```diff
@@ -453,23 +487,33 @@ int main(int argc, char ** argv) {
...
+
+            executeConvertToWAV(ffmpeg_cmd);
```
felrock, line 501: since we ignore the string on the return here, perhaps make it into void?

felrock (commented):
Nice addition! Makes sense if someone has mp3 files or similar that wants to send those. I just have a few comments about the outlining stuff. And also it would be great if you updated the README as well with the new parameter.

examples/server/server.cpp
felrock: Also, if the conversion failed we need to let the user know somehow. You could do like i did after "::read_wav" where i send a error_resp via "res.set_content(error_resp)". The error could be something like:

{"error": "File conversion failed."}
or and also 
{"error": "File could not be converted due to no ffmpeg installed."}

examples/server/server.cpp
bobqianic: > We can't create a file with the same name as the file we recieve, if the request/file is the in the same current working directory it will remove the file completely. Please keep the filename.

If I understand correctly, retaining the filename as specified by the user could pose a security risk.

examples/server/server.cpp
```diff
@@ -453,18 +483,29 @@ int main(int argc, char ** argv) {
...
 
+        // if file is not wav, convert to wav
+        if(sparams.ffmpeg_converter) {
+            std::string error_resp = "{\"error\":\"Failed to execute ffmpeg command.\"}";
+            bool isConverted = convert_to_WAV(temp_filename, error_resp);
+            // if isConverted is false, return error response and unlock mutex and print error
+            if(!isConverted) {
+                res.set_content(error_resp, "application/json");
+                whisper_mutex.unlock();
+                return;
```
sapoepsilon, line 500: image

examples/server/server.cpp
```diff
@@ -217,6 +221,32 @@ struct whisper_print_user_data {
...
 
+bool convert_to_WAV( std::string& temp_filename, std::string &error_resp) {
+    std::ostringstream cmd_stream;
+    std::string converted_filename_temp = temp_filename + "_temp.wav";
+    cmd_stream << "ffmpeg -i \"" << temp_filename << "\" -ar 16000 -ac 1 -c:a pcm_s16le \"" << converted_filename_temp << "\" 2>&1";
+    std::string cmd = cmd_stream.str();
+
+    int status = std::system(cmd.c_str());
+    if (status != 0) {
+        error_resp = "{\"error\":\"FFmpeg conversion failed. Check if FFmpeg is installed\"}";
```
sapoepsilon, line 232: I'm not sure if this is what we want.

We can obtain the exact output error output from system(), but that requires additional code. I can implement that if it makes more sense.

However, it will mostly fail if ffmpeg is not installed.

examples/server/server.cpp
felrock: In a way it could, if you would name the file 'server' it would remove the binary and crash the server I think. I haven't tried but interesting take.

examples/server/server.cpp
felrock: Yes, I think that is enough and will cover most cases.

felrock (approved):
Nice, I checkout out the change and played around with it and it seems to work fine.

examples/server/server.cpp
```diff
@@ -217,6 +221,32 @@ struct whisper_print_user_data {
...
 
+bool convert_to_WAV(const std::string& temp_filename, std::string &error_resp) {
+    std::ostringstream cmd_stream;
+    std::string converted_filename_temp = temp_filename + "_temp.wav";
+    cmd_stream << "ffmpeg -i \"" << temp_filename << "\" -ar 16000 -ac 1 -c:a pcm_s16le \"" << converted_filename_temp << "\" 2>&1";
+    std::string cmd = cmd_stream.str();
+
+    int status = std::system(cmd.c_str());
+    if (status != 0) {
+        error_resp = "{\"error\":\"FFmpeg conversion failed. Check if FFmpeg is installed\"}";
```
felrock, line 232: I got an idea, we can put this check before we even start the server. And let the user who starts the server up know that there is no ffmpeg and perhaps exit. Then we don't need to check for it for every request.

ggerganov: Yup!

examples/server/server.cpp
```diff
@@ -155,6 +156,8 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
     fprintf(stderr, "  --host HOST,                   [%-7s] Hostname/ip-adress for the server\n", sparams.hostname.c_str());
     fprintf(stderr, "  --port PORT,                   [%-7d] Port number for the server\n", sparams.port);
     fprintf(stderr, "  --public PATH,                 [%-7s] Path to the public folder\n", sparams.public_path.c_str());
+    fprintf(stderr, "  --convert,                     [%-7s] Convert audio to WAV, requires ffmpeg on the server",
+            sparams.ffmpeg_converter ? "true" : "false");
```
ggerganov, line 160:
```suggestion
    fprintf(stderr, "  --convert,                     [%-7s] Convert audio to WAV, requires ffmpeg on the server", sparams.ffmpeg_converter ? "true" : "false");
```

examples/server/server.cpp
```diff
@@ -43,6 +43,7 @@ struct server_params
     int32_t port          = 8080;
     int32_t read_timeout  = 600;
     int32_t write_timeout = 600;
+    bool ffmpeg_converter = false;
```
ggerganov, line 46:
```suggestion
    int32_t write_timeout = 600;
    
    bool ffmpeg_converter = false;
```

ggerganov (approved):

examples/server/server.cpp
```diff
@@ -404,6 +447,9 @@ int main(int argc, char ** argv) {
         exit(0);
     }
 
+    if (sparams.ffmpeg_converter) {
+        check_FFmpeg_availibility();
```
sapoepsilon, line 451: image

examples/server/server.cpp
```diff
@@ -217,6 +221,45 @@ struct whisper_print_user_data {
...
+
+bool convert_to_WAV(const std::string& temp_filename, std::string &error_resp) {
+    std::ostringstream cmd_stream;
+    std::string converted_filename_temp = temp_filename + "_temp.wav";
+    cmd_stream << "ffmpeg -i \"" << temp_filename << "\" -ar 16000 -ac 1 -c:a pcm_s16le \"" << converted_filename_temp << "\" 2>&1";
+    std::string cmd = cmd_stream.str();
+
+    int status = std::system(cmd.c_str());
+    if (status != 0) {
+        error_resp = "{\"error\":\"FFmpeg conversion failed.\"}";
```
sapoepsilon, line 245: I think it could still be a good idea to have this error response here, since FFmpeg can fail for other reasons.

examples/server/server.cpp
```diff
@@ -459,6 +505,18 @@ int main(int argc, char ** argv) {
         temp_file << audio_file.content;
         temp_file.close();
 
+        // if file is not wav, convert to wav
+        
+        if(sparams.ffmpeg_converter) {
```
ggerganov, line 510:
```suggestion
        if (sparams.ffmpeg_converter) {
```

examples/server/server.cpp
```diff
@@ -459,6 +505,18 @@ int main(int argc, char ** argv) {
...
+            std::string error_resp = "{\"error\":\"Failed to execute ffmpeg command.\"}";
+            bool isConverted = convert_to_WAV(temp_filename, error_resp);
+            if(!isConverted) {
```
ggerganov, line 513:
```suggestion
            const bool is_converted = convert_to_WAV(temp_filename, error_resp);
            if (!is_converted) {
```

examples/server/server.cpp
```diff
@@ -217,6 +221,45 @@ struct whisper_print_user_data {
     int progress_prev;
 };
 
+void check_FFmpeg_availibility() {
```
ggerganov, line 224:
```suggestion
void check_ffmpeg_availibility() {
```

examples/server/server.cpp
```diff
@@ -217,6 +221,45 @@ struct whisper_print_user_data {
...
+        std::cout << "ffmpeg is available." << std::endl;
+    } else {
+        // ffmpeg is not available
+        std::cout << "ffmpeg is not found. Please ensure that ffmpeg is installed ";
+        std::cout << "and that its executable is included in your system's PATH. ";
+        exit(0);
+    }
+}
+
+bool convert_to_WAV(const std::string& temp_filename, std::string &error_resp) {
```
ggerganov, line 237:
```suggestion
bool convert_to_wav(const std::string& temp_filename, std::string &error_resp) {
```

ggerganov (approved):
Some minor style comments

examples/server/server.cpp
vincentfretin: `check_ffmpeg_availability` even better :)

sapoepsilon:
> Some minor style comments

@ggerganov Just implemented them, thank you. Is it ok to merge, then?

bobqianic (commented):
To clarify potential risks, I think it's important to include this type of warning message in the README.

> [!WARNING]  
> **Do not run the server example with administrative privileges and ensure it's operated in a sandbox environment, especially since it involves risky operations like accepting user file uploads and using ffmpeg for format conversions. Always validate and sanitize inputs to guard against potential security threats.**

bobqianic (approved):

PR #1541, opened 2023-11-23
server : add --print-realtime param

ecneladis:
Currently it always prints transcription on the server side.
This commit disabled it by default and enables it with `--print-realtime / -pr`  parameter.

### DIFF
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -72,6 +72,7 @@ struct whisper_params {
     bool no_fallback     = false;
     bool print_special   = false;
     bool print_colors    = false;
+    bool print_realtime  = false;
     bool print_progress  = false;
     bool no_timestamps   = false;
     bool use_gpu         = true;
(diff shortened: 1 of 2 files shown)


ggerganov (approved):
Great! Happy to see contributions to the `server` example - let's make it really useful!

felrock (commented):
Nice addition! Just a minor question about double output

felrock:
I found out why, I think we don't need to set it just use it to set the callbacks

Found this on whisper.cpp:444
`        bool print_realtime;    // print results from within whisper.cpp (avoid it, use callback instead)`

examples/server/server.cpp
```diff
@@ -512,7 +515,7 @@ int main(int argc, char ** argv) {
 
             wparams.strategy = params.beam_size > 1 ? WHISPER_SAMPLING_BEAM_SEARCH : WHISPER_SAMPLING_GREEDY;
 
-            wparams.print_realtime   = false;
+            wparams.print_realtime   = params.print_realtime;
```
felrock, line 518:
```suggestion
            wparams.print_realtime   = false;
```
use `params.print_realtime` inside of `whisper_print_segment_callback` to suppress the print instead.

ecneladis: Good catch, resolved in https://github.com/ggerganov/whisper.cpp/pull/1541/commits/0d4b5da60ec1b8c42e3ce9a4946a81110c9578b0.

ecneladis:
After fix in https://github.com/ggerganov/whisper.cpp/pull/1541/commits/0d4b5da60ec1b8c42e3ce9a4946a81110c9578b0., with `--print-realtime` (`./server -m models/ggml-large.bin --port 8081 -pc -pp --print-realtime`)

Screenshot 2023-11-23 at 20 09 28

Without `--print-realtime`:
Screenshot 2023-11-23 at 20 09 58

ggerganov:
Oops, didn't notice review from @felrock is pending - sorry about that

felrock:
No worries, looks good!

Issue #1387, opened 2023-10-22
GGML_USE_METAL=1 fails on iPadOS running iOS 15/16 (and 17 perhaps)


ronyfadel:
Error is:

```
ggml_metal_init: load pipeline error: Error Domain=AGXMetalA12 Code=3 "Encountered unlowered function call to air.simd_max.f32" UserInfo={NSLocalizedDescription=Encountered unlowered function call to air.simd_max.f32}
```

If I remove calls to `simd_max`, it still fails with:

```
ggml_metal_init: load pipeline error: Error Domain=AGXMetalA12 Code=3 "Encountered unlowered function call to air.simd_sum.f32" UserInfo={NSLocalizedDescription=Encountered unlowered function call to air.simd_sum.f32}
```

etc.. i.e. all simd func calls are failing.

ronyfadel:
@ggerganov what do you think?

This has been mentioned here as well: https://github.com/ggerganov/llama.cpp/pull/2564 (cc @jhen0409)

jhen0409:
Related https://github.com/ggerganov/llama.cpp/discussions/2550

The A-chip models A12Z and before not supported some SIMD operations like simd_max and simd_sum we used in Metal. To fix this, need some works to write new kernels or patch.

I was thinking to do this but not very urgency, and now I think it not worthy to do because it may make affect for future works and neglected maintenance, so I choices to do #1293 for disable Metal in unsupported devices.

ronyfadel:
@bobqianic how is this completed?

Can you please link the related PR or issue?

bobqianic:
> The A-chip models A12Z and before not supported some SIMD operations like simd_max and simd_sum we used in Metal. To fix this, need some works to write new kernels or patch.

#1293 was merged last week. Therefore, even when using `GGML_USE_METAL=1` during compilation, the `-ng` flag can prevent device failures. It's worth noting that the Metal performance on `Apple2` to `Apple6` family devices is relatively low. Running on `ANE` might result in faster transcription speeds. If you prefer to keep this issue open, I'm happy to do that for you.

https://github.com/ggerganov/whisper.cpp/pull/1270#issuecomment-1718690468

sindresorhus:
This limitation is not documented anywhere, so it's impossible to know that it doesn't work on some devices without scouring the issue tracker.

I think Whisper should simply detect the unsupported devices and not activate GPU mode even if instructed to on those devices.

[...]

PR #1547, opened 2023-11-24
metal : add backend function to check device family support

ggerganov:
ref #1387

When Apple7 is not supported, we fallback to CPU automatically

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -1078,6 +1078,11 @@ static ggml_backend_t whisper_backend_init(const whisper_context_params & params
         if (!backend_gpu) {
             WHISPER_LOG_ERROR("%s: ggml_backend_metal_init() failed\n", __func__);
         }
+        if (!ggml_backend_metal_supports_family(backend_gpu, 7)) {
+            WHISPER_LOG_ERROR("%s: Metal GPU does not support family 7 - falling back to CPU\n", __func__);
+            ggml_backend_free(backend_gpu);
+            backend_gpu = NULL;
+        }
     }
 #endif
 
--- a/ggml-metal.h
+++ b/ggml-metal.h
@@ -52,6 +52,11 @@ void ggml_metal_free(struct ggml_metal_context * ctx);
 void * ggml_metal_host_malloc(size_t n);
 void   ggml_metal_host_free  (void * data);
 
+// helper to check if the device supports a specific family
+// ideally, the user code should be doing these checks
+// ref: https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf
+bool ggml_metal_supports_family(struct ggml_metal_context * ctx, int family);
+
 // set the number of command buffers to use
 void ggml_metal_set_n_cb(struct ggml_metal_context * ctx, int n_cb);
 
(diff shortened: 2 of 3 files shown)


whisper.cpp
```diff
@@ -1078,6 +1078,11 @@ static ggml_backend_t whisper_backend_init(const whisper_context_params & params
         if (!backend_gpu) {
             WHISPER_LOG_ERROR("%s: ggml_backend_metal_init() failed\n", __func__);
         }
+        if (!ggml_backend_metal_supports_family(backend_gpu, 7)) {
```
sindresorhus, line 1081: I don't think this is correct. It should disable it on family 7 **and older**.

sindresorhus: Older would be Apple 6, 5, etc.

ggerganov: Ah sorry, yes.

I thought somebody said that Apple7 works?

sindresorhus: Yes, should be disabled on family 6 and older. I got confused since it's disabling 7 here.

ggerganov: Yup, let me know if you still think this needs to be changed.

The way it is written now it should do the following:

- Apple 1-6 -> fallback to CPU
- Apple 7, 8, 9, .. -> use GPU

The assumption is that if Apple7 is supported, then due to backwards compatibility all previous families (1-6) are also supported, so no need to check for those individually.

Josscii: the order seems to be wrong, we should check the compatibility first, then call ggml_backend_metal_init. otherwise it still crash on lower device.

Josscii: I think, we should do something like this:
```c
#ifdef GGML_USE_METAL
    if (!ggml_backend_metal_supports_family(backend_gpu, 7)) {
        WHISPER_LOG_ERROR("%s: Metal GPU does not support family 7 - falling back to CPU\n", __func__);
    }
    
    if (params.use_gpu) {
        WHISPER_LOG_INFO("%s: using Metal backend\n", __func__);
        ggml_metal_log_set_callback(whisper_log_callback_default, nullptr);
        backend_gpu = ggml_backend_metal_init();
        if (!backend_gpu) {
            WHISPER_LOG_ERROR("%s: ggml_backend_metal_init() failed\n", __func__);
        }
    }
#endif
```

Josscii: I think we should do something like this:

```c
    if (!ggml_backend_metal_supports_family(backend_gpu, 7)) {
        WHISPER_LOG_ERROR("%s: Metal GPU does not support family 7 - falling back to CPU\n", __func__);
    }
    
    if (params.use_gpu) {
        WHISPER_LOG_INFO("%s: using Metal backend\n", __func__);
        ggml_metal_log_set_callback(whisper_log_callback_default, nullptr);
        backend_gpu = ggml_backend_metal_init();
        if (!backend_gpu) {
            WHISPER_LOG_ERROR("%s: ggml_backend_metal_init() failed\n", __func__);
        }
    }
```

but right now, it's not possible, because we can't get the ggml_backend_t if it is not inited, I think there need some refactor

sindresorhus: Was this done?

Josscii: This seems done, right now it do check in ggml_backend_metal_init to avoid crash

PR #1549, opened 2023-11-24
Add audio capture with ffmpeg

rgerganov:
Capture audio from arbitrary audio streams and resample to a specified sample rate. Use the same interface as when capturing from a mic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -345,9 +345,11 @@ clean:
 #
 
 CC_SDL=`sdl2-config --cflags --libs`
+CC_FFMPEG=`pkg-config --cflags --libs libavformat libavcodec libswresample libavutil`
 
 SRC_COMMON     = examples/common.cpp examples/common-ggml.cpp
 SRC_COMMON_SDL = examples/common-sdl.cpp
+SRC_COMMON_FFMPEG = examples/common-ffmpeg.cpp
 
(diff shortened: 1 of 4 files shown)


rgerganov:
I am playing with transcription of network streams and I need to capture and transcribe audio programmatically, without using the ffmpeg binary. Decoding and resampling audio with libav is straightforward using their [examples](https://github.com/FFmpeg/FFmpeg/blob/master/doc/examples/resample_audio.c). The new capture implementation is using the same interface as the one in the `stream` example and can be used as a drop-in replacement:
```
$ streamlink -v --player-external-http 'https://www.twitch.tv/trumporbiden2024' 720p
[cli][info] Found matching plugin twitch for URL https://www.twitch.tv/trumporbiden2024
[cli][info] Available streams: audio_only, 160p (worst), 360p, 480p, 720p_alt, 720p (best)
[cli][info] Starting server, access with one of:
[cli][info]  http://127.0.1.1:39625/

$ ./stream -m ./models/ggml-base.en.bin -t 6 --step 0 --length 10000 -vth 0.7 -u http://127.0.0.1:39625/
```

If you find this useful, I can add another example or rework `stream` to support both mic and network stream.

ggerganov:
Yup, reworking `stream` to be able to work both with mic and URL would be great

PR #1554, opened 2023-11-25
CI : Add CUDA 11.8.0 support

### LINKED ISSUE #1552: Why not runing whisper CUDA on GPU 1050TI on windows?
I downloaded this version:  whisper-cublas-bin-x64.zip
I start decryption, all work goes on the processor.
GPU is not loaded.
The console shows: CUDA is available.
Attached screenshots.
Can you tell me how to run whisper with gpu usage on windows.

Cmd_run_list
CPU_Load

bobqianic:
The reason for the inability to run is due to the incompatibility between the CUDA driver and CUDA toolkit. Previously, the CI for CUDA used version 12.1.0 of the CUDA toolkit, which requires a driver of at least version `>=527.41` on Windows. The file `whisper.cpp` will dynamically link with the runtime in the CUDA toolkit, and the runtime will in turn dynamically link with the driver. The incompatibility causes it to fail to run. The solution is simple: adding a Windows binary compiled with version 11.X of the CUDA toolkit can solve the problem.

image

image

Reference: [here](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER.html#:~:text=CUDA%20Runtime%20API%20calls%20operate,current%20to%20the%20calling%20thread.)

Closes #1552

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -285,7 +285,7 @@ jobs:
         arch: [x64]
         cublas: [ON]
         sdl2: [ON]
-        cuda-toolkit: [12.2.0, 11.8.0, 10.2.89]
+        cuda-toolkit: [12.2.0, 11.8.0]
         include:
           - arch: x64
             s2arc: x64
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -301,9 +301,9 @@ jobs:
 
       - name: Install CUDA Toolkit
         id: cuda-toolkit
-        uses: Jimver/cuda-toolkit@v0.2.10
+        uses: Jimver/cuda-toolkit@v0.2.11
         with:
-          cuda: ${{ matrix.cuda-toolkit }}
+          cuda: '${{ matrix.cuda-toolkit }}'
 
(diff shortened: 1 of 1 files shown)


bobqianic:
> Are there some users who cannot update their drivers?

Yes, there are certain users who might face issues updating their drivers. This includes individuals using Kepler GPUs, such as the K80 or K40 models, and those who do not have administrative privileges on their computers. Also, for instance, some users who have installed other software might need specific versions of CUDA drivers. It would be quite troublesome to reinstall drivers many times just to run whisper.cpp.

bobqianic:
> can't we just say "please update your driver" and keep our CI simple?

Yes. We can also mention that in README. Actually, adding a CI won’t waste much time, as they run in parallel. The total duration of the CI is determined by the longest running one. Currently, the longest ones are armv7 and ppc64le in ubuntu-latest-gcc. In theory, these could be merged with ubuntu-latest, which would speed up the CI process.

ggerganov (approved):
Thank you for looking into this @bobqianic !

Speeding-up the CI would be useful in case you have ideas how to do it!

PR #1561, opened 2023-11-27
Remove #if arch(arm) check in Swift Package Manager

Finnvoor:
I assume the `#if arch(arm) || arch(arm64)` is meant to prevent older Intel macs from using Metal, however checking this here doesn't really work.  This check is executed on the machine compiling the package, and it is very likely that the machine compiling the package is not the machine that a compiled app will run on.  

### Current Behaviour:
- iOS app compiled on an Apple Silicon mac == uses Metal
- iOS app compiled on an Intel mac == won't use Metal
- macOS universal app compiled on Apple Silicon mac == both Intel and Apple Silicon will use metal
- macOS universal app compiled on Intel mac == neither Intel nor Apple Silicon will use metal
This is especially problematic for me as [Xcode Cloud](https://developer.apple.com/xcode-cloud/) builds are currently done on Intel machines, so Metal will not be used in built apps.

This PR enables Metal no matter what the compiling architecture is.  In the case of running a universal app on an Intel mac, there is still the issue that Metal will be enabled when it probably shouldn't be (maybe a runtime check should be added?), however I think it is much more likely that you will want Metal enabled when using SPM (all iOS devices are Apple Silicon, and more and more macs are Apple Silicon)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -2,33 +2,14 @@
 
 import PackageDescription
 
-#if arch(arm) || arch(arm64)
-let platforms: [SupportedPlatform]? = [
-    .macOS(.v12),
-    .iOS(.v14),
-    .watchOS(.v4),
-    .tvOS(.v14)
-]
-let exclude: [String] = []
-let resources: [Resource] = [
-    .process("ggml-metal.metal")
-]
-let additionalSources: [String] = ["ggml-metal.m"]
-let additionalSettings: [CSetting] = [
-    .unsafeFlags(["-fno-objc-arc"]),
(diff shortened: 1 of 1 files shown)


sindresorhus (approved):

sindresorhus:
Should be done in https://github.com/ggerganov/llama.cpp/blob/master/Package.swift too.

// @kchro3 @jhen0409

bobqianic (approved):

kchro3:
Opened a PR for llama.cpp

PR #1565, opened 2023-11-28
Backport .srt output format to examples/server

osdrv:
This commit adds a support for .srt format to Whisper server. The code is effectively backported from examples/main. The output mimetype is set to application/x-subrip as per https://en.wikipedia.org/wiki/SubRip.

Example usage:
```
curl 127.0.0.1:8080/inference \
  -H "Content-Type: multipart/form-data" \
  -F file="@<file-path>" \
  -F temperature="0.2" \
  -F response-format="srt"
```

### DIFF
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -11,6 +11,7 @@
 #include <thread>
 #include <vector>
 #include <cstring>
+#include <sstream>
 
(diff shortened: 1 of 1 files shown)


osdrv:
@ggerganov Thanks very much!

PR #1568, opened 2023-11-28
cmake : install required ggml.h header

gjasny:


lazka:
I went with renaming ggml.h to whisper-ggml.h until this is fixed.

PR #1582, opened 2023-12-01
Fix crash when ggml_backend_metal_init() fails on old Apple device.

bjnortier:
If ggml_backend_metal_init() fails (e.g. iPad Pro M1 2nd Gen), a valid struct is still returned but `ctx` is null.
A subsequent call to ggml_backend_metal_supports_family() crashes because is ctx is null.

So this PR implements:
Return NULL from ggml_backend_metal_init() when context is NULL.
Don't call ggml_backend_metal_supports_family() when backend_gpu is NULL;


slaren:
I think this was fixed in https://github.com/ggerganov/ggml/pull/620

bobqianic:
I found out that this issue seems to have been fixed in a certain update, so I'm now closing this PR.

https://github.com/ggerganov/whisper.cpp/blob/0b9af32a8b3fa7e2ae5f15a9a08f5b10394993f5/whisper.cpp#L1075-L1081

PR #1588, opened 2023-12-03
fix: remove hallucinations from silent audio

ex3ndr:
This line is very important since on silent audio it would heavily hallucinate.

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -4589,7 +4589,7 @@ static void whisper_process_logits(
 
         // suppress sot and nosp tokens
         logits[vocab.token_sot]  = -INFINITY;
-        logits[vocab.token_nosp] = -INFINITY; // TODO: ignore this token for now
+        // logits[vocab.token_nosp] = -INFINITY; // Uncommenting this would produce hallucinations on silent audio
 
         // [TDRZ] when tinydiarize is disabled, suppress solm token
         if (params.tdrz_enable == false) {


ggerganov:
Can we get some examples where this change makes a difference in the output?

bygreencn:
It does really reduce the incidence rate. Even if it occurs a few times, it will only be repeated once. 
Moreover, could anyone check the information at [openai 914](https://github.com/openai/whisper/pull/914)?

jxy:
the openai/whisper skips the current segment if the probability of no speech token is high https://github.com/openai/whisper/blob/e58f28804528831904c3b6f2c0e473f346223433/whisper/transcribe.py#L243-L255

we likely need this?

ex3ndr:
I have re-tested it, but it feels to be marginal improvement. I will try to implement ignoring segments with high probability of silence

ex3ndr:
BTW, i am testing on this short segments that i have recorded. I can't get most of them to be reliably detected as silence in most models.
[silent.zip](https://github.com/ggerganov/whisper.cpp/files/13592502/silent.zip)

bobqianic:
> in most models.

Whisper models?

ex3ndr:
> > in most models.
> 
> Whisper models?

Yes.

ex3ndr:
I have tried to log `nosp` probabilities instead of token probability and it is almost always zero for most cases that i don't think this is the reason it doesn't perform well.

ex3ndr:
image
I have printed out the token then probability of a token then probability of a NOSP token at the same place.

bobqianic:
At least this is heading in the right direction. I'm developing something similar to OpenAI's approach, utilizing `token_nosp` to detect periods of silence.

bobqianic:
> the openai/whisper skips the current segment if the probability of no speech token is high https://github.com/openai/whisper/blob/e58f28804528831904c3b6f2c0e473f346223433/whisper/transcribe.py#L243-L255
> 
> we likely need this?

Sure

whisper.cpp
```diff
@@ -4589,7 +4589,7 @@ static void whisper_process_logits(
 
         // suppress sot and nosp tokens
         logits[vocab.token_sot]  = -INFINITY;
-        logits[vocab.token_nosp] = -INFINITY; // TODO: ignore this token for now
+        // logits[vocab.token_nosp] = -INFINITY; // Uncommenting this would produce hallucinations on silent audio
```
bobqianic, line 4592: Although it is said that `token_nosp` is the direction to solve the hallucination, it is definitely problematic for you to cancel the suppress `token_nosp`. First of all, we only hope that the output of the model contains meaningful and visible tokens (except for timestamps). Your cancellation of suppress `token_nosp` will cause this token to possibly appear in the output of the model, which is something we do not want to see. Secondly, the key to solving hallucination lies in finding a way to skip silence. `token_nosp` is used to tell you how likely it is that this segment is silent, so that we can skip silence. Therefore, merely cancelling suppress `token_nosp` without any other action cannot solve hallucination.

PR #1589, opened 2023-12-03
Fix: some thread safety issues

chatgpt-1:
When running the large-v3 model, long recordings seem to have thread safety issues, randomly causing all subsequent texts to be the same sentence.

Build with CoreML:
```
./models/generate-coreml-model.sh large-v3
WHISPER_COREML=1 make -j
```

Running the command:
```
./main -m models/ggml-large-v3.bin -f /home/test.wav  --print-colors --output-txt true --output-file result
```

Pasted Graphic 12

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -38,6 +38,9 @@
 #include <random>
 #include <functional>
 
+#include <mutex>
+std::mutex kv_mutex;
+
 #if defined(_MSC_VER)
 #pragma warning(disable: 4244 4267) // possible loss of data
 #endif
@@ -996,7 +999,6 @@ static int32_t whisper_kv_cache_cell_max(const struct whisper_kv_cache & cache) 
             return i + 1;
         }
     }
-
     return 1;
 }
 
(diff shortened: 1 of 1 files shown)


whisper.cpp
```diff
@@ -5452,10 +5460,11 @@ int whisper_full_with_state(
                         if (decoder.completed || decoder.failed) {
                             continue;
                         }
-
+                        kv_mutex.lock();
                         whisper_kv_cache_seq_rm(state->kv_self, j,                           -1, -1);
                         whisper_kv_cache_seq_cp(state->kv_self, WHISPER_MAX_DECODERS + j, j, -1, -1);
                         whisper_kv_cache_seq_rm(state->kv_self, WHISPER_MAX_DECODERS + j,    -1, -1);
+                        kv_mutex.unlock();
```
bobqianic, line 5467: Where is multithreading used in the code you referred to?

chatgpt-1: You're right, this is not the root cause, it's just because the voice can come out normally after I made the modification, so I didn't look carefully. It looks like I haven't found the root cause. Still investigating.

chatgpt-1: image

PR #1595, opened 2023-12-06
wchess: whisper assisted chess

fraxy-v:
wchess.asm: Web assembly + [chessboardjs](https://chessboardjs.com/) library
wchess.cmd: Command line binary for testing and debugging.

Note: It seems to me that the (grammar) end-of-token modification used earlier might have had a positive effect. So it may be worth looking into it again.


ggerganov (approved):
Awesome job!

I encounter some bugs from time to time with the chess logic, but it's not a big issue since the main point of the example is to demonstrate the voice recognition functionality.

Btw, I have now added the grammar to be displayed in the UI so this makes debugging a bit easier.

oakmac:
Fun use of chessboardjs here 🤓 🎉

PR #1604, opened 2023-12-07
Add ability to extract encoder embedding.

colinator:
Adds the ability to get the encoder output. No other top-level methods expose ggml_tensors though - I'm not sure that's cool. It seemed the quickest way. What I'm doing, eventually, is this:

```
ggml_tensor * tensor = whisper_get_encoder_embedding(ctx);
std::vector<float> tensor_data(ggml_nelements(tensor));
ggml_backend_tensor_get(tensor, tensor_data.data(), 0, ggml_nbytes(tensor))
```
        
 ... so maybe it should expose a method that returns or populates a vector, instead of returning a ggml_tensor? What do you think?

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -6055,6 +6055,16 @@ float whisper_full_get_token_p(struct whisper_context * ctx, int i_segment, int 
     return ctx->state->result_all[i_segment].tokens[i_token].p;
 }
 
+struct ggml_tensor * whisper_get_encoder_embedding(struct whisper_context * ctx)
+{
+    return ctx->state->embd_enc;
+}
+
+struct ggml_tensor * whisper_get_encoder_embedding_from_state(struct whisper_state * state)
+{
+    return state->embd_enc;
+}
+
 // =================================================================================================
 
 //
--- a/whisper.h
+++ b/whisper.h
@@ -605,6 +605,10 @@ extern "C" {
     WHISPER_API float whisper_full_get_token_p           (struct whisper_context * ctx, int i_segment, int i_token);
     WHISPER_API float whisper_full_get_token_p_from_state(struct whisper_state * state, int i_segment, int i_token);
 
+    // Get the tensor representing the encoder output
+    WHISPER_API struct ggml_tensor * whisper_get_encoder_embedding           (struct whisper_context * ctx);
+    WHISPER_API struct ggml_tensor * whisper_get_encoder_embedding_from_state(struct whisper_state * state);
+        
     ////////////////////////////////////////////////////////////////////////////
 
     // Temporary helpers needed for exposing ggml interface


whisper.cpp
```diff
@@ -6055,6 +6055,16 @@ float whisper_full_get_token_p(struct whisper_context * ctx, int i_segment, int
...
+struct ggml_tensor * whisper_get_encoder_embedding(struct whisper_context * ctx)
+{
+    return ctx->state->embd_enc;
+}
+
+struct ggml_tensor * whisper_get_encoder_embedding_from_state(struct whisper_state * state)
+{
+    return state->embd_enc;
+}
+
```
bobqianic, line 6067: Is it possible to integrate `ggml_backend_tensor_get()` within `whisper_get_encoder_embedding()` so that we can bypass the need to handle `ggml_tensor` directly?

```suggestion
void whisper_get_encoder_embedding(struct whisper_context * ctx, void * data) {
    ggml_tensor * tensor = ctx->state->embd_enc;
    ggml_backend_tensor_get(tensor, data, 0, ggml_nbytes(tensor));
}

void whisper_get_encoder_embedding_from_state(struct whisper_state * state, void * data) {
    ggml_tensor * tensor = state->embd_enc;
    ggml_backend_tensor_get(tensor, data, 0, ggml_nbytes(tensor));
}

```

whisper.h
```diff
@@ -605,6 +605,10 @@ extern "C" {
     WHISPER_API float whisper_full_get_token_p           (struct whisper_context * ctx, int i_segment, int i_token);
     WHISPER_API float whisper_full_get_token_p_from_state(struct whisper_state * state, int i_segment, int i_token);
 
+    // Get the tensor representing the encoder output
+    WHISPER_API struct ggml_tensor * whisper_get_encoder_embedding           (struct whisper_context * ctx);
+    WHISPER_API struct ggml_tensor * whisper_get_encoder_embedding_from_state(struct whisper_state * state);
```
bobqianic, line 610:
```suggestion
    // Get the tensor representing the encoder output
    WHISPER_API void whisper_get_encoder_embedding           (struct whisper_context * ctx, void * data);
    WHISPER_API void whisper_get_encoder_embedding_from_state(struct whisper_state * state, void * data);
```

whisper.cpp
bobqianic: Alright, I see the flaw in my suggestion. It makes the API ineffective because you can't determine the buffer size without first knowing the size of `embd_enc`. If you already know the size, why not add an extra line to fetch the data directly? It seems like we're going around our elbow to get to our thumb : )

bobqianic (approved):

whisper.cpp
```diff
@@ -6055,6 +6055,16 @@ float whisper_full_get_token_p(struct whisper_context * ctx, int i_segment, int
     return ctx->state->result_all[i_segment].tokens[i_token].p;
 }
 
+struct ggml_tensor * whisper_get_encoder_embedding(struct whisper_context * ctx)
+{
+    return ctx->state->embd_enc;
+}
```
bobqianic, line 6061:
```suggestion
struct ggml_tensor * whisper_get_encoder_embedding(struct whisper_context * ctx) {
    return ctx->state->embd_enc;
}
```

whisper.cpp
```diff
@@ -6055,6 +6055,16 @@ float whisper_full_get_token_p(struct whisper_context * ctx, int i_segment, int
...
+
+struct ggml_tensor * whisper_get_encoder_embedding_from_state(struct whisper_state * state)
+{
+    return state->embd_enc;
+}
```
bobqianic, line 6066:
```suggestion
struct ggml_tensor * whisper_get_encoder_embedding_from_state(struct whisper_state * state) {
    return state->embd_enc;
}
```

bobqianic (changes requested):
Style

ggerganov (commented):
> so maybe it should expose a method that returns or populates a vector, instead of returning a ggml_tensor?

Yes, populating a buffer with the data would be more portable compared to returning `ggml_tensor`:

```c
WHISPER_API void whisper_get_encoder_embedding(float * buffer);
```

bobqianic:
> > so maybe it should expose a method that returns or populates a vector, instead of returning a ggml_tensor?
> 
> Yes, populating a buffer with the data would be more portable compared to returning `ggml_tensor`:
> 
> ```c
> WHISPER_API void whisper_get_encoder_embedding(float * buffer);
> ```

This leads to the question: how can user determine the size of the buffer? In this scenario, we end up with only a float pointer, which points to the buffer we've just filled in `whisper_get_encoder_embedding`

ggerganov:
The buffer size can be determined using the model parameters:

https://github.com/ggerganov/whisper.cpp/blob/2623640cd6b8e077ffb1fbe3c265208da3a0ca6d/whisper.h#L347-L358

PR #1607, opened 2023-12-08
Fix Metal API validation errors

Finnvoor:
The Metal API validator doesn't like setting nil for a buffer and throws error `failed assertion Compute Function(kernel_soft_max_4): missing buffer binding at index 1 for src1[0].`

Only setting non-nil buffers fixes the error, even though the binding is still "missing"


ggerganov:
I fixed this in 7bc4d22337c2ae5538e2e6d89aabe157c2b12eb3

Currently, only the softmax kernel requires optional `src1` argument - all other should never be nil.

PR #1617, opened 2023-12-10
target windows 8 or above for prefetchVirtualMemory in llama-talk

Kreijstal:
Since we use prefetchVirtualMemory we specify we target win 8 or above, otherwise other compilers will refuse to use the prefetchVirtualMemory api, (I understand you are loading it dynamically but the header definition has this limitation)


bobqianic:
https://github.com/ggerganov/llama.cpp/pull/4405

bobqianic (approved):

PR #1649, opened 2023-12-16
auto transcription implemented

artshcherbina:


jensdraht1999:
@artshcherbina Hello, I think, this is good, but I wanted to asked, if this can be implemented also to be in the .srt file and so on? And if yes, what happens, does the timestamps correctly align with the original file?

I implemented something similar in Windows via ffmpeg and silence detection.

1) Split the audio in 1 second samples.
2.) If the audio is too low (50db), then regard it as silence.
3.) Replace all audio samples, that are regarded as silent with a sound sample (for example white noise / or any other voice or silent audio)

However, this did not improve anything unfortunately.

The sad thing is, that Whisper is trying to transcribe literally everything.

Do you any audio samples, I could try to test out with my script, I have a few scripts, which I would like to test out and share with the community, if they are working good or better.

artshcherbina:
Hello, @jensdraht1999 .
I haven't worked with .srt files.
Currently, I don't have any samples.
You may record some samples with your mic for testing

jensdraht1999:
> Hello, @jensdraht1999 . I haven't worked with .srt files. Currently, I don't have any samples. You may record some samples with your mic for testing

I wonder, if you could resolve the rest of the issues and then merge it.

artshcherbina:
My changes break some default behavior. 
So I have no plan to merge this PR now.

PR #1651, opened 2023-12-16
talk-llama: speed up with Metal and CoreML

Maverobot:
Previously, talk-llama did not leverage Metal and CoreML functionalities when built with `cmake`, leading to unexpectedly poor performance on Apple Silicon devices.

This pull request addresses the issue by including necessary sources, libraries, and compile flags. As a result, talk-llama has been significantly optimized for performance. On my M2 Ultra Mac Studio, these modifications have notably improved the user experience.

Unfortunately, due to the unavailability of a substantial NVIDIA GPU, I haven't been able to verify if these changes positively impact performance on CUDA PC setups.

### DIFF
--- a/examples/talk-llama/CMakeLists.txt
+++ b/examples/talk-llama/CMakeLists.txt
@@ -1,3 +1,11 @@
+function(prefix_all NEW_LIST_NAME LIST_NAME PREFIX)
+    set(NEW_LIST "")
+    foreach(ITEM ${${LIST_NAME}})
+        set(NEW_LIST ${NEW_LIST} "${PREFIX}${ITEM}")
+    endforeach()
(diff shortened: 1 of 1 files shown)


bobqianic:
I'm sorry that I can't merge your PR into master. The main reason for the issue you described is that during a previous sync of llama, the CMakeLists.txt was accidentally uploaded incorrectly, resulting in talk-llama not linking the whispercpp library files. This issue has been fixed in https://github.com/ggerganov/whisper.cpp/pull/1669. Personally, I think dynamically linking the whispercpp library would be better, as it can reduce the size.

You can try the latest version of whisper.cpp to see if it has fixed the issue you mentioned. If the problem still exists, feel free to open a new PR.

PR #1658, opened 2023-12-20
<feat>: Add android logcat support.

xbwee1024:
debug log will disapear if it's called in android app, so this PR add logcat instead of printf.

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -118,6 +118,9 @@ static void byteswap_tensor(ggml_tensor * tensor) {
 // logging
 //
 
+#ifdef WHISPER_LOGCAT
+#include <android/log.h>
+#endif
 WHISPER_ATTRIBUTE_FORMAT(2, 3)
 static void whisper_log_internal        (ggml_log_level level, const char * format, ...);
 static void whisper_log_callback_default(ggml_log_level level, const char * text, void * user_data);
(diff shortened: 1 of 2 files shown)


whisper.cpp
```diff
@@ -118,6 +118,9 @@ static void byteswap_tensor(ggml_tensor * tensor) {
 // logging
 //
 
+#ifdef WHISPER_LOGCAT
+#include <android/log.h>
+#endif
```
bobqianic, line 123: To be honest, I'm not convinced that including `<android/log.h>` in whisper.cpp is the best approach. It seems more practical to use a callback instead. 

I'm also wondering why do we continue to use `WHISPER_PRINT_DEBUG` when we already have `WHISPER_LOG`? Couldn't we just create a function named `WHISPER_LOG_DEBUG` that offers both console output and callback functionality? @ggerganov

xbwee1024: Well, I see the callback logic just now,  I agree with you, it's the right way!

ggerganov: Yup, logcat is better to not be part of `whisper.cpp` - utilize the log callback

Regarding `WHISPER_PRINT_DEBUG` - it's used for debugging that I don't think will be really useful to the "outside world". But we could add `WHISPER_LOG_DEBUG` instead for consistency. Probably have to extend `enum ggml_log_level` with new `GGML_LOG_LEVEL_DEBUG` value

bobqianic:
I am now closing this PR because you can now use a callback to get all the log information, including DEBUG.

Issue #1661, opened 2023-12-20
Quantized model is not working properly when CUBLAS is ON


bobqianic:
### Discussed in https://github.com/ggerganov/whisper.cpp/discussions/1656

<div type='discussions-op-text'>

<sup>Originally posted by **Sing303** December 19, 2023</sup>
Now when I try to use quantization models with FULL GPU CUBLAS, instead of recognizing text it writes nonsense, all kinds of signs instead of words.</div>

Sing303:
At the same time we created :)
https://github.com/ggerganov/whisper.cpp/issues/1662

bobqianic:
**Tried syncing the latest ggml from llama.cpp, but no luck. Both talk-llama and whisper are behaving identically.**

`tinyllama-1.1b-chat-v0.3.Q4_0.gguf` `ggml-model-whisper-base.bin`

![image](https://github.com/ggerganov/whisper.cpp/assets/129547291/280ebb16-2cd8-4d73-b41f-ab50b0b08a64)

`ggml-model-whisper-base-q5_1.bin`

![image](https://github.com/ggerganov/whisper.cpp/assets/129547291/83d71b75-040a-47ea-8961-b4935a8c960b)

bobqianic:
The output logits of quantized models are near zero which is not correct. I dumped the `embd_enc` values, and I noticed that when using CUDA as the backend, the results significantly differ from those obtained using a CPU backend.

bobqianic:
Do you know how to fix this bug? @slaren @ggerganov

bobqianic:
I find it strange, if the error occurred during the `build_graph` process, then all backends should have a problem. Currently, it is known that only the CUDA backend has an issue. I have already synchronized the latest `ggml` from llama.cpp. But if the CUDA backend really has a problem, llama.cpp should have discovered this issue much earlier. Maybe it's because few people use it? I will try other quantization modes besides Q5_1.

bobqianic:
Okay, I've tested all the quantization modes, and none of them work properly on CUDA. I wonder if our quantization method has changed? All `wstate_embd_enc` outputs have one thing in common: **`there is a significant difference`** between the outputs on CUDA and CPU backends, which does not exist in the properly functioning FP16 mode.

slaren:
As far as I can tell, quantized models work correctly both with the gpt-2 example in ggml and in llama.cpp. I am not sure what is different in whisper.cpp.

bobqianic:
[...]

PR #1667, opened 2023-12-20
cmake : Resolve quantized model issue when CUBLAS enabled

bobqianic:
Thanks for the help provided by @slaren, the problem is finally solved!
See #1661 for details


Sing303:
Something similar has been done here https://github.com/ggerganov/whisper.cpp/pull/1499

ggerganov (approved):

PR #1669, opened 2023-12-20
examples : Revert CMakeLists.txt for talk-llama

bobqianic:
The current version of CMakeLists.txt isn't functioning correctly with the `-DWHISPER_CUBLAS=1` setting. Additionally, we need to address the CI setup, as it currently lacks coverage for this setting.


ggerganov (approved):
Thank you!

PR #1672, opened 2023-12-21
CI : Add coverage for talk-llama when WHISPER_CUBLAS=1

bobqianic:
It's normal for the CI of this PR to not pass. We should first merge #1669, and then the CI of this PR will pass, because there is an issue with talk-llama's CMake.


ggerganov (approved):
Merge if CI is green

PR #1674, opened 2023-12-21
ci: build and push docker image

OpenWaygate:
This pr add two Dockerfiles and an action to build and push docker image to ghcr, I also update the readme about the docker section.

A lot of work is finished by reference the corresponding part in [ggerganov/llama.cpp](https://github.com/ggerganov/llama.cpp), which is really helps!

### DIFF
--- a/models/download-ggml-model.sh
+++ b/models/download-ggml-model.sh
@@ -56,8 +56,8 @@ function list_models {
     printf "\n\n"
 }
 
-if [ "$#" -gt 2 ]; then
-    printf "Usage: $0 <model> <models_path>\n"
+if [ "$#" -lt 1 ] || [ "$#" -gt 2 ]; then
+    printf "Usage: $0 <model> [models_path]\n"
     list_models
 
     exit 1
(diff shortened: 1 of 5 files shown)


models/download-ggml-model.sh
```diff
@@ -56,8 +56,8 @@ function list_models {
     printf "\n\n"
 }
 
-if [ "$#" -ne 1 ]; then
-    printf "Usage: $0 <model>\n"
+if [ "$#" -gt 2 ]; then
```
bobqianic, line 59:
```suggestion
if [ "$#" -lt 1 ] || [ "$#" -gt 2 ]; then
```

models/download-ggml-model.sh
```diff
@@ -56,8 +56,8 @@ function list_models {
...
+    printf "Usage: $0 <model> <models_path>\n"
```
bobqianic, line 60:
```suggestion
    printf "Usage: $0 <model> [<models_path>]\n"
```

bobqianic (changes requested):

bobqianic:
Let me get some sleep first, and I’ll merge it later today.

bobqianic (approved):

PR #1677, opened 2023-12-22
sync : ggml (ggml_scale, ggml_row_size, etc.)

ggerganov:


ggml-alloc.c
```diff
@@ -449,11 +449,10 @@ static void init_view(ggml_gallocr_t galloc, struct ggml_tensor * view, bool upd
     if (update_backend) {
         view->backend = view->view_src->backend;
     }
-    view->buffer  = view->view_src->buffer;
+    // views are initialized in the alloc buffer rather than the view_src buffer
+    view->buffer  = alloc->buffer;
```
ggerganov, line 453: @slaren After the sync, the following command produces incorrect transcription:

```bash
make samples
make -j && ./main -m ./models/ggml-medium.en.bin -f ./samples/gb0.wav
```

Reverting just this line fixes the issue. Any guess what this could be related to?
Maybe something is incompatible with the `whisper_allocr_graph_realloc` logic

slaren: I think it is because this change interferes with auto-inline logic. This should be fix it:

```diff
diff --git a/ggml-alloc.c b/ggml-alloc.c
index a97436b..a27dd54 100644
--- a/ggml-alloc.c
+++ b/ggml-alloc.c
@@ -72,7 +72,7 @@ static void remove_allocated_tensor(ggml_tallocr_t alloc, struct ggml_tensor * t

 // check if a tensor is allocated by this buffer
 static bool ggml_tallocr_is_own(ggml_tallocr_t alloc, const struct ggml_tensor * tensor) {
-    return tensor->buffer == alloc->buffer;
+    return tensor->buffer == alloc->buffer && (!tensor->view_src || tensor->view_src->buffer == alloc->buffer);
 }

 static bool ggml_is_view(struct ggml_tensor * t) {
```

Issue #1664, opened 2023-12-20
Crash at __ggml_metal_graph_compute due to failed assertion -- SIGABRT ggml_metal_graph_compute: command buffer 0 failed with status 5


Digipom:
Whisper 1.5.2 via Swift package

Whisper logs:
```
whisper_init_from_file_with_params_no_state: loading model from '/var/mobile/Containers/Data/Application/6C3E9FA2-AC0F-4DA5-BD68-ADC424E7F5CE/Library/Application Support/ggml-base.bin'
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51865
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 512
whisper_model_load: n_audio_head  = 8
whisper_model_load: n_audio_layer = 6
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 512
whisper_model_load: n_text_head   = 8
whisper_model_load: n_text_layer  = 6
whisper_model_load: n_mels        = 80
whisper_model_load: ftype         = 1
whisper_model_load: qntvr         = 0
whisper_model_load: type          = 2 (base)
whisper_model_load: adding 1608 extra tokens
whisper_model_load: n_langs       = 99
whisper_backend_init: using Metal backend
ggml_metal_init: allocating
ggml_metal_init: picking default device: Apple A15 GPU
ggml_metal_init: loading '/var/containers/Bundle/Application/078A9408-F9A4-4B75-8AAC-2AADFD3FC35C/EasyTranscription.app/whisper_whisper.bundle/default.metallib'
whisper_model_load:    Metal buffer size =   147.47 MB
whisper_model_load: model size    =  147.37 MB
whisper_backend_init: using Metal backend
ggml_metal_init: allocating
ggml_metal_init: picking default device: Apple A15 GPU
ggml_metal_init: loading '/var/containers/Bundle/Application/078A9408-F9A4-4B75-8AAC-2AADFD3FC35C/EasyTranscription.app/whisper_whisper.bundle/default.metallib'
whisper_init_state: kv self size  =   16.52 MB
whisper_init_state: kv cross size =   18.43 MB
whisper_init_state: compute buffer (conv)   =   14.92 MB
whisper_init_state: compute buffer (encode) =   86.06 MB
whisper_init_state: compute buffer (cross)  =    4.84 MB
whisper_init_state: compute buffer (decode) =   96.55 MB
[...]

PR #1714, opened 2024-01-02
Add error handling to graph_compute

### LINKED ISSUE #1670: METAL: The library on iOS, Swift code, fails when running a transcription in background
Hi,
running the transcription with the app in foreground, when passing to background I immediately got an exception in whisper.cpp.
I am using whisper.cpp with METAL, so I'm running on GPU.
I suppose that this is the problem: https://developer.apple.com/documentation/metal/gpu_devices_and_work_submission/preparing_your_metal_app_to_run_in_the_background

I'm using the latest commit available, `940de9d`, and the problem occurs here, in file `ggml-metal`:

Screenshot 2023-12-21 alle 13 43 22

What are the possible strategies to not going in error or to pause the inference and get back to the transcription progress when the app came back to foreground ?

Thank you so much for your work!

finnvoor:
We are using whisper.cpp in an iOS app and due to GPU execution not being permitted when an iOS app is in the background, we are seeing crashes at https://github.com/ggerganov/whisper.cpp/blob/f9ca90256bf691642407e589db1a36562c461db7/ggml-metal.m#L2385

This PR returns a bool from graph_compute indicating failure instead of asserting false to avoid crashing.  Wasn't sure if this was the best way to fix this issue, but it is working for us at least.  All backends other than metal currently just return true.

Please let me know if I missed something or if anyone has better ideas for error handling.

Fixes #1670 and #1664

### DIFF
--- a/bindings/ruby/ext/ggml-backend-impl.h
+++ b/bindings/ruby/ext/ggml-backend-impl.h
@@ -70,7 +70,7 @@ extern "C" {
         void                      (*graph_plan_compute)(ggml_backend_t backend, ggml_backend_graph_plan_t plan);
 
         // compute graph without a plan
-        void (*graph_compute)(ggml_backend_t backend, struct ggml_cgraph * cgraph);
+        bool (*graph_compute)(ggml_backend_t backend, struct ggml_cgraph * cgraph);
 
         // check if the backend supports an operation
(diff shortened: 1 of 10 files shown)


ggerganov (approved):
Should be good to merge (cc @slaren)

slaren (approved):
I think this is good, we need to make more errors recoverable.

PR #1725, opened 2024-01-05
refactor: get all scripts to be POSIX Compliant

sonphantrung:
It will make scripts faster in [Debian/Ubuntu](https://askubuntu.com/a/976504) systems, and remove the need to install bash for FreeBSD users :D.


ggerganov:
`download-ggml-model.sh` fails on my Mac:

```bash
$ echo $SHELL
/bin/sh

$ ./models/download-ggml-model.sh base.en
Invalid model: %model

  Available models: tiny.en tiny tiny-q5_1 tiny.en-q5_1 base.en base base-q5_1 base.en-q5_1 small.en small.en-tdrz small small-q5_1 small.en-q5_1 medium medium.en medium-q5_0 medium.en-q5_0 large-v1 large-v2 large-v3 large-v3-q5_0

```

sonphantrung:
@ggerganov I think it should be good now.

PR #1727, opened 2024-01-05
chore: add missing large-v2 variant to scripts

mightymatth:
As `large` variant became the default for `large-v3` model, `large-v2` variant was missing in a number of scripts.
This PR adds `large-v2` to missing places.


bobqianic:
You should update your branch first, as it seems quite outdated.

mightymatth:
yep, wrong remote..

PR #1728, opened 2024-01-05
fix temperature parameter in http api

ex3ndr:
Right now 'temperature' in request changes a wrong parameter, which could confuse a lot of people.

### DIFF
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -602,7 +602,7 @@ int main(int argc, char ** argv) {
             wparams.greedy.best_of        = params.best_of;
             wparams.beam_search.beam_size = params.beam_size;
 
-            wparams.temperature_inc  = params.userdef_temp;
+            wparams.temperature  = params.userdef_temp;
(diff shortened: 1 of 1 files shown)


bobqianic:
However, given the default value of `userdef_temp` in server.cpp, it turns out that `userdef_temp` is indeed `temperature_inc`

https://github.com/ggerganov/whisper.cpp/blob/70ccda5669cffeb9bcc7f7d61f141d8b9f61dac3/examples/server/server.cpp#L66

I think it would be clearer if we renamed the variable and request keyword to something more intuitive.

https://github.com/ggerganov/whisper.cpp/blob/70ccda5669cffeb9bcc7f7d61f141d8b9f61dac3/examples/server/server.cpp#L423-L426

https://github.com/ggerganov/whisper.cpp/blob/0b9af32a8b3fa7e2ae5f15a9a08f5b10394993f5/whisper.cpp#L5023-L5032

ggerganov:
I've pushed a proper fix in https://github.com/ggerganov/whisper.cpp/pull/1729 - please review

PR #1735, opened 2024-01-08
Remove unnecessary logging and added python file for python processing

contractorwolf:
I am using this whisper.cpp library for ultra fast processing using apple silicon (my M1 Macbook). I couldnt find a great way to do it until I found this library.  The other Python usages didnt seem to allow for much customization and I wanted to run the main with limited logging messages, so I created this branch.  This just adds a simple wrapper for calling the main, to run it you can just do this from the command line:

`>python stt_wav.py "./wake_word_detected16k.wav"`

(if the file is at the root)

feedback appreciated


ggerganov:
This is a better way to do it: https://github.com/ggerganov/whisper.cpp/pull/1740

PR #1744, opened 2024-01-09
rebase against your -np changes (thx) and add better python file to be used on the command line or as library

contractorwolf:
can be used on the command line like this:
> python whisper_processor.py "./audio/wake_word_detected16k.wav"

or included in a python file like this:
```
import whisper_processor

try:
    result = whisper_processor.process_audio("./audio/wake_word_detected16k.wav", "base.en")
    print(result)
except Exception as e:
    print(f"Error: {e}")
```

see test_whisper_processor.py for example, adjust paths as necessary


contractorwolf:
any chance on getting this in @ggerganov? I would really like it if I could say I contributed to this great project

ggerganov:
Create a folder `examples/processor.py`, move the files there and we can merge

ggerganov (approved):

PR #1747, opened 2024-01-10
sync: ggml

zshannon:
Loading GGUF model throws an assertion error like was [reported in llama.cpp](https://github.com/ggerganov/llama.cpp/issues/4837) when using whisper & llama side-by-side because the ggml dep is out of sync.

I ran `./extra/sync-ggml.sh` to generate this PR, not sure if there's more to it than that but doing so fixed the crasher for me.


ggerganov:
I'll try to sync the project later today. Need to bring some stuff from `llama.cpp` first

PR #1749, opened 2024-01-10
Optional Piper TTS support for talk-llama example.

RhinoDevel:
Add commented-out command to optionally use Piper (https://github.com/rhasspy/piper) as text-to-speech solution for the talk-llama example. Piper voices sound almost like real people which is a big improvement (e.g.) from something like espeak.


chenrq2005:
Thanks for continuously adding functions for the output options!

Issue #1748, opened 2024-01-10
Sample SwiftUI app crashes on model load


cyrilzakka:
The accompanying SwiftUI example app crashes on all models tested on the assert:
```
whisper_init_from_file_with_params_no_state: loading model from '/private/var/containers/Bundle/Application/FBE0AA37-658E-4E0E-A5D7-326D14E1C818/whisper.swiftui.app/models/ggml-tiny.en.bin'
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51864
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 384
whisper_model_load: n_audio_head  = 6
whisper_model_load: n_audio_layer = 4
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 384
whisper_model_load: n_text_head   = 6
whisper_model_load: n_text_layer  = 4
whisper_model_load: n_mels        = 80
whisper_model_load: ftype         = 1
whisper_model_load: qntvr         = 0
whisper_model_load: type          = 1 (tiny)
whisper_model_load: adding 1607 extra tokens
whisper_model_load: n_langs       = 99
whisper_backend_init: using Metal backend
ggml_metal_init: allocating
ggml_metal_init: picking default device: Apple A16 GPU
ggml_metal_init: loading '/var/containers/Bundle/Application/FBE0AA37-658E-4E0E-A5D7-326D14E1C818/whisper.swiftui.app/ggml_ggml.bundle/default.metallib'
ggml_backend_metal_buffer_type_alloc_buffer: allocated buffer, size =    73.61 MiB, (   74.73)
whisper_model_load:    Metal buffer size =    77.18 MB
whisper_model_load: model size    =   77.11 MB
whisper_backend_init: using Metal backend
ggml_metal_init: allocating
ggml_metal_init: picking default device: Apple A16 GPU
ggml_metal_init: loading '/var/containers/Bundle/Application/FBE0AA37-658E-4E0E-A5D7-326D14E1C818/whisper.swiftui.app/ggml_ggml.bundle/default.metallib'
ggml_backend_metal_buffer_type_alloc_buffer: allocated buffer, size =     7.88 MiB, (   82.61)
whisper_init_state: kv self size  =    8.26 MB
ggml_backend_metal_buffer_type_alloc_buffer: allocated buffer, size =     8.80 MiB, (   91.41)
whisper_init_state: kv cross size =    9.22 MB
[...]

PR #1750, opened 2024-01-10
fix/1748

nank1ro:
closes https://github.com/ggerganov/whisper.cpp/issues/1748


ggerganov:
Did you verify that the examples build and work correctly?
I'm asking because I have some memory that using a commit instead of a branch can cause issues - not really sure though.

nank1ro:
Yes just tried on my Mac and on my iPhone and it works

PR #1772, opened 2024-01-15
server : fix building and simplify lib deps on Windows

przemoc:


przemoc:
This change is similar to sister change in llama.cpp from 0.5y ago, but related to different example:

- ggerganov/llama.cpp/pull/2235

ggerganov (approved):

przemoc:
Current `freeBSD-latest` failure of commit f5f159c320d3d4be0ccb9746cde5323933ebc453 in master branch:
https://github.com/ggerganov/whisper.cpp/actions/runs/7529770278/job/20494705057

```
  Mounting disk
  Finished partitioning on disk2
  /usr/bin/sudo umount /Volumes/RES
  /usr/bin/hdiutil detach /dev/disk2
  hdiutil: couldn't eject "disk2" - Resource busy
```

seems to be some temporary environment/setup issue and not related in any way to changes introduced via this PR.

ggerganov:
Yes, this CI job fails quite often in this way - nothing we can do

PR #1833, opened 2024-02-04
New example: Bidirectional gRPC streaming (async) transcription

shanelenagh:
In playing with a potential use case of whisper.cpp, I found it expedient to implement a gRPC interface to whisper.cpp, particularly the streaming example of it.  This version uses the gRPC bidirectional streams based on a simple protobuf that I came up with.  This allows whisper.cpp to be called as a realtime service from a variety of client environments using the fairly efficient and flexible gRPC protocol.  I didn't include it in this PR, but I have done two different Python clients to demonstrate the "client" side of this gRPC interaction (this stream.grpc code being the "server"), both as a CLI client and a minimal web client (using websockets)--if there is interest in having this included in the main PR here I could add them as a subdirectory as well:

https://github.com/shanelenagh/whisper.cpp/tree/stream.grpc/examples/stream.grpc/python-test-clients

I don't claim to be a C++ expert at all; started my studies in programming with C in 1995, but over my career as a developer I have spent 90% of it in Java--so there are no doubt ways the code can be improved.  But given this is just an "example," and I have indeed proven that it appears to work in my use case (sure there is some bug or two lurking) I thought it safe to submit the PR in the current state.  Let me know if you think the broader community might find this useful, or if you need additional information (screencasts of usage, etc.).  You do need the C++ version of gRPC installed to do this build (will add that doc to the readme here shortly).

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -14,6 +14,11 @@ if (WHISPER_SDL2)
     message(STATUS "SDL2_LIBRARIES = ${SDL2_LIBRARIES}")
 endif()
 
+if (WHISPER_GRPC)
+    # GRPC
+    include(GRPC.cmake)
+endif()
+
 if (WHISPER_CLBLAST)
     find_package(CLBlast REQUIRED)
 endif()
@@ -54,6 +59,64 @@ if (WHISPER_SDL2)
     set_target_properties(${TARGET} PROPERTIES POSITION_INDEPENDENT_CODE ON)
 endif()
 
+if (WHISPER_GRPC)
+    # common-grpc
+
+    set(TARGET common-grpc)
+
+    #
+    # GENERATE PROTOBUF CODE
+    #
+
+    # Proto file
+    get_filename_component(sg_proto "transcription.proto" ABSOLUTE)
+    get_filename_component(sg_proto_path "${sg_proto}" PATH)
+
+    # Generated sources
+    set(sg_proto_srcs "transcription.pb.cc")
+    set(sg_proto_hdrs "transcription.pb.h")
+    set(sg_grpc_srcs "transcription.grpc.pb.cc")
+    set(sg_grpc_hdrs "transcription.grpc.pb.h")
+    add_custom_command(
+        OUTPUT "${sg_proto_srcs}" "${sg_proto_hdrs}" "${sg_grpc_srcs}" "${sg_grpc_hdrs}"
+        COMMAND ${_PROTOBUF_PROTOC}
+        ARGS --grpc_out "${CMAKE_CURRENT_BINARY_DIR}"
+            --cpp_out "${CMAKE_CURRENT_BINARY_DIR}"
+            -I "${sg_proto_path}"
+            --plugin=protoc-gen-grpc="${_GRPC_CPP_PLUGIN_EXECUTABLE}"
+            "${sg_proto}"
+        DEPENDS "${sg_proto}")    
+
+    #
+    # BUILD common GRPC audio library (with our protobuf)
+    #
+
+    # Common deps
+    set(sg_common_srcs "common-grpc.cpp")
+    set(sg_common_hdrs "common-grpc.h")    
+    
+    # Include generated *.pb.h files
+    include_directories("${CMAKE_CURRENT_BINARY_DIR}")
+
+    add_library(${TARGET} 
+        ${sg_grpc_srcs}
+        ${sg_grpc_hdrs}
+        ${sg_proto_srcs}
+        ${sg_proto_hdrs}
+        ${sg_common_srcs}
+        ${sg_common_hdrs}  
+    )
+    target_link_libraries(${TARGET} 
+        ${_REFLECTION}
+        ${_GRPC_GRPCPP}
+        ${_PROTOBUF_LIBPROTOBUF}
+    )    
+
+    include(DefaultTargetOptions)
+
+    set_target_properties(${TARGET} PROPERTIES POSITION_INDEPENDENT_CODE ON)
+endif()
+
 # examples
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
@@ -69,6 +132,7 @@ elseif(CMAKE_JS_VERSION)
 else()
     add_subdirectory(main)
     add_subdirectory(stream)
+    add_subdirectory(stream.grpc)
     add_subdirectory(server)
     add_subdirectory(command)
     add_subdirectory(bench)
--- a/examples/GRPC.cmake
+++ b/examples/GRPC.cmake
@@ -1,0 +1,121 @@
+# Copyright 2018 gRPC authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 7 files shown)


shanelenagh:
README.md added

ggerganov:
Thank you for this example - I'm not familiar with gRPC, but it does seem useful!

I'm not sure if we can merge this however - it brings a bit of 3rd parties (protobuf, gRPC). I try to keep the dependencies in the project to a minimum, otherwise it becomes difficult to maintain and it also means less portability.

There is a pending PR #1823 with stdin streaming example which I am more interested in merging. Similar to this PR, it also builds on top of the existing `stream` example. Want to see if it will find a way to reuse the `stream` code instead of copy-pasting it and if it does, maybe the same strategy can be applied here as well

But still, the dependencies might remain a problem. I guess this example could live in it's own repository. If you decide to do it like that, I will happily reference it from here

shanelenagh:
> Thank you for this example - I'm not familiar with gRPC, but it does seem useful!
> 
> I'm not sure if we can merge this however - it brings a bit of 3rd parties (protobuf, gRPC). I try to keep the dependencies in the project to a minimum, otherwise it becomes difficult to maintain and it also means less portability.
> 
> There is a pending PR #1823 with stdin streaming example which I am more interested in merging. Similar to this PR, it also builds on top of the existing `stream` example. Want to see if it will find a way to reuse the `stream` code instead of copy-pasting it and if it does, maybe the same strategy can be applied here as well
> 
> But still, the dependencies might remain a problem. I guess this example could live in it's own repository. If you decide to do it like that, I will happily reference it from here

Fair enough.  The zero-dependency aspect of whisper.cpp is one of the things I love most about it (the speed and accuracy being awesome as well, of course).  I structured this change the same way the original stream example was setup, whereas if you don't set the variable it will not build it and thus not require that dependency, and yes, it similarly requires a separate library (e.g., SDL in the case of the original, and gRPC in this one--and to be clear protobuf is part of gRPC, so that is a single dependency there).  I guess OpenVINO and CUBLAS are somewhat similar as well (as an aside, I am really glad OpenVINO was added as an option, as I see tremendous improves on the hardware we are using), but I guess those are more core to what the library does.  But yeah, if you do want to use the gRPC interface and build it, you will need gRPC installed.

I have my original work in a fork of this, and I will likely just keep it there, with this as upstream, unless someone else sees some utility in a "separate" project and convinces me that is worth doing.  I want to be able to rebase from whisper.cpp frequently to get the improvements being made, so that seems to be the easiest way.  And I want it to be known that I definitely on the shoulders of giants like you and your team for the core value of the tool. :-)  Thank you so much for the wonderful work you and the team have done on this.

shanelenagh:
I just took a look at the stdin PR, and yeah, we both seem to be wrestling with the duplication issue.  I wonder if an "abstract" (had to look up the CPP parlance) class for audio_async could be defined for any one of us wanting to supply an "audio provider" for the stream service.  Even the concrete implementation of most of the audio_async methods of our versions is 90% the same as the original SDL version--those could be implemented once and then have a pure virtual "get_bytes" method that is implemented by the concrete provider?  Then the stream.cpp code could be the same as well (say calling a factory method that passes the params struct, with a new param specifying the "audio input interface" for SDL, stdin, etc.).  I would love to take a stab at refactoring that out, if I can get a bit more time.

shanelenagh:
> I just took a look at the stdin PR, and yeah, we both seem to be wrestling with the duplication issue. I wonder if an "abstract" (had to look up the CPP parlance) class for audio_async could be defined for any one of us wanting to supply an "audio provider" for the stream service. Even the concrete implementation of most of the audio_async methods of our versions is 90% the same as the original SDL version--those could be implemented once and then have a pure virtual "get_bytes" method that is implemented by the concrete provider? Then the stream.cpp code could be the same as well (say calling a factory method that passes the params struct, with a new param specifying the "audio input interface" for SDL, stdin, etc.). I would love to take a stab at refactoring that out, if I can get a bit more time.

I took the stdin version (copied @regularfry's branch to my fork), and I am starting to refactor the existing stream method to use the following abstract class I created in a new "common-asyncaudio.h"--compiles and runs, but that's as far as I have gotten so far (next step is to move the stdin to use that abstraction, add the param for the input source, add the factory method for dynamically returning the concrete instance that is specified in the param, etc.): https://github.com/shanelenagh/whisper.cpp/commit/9c1a270a346e61441e73120dad191f97799f3e0e

```c++
#pragma once


#include <atomic>
#include <cstdint>
#include <vector>
#include <mutex>
#include <thread>

// command-line parameters
struct whisper_params {
    int32_t n_threads  = std::min(4, (int32_t) std::thread::hardware_concurrency());
    int32_t step_ms    = 3000;
    int32_t length_ms  = 10000;
    int32_t keep_ms    = 200;
    int32_t capture_id = -1;
    int32_t max_tokens = 32;
    int32_t audio_ctx  = 0;

    float vad_thold    = 0.6f;
    float freq_thold   = 100.0f;

    bool speed_up      = false;
    bool translate     = false;
    bool no_fallback   = false;
    bool print_special = false;
    bool no_context    = true;
    bool no_timestamps = false;
    bool tinydiarize   = false;
    bool save_audio    = false; // save audio to wav file
    bool use_gpu       = true;

    std::string language  = "en";
    std::string model     = "models/ggml-base.en.bin";
    std::string fname_out;
};

//
// Abstract interface for audio capture
//
class audio_async {
public:
    audio_async(int len_ms) { };
    ~audio_async() { };

    virtual bool init(whisper_params params, int sample_rate) = 0;

    virtual bool resume() = 0;
    virtual bool pause() = 0;
    virtual bool clear() = 0;

    // get audio data from the circular buffer
    virtual void get(int ms, std::vector<float> & audio) = 0;

    // callback to be called by audio source
    virtual void callback(uint8_t * stream, int len) = 0;
};
```
and then the SDL one implements that interface (snippet of modified common-sdl.h):
```c++
//
// SDL Audio capture
//
class audio_async_sdl : public audio_async {
public:
    audio_async_sdl(int len_ms);
    ~audio_async_sdl();

    bool init(whisper_params params, int sample_rate) override;

    // start capturing audio via the provided SDL callback
    // keep last len_ms seconds of audio in a circular buffer
    bool resume() override;
    bool pause() override;
    bool clear() override;

    // callback to be called by SDL
    void callback(uint8_t * stream, int len) override;

    // get audio data from the circular buffer
    void get(int ms, std::vector<float> & audio) override;

private:
    SDL_AudioDeviceID m_dev_id_in = 0;
    // etc...
};
```

shanelenagh:
Been taken away from the project for a few days by work and home life, but I just factored out the common code into the abstract class--now just the SDL code remains in the subclass (next step is to make the stdin version do the same, and create the factory method that creates the right version based on a new CLI param): https://github.com/shanelenagh/whisper.cpp/commit/2a2307bbcc6b4375c2073d87af7759776a98f0c2

muety:
Would love to have this merged! :raised_hands:

atmosfar:
@shanelenagh 

Hi, 

I'm running into some issues when trying to build a script on top of whisper-stream, specifically to do with the varying number of lines written and overwritten before final output of the text. It looks like you may have overcome the same problem when writing your python code. Would you mind adding some more comments to your "[print_utterance](https://github.com/ggml-org/whisper.cpp/blob/4383bb18ca57d1074ad1a1ab144befad207e5a09/examples/stream.grpc/python-test-clients/audio_sender.py#L32)" function to explain your logic a bit more clearly? It seems like you're going for some kind of local agreement approach but I don't completely follow it.

I tried using a simple counter approach based on the "[params.length_ms / params.step_ms - 1](https://github.com/ggml-org/whisper.cpp/blob/master/examples/stream/stream.cpp#L137)" calculation in the cpp but it isn't reliable and often ends up being off by one somehow. 

Thanks!

PR #1842, opened 2024-02-07
Embed Metal library source into compiled binary

didzis:
This PR adds option to embed ggml-metal.metal file directly into exectuable.


CMakeLists.txt
```diff
@@ -68,6 +68,7 @@ if (APPLE)
     option(WHISPER_METAL_NDEBUG          "whisper: disable Metal debugging"      OFF)
     option(WHISPER_COREML                "whisper: enable Core ML framework"     OFF)
     option(WHISPER_COREML_ALLOW_FALLBACK "whisper: allow non-CoreML fallback"    OFF)
+    option(WHISPER_EMBED_METAL_LIBRARY   "whisper: embed Metal library"          OFF)
```
ggerganov, line 71: Rename to `WHISPER_METAL_EMBED_LIBRARY`

ggerganov (commented):
Huh, that's a neat trick!

Is there a way to avoid the extra file `ggml-metal-embed.s`?
For example, could it be part of `ggml-metal.m` just surrounded by a specific define like `GGML_METAL_BUILD_ASM`

```c
#ifndef GGML_METAL_BUILD_ASM
... ggml-metal.m source ...
#else
... ggml-metal-embed.s ...
#endif
```

Or something similar - just looking for ways to avoid the extra file

CMakeLists.txt
didzis: It looks like it is not possible to embed this approach into Objective-C source, however if  the aim is to avoid having another file in the source tree, I updated the build process so that the assembly is autogenerated during the build phase.

didzis:
It looks like it is not possible to embed this approach into Objective-C source, however if  the aim is to avoid having another file in the source tree, I updated the build process so that the assembly is autogenerated during the build phase.

ggerganov (approved):

Issue #1855, opened 2024-02-10
Variable audio_ctx consistently produces ~3x speedup for short audio clips


dscripka:
When transcribing short audio clips (e.g., <30 seconds, and usually between 5-10 seconds), I've noticed that the `audio_ctx` parameter can greatly increase performance when set appropriately. After some experimentation, it seems that using the length of the audio clip to scale the value of `audio_ctx` works quite well. I have been using `audio_ctx = (audio length in seconds/30 seconds)*1500 + 128` somewhat arbitrarily.

To confirm these observations I did some comparisons using clips from the Common Voice dataset, the `base.en` model, and this hardware configuration:

CPU: Intel i7-11700K
RAM: DDR4 - 3200 Mhz

`whisper.cpp` build details:

```
whisper_init_from_file_with_params_no_state: loading model from 'models/ggml-base.en.bin'
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51864
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 512
whisper_model_load: n_audio_head  = 8
whisper_model_load: n_audio_layer = 6
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 512
whisper_model_load: n_text_head   = 8
whisper_model_load: n_text_layer  = 6
whisper_model_load: n_mels        = 80
whisper_model_load: ftype         = 1
whisper_model_load: qntvr         = 0
whisper_model_load: type          = 2 (base)
whisper_model_load: adding 1607 extra tokens
whisper_model_load: n_langs       = 99
whisper_model_load:      CPU total size =   147.46 MB (1 buffers)
whisper_model_load: model size    =  147.37 MB
whisper_init_state: kv self size  =   16.52 MB
whisper_init_state: kv cross size =   18.43 MB
whisper_init_state: compute buffer (conv)   =   16.17 MB
whisper_init_state: compute buffer (encode) =   94.42 MB
whisper_init_state: compute buffer (cross)  =    5.08 MB
whisper_init_state: compute buffer (decode) =  105.96 MB

[...]

PR #1857, opened 2024-02-11
added audio_ctx argument to main and server examples

dscripka:
Adds the ability to use the `audio_ctx` argument with the `main` and `server` examples, given that in certain scenarios it may provide a significant performance improvement (#1855).

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -64,6 +64,7 @@ struct whisper_params {
     int32_t max_len      =  0;
     int32_t best_of      = whisper_full_default_params(WHISPER_SAMPLING_GREEDY).greedy.best_of;
     int32_t beam_size    = whisper_full_default_params(WHISPER_SAMPLING_BEAM_SEARCH).beam_search.beam_size;
+    int32_t audio_ctx = 1500;
 
     float word_thold    =  0.01f;
     float entropy_thold =  2.40f;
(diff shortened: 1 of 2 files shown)


examples/main/main.cpp
```diff
@@ -64,6 +64,7 @@ struct whisper_params {
     int32_t max_len      =  0;
     int32_t best_of      = whisper_full_default_params(WHISPER_SAMPLING_GREEDY).greedy.best_of;
     int32_t beam_size    = whisper_full_default_params(WHISPER_SAMPLING_BEAM_SEARCH).beam_search.beam_size;
+    int32_t audio_ctx = 1500;
```
ggerganov, line 67: Default it to `0` - this way `whisper.cpp` will automatically use the correct value:

```suggestion
    int32_t audio_ctx   = 0;
```

ggerganov (approved):

examples/server/server.cpp
```diff
@@ -60,6 +60,7 @@ struct whisper_params {
     int32_t max_len       = 0;
     int32_t best_of       = 2;
     int32_t beam_size     = -1;
+    int32_t audio_ctx     = 1500;
```
ggerganov, line 63: here too 😄 

```suggestion
    int32_t audio_ctx     = 0;
```

ggerganov (approved):

examples/server/server.cpp
dscripka: Ah, right  😄

PR #1859, opened 2024-02-12
whisper : fix usage of extenral encoders (e.g. CoreML)

ggerganov:
```bash
# download base.en
./models/download-ggml-model.sh base.en

# create CoreML model
./models/generate-coreml-model.sh base.en

# build and with CoreML support
WHISPER_COREML=1 make samples
WHISPER_COREML=1 make -j && ./main -m ./models/ggml-base.en.bin -f samples/gb0.wav
```

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -1659,25 +1659,19 @@ static struct ggml_cgraph * whisper_build_graph_conv(
         ggml_set_name(cur, "embd_conv");
         wstate.embd_conv = cur;
     } else {
-#ifdef WHISPER_USE_COREML
+        // keep the "mel" tensor alive - we will use it to store the input data for the external encoders
+        // TODO: is there a better way to do this
+        mel = ggml_scale(ctx0, mel, 1.0f);
+        ggml_build_forward_expand(gf, mel);
+
         cur = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, n_state, n_ctx);
-        ggml_allocr_alloc(alloc, cur);
-
-        if (!ggml_allocr_is_measure(alloc)) {
-            whisper_coreml_encode(wstate.ctx_coreml, mel->ne[0], mel->ne[1], (float *) mel->data, (float *) cur->data);
-        }
-#endif
-#ifdef WHISPER_USE_OPENVINO
-        cur = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, n_state, n_ctx);
-        ggml_allocr_alloc(alloc, cur);
-
-        if (!ggml_allocr_is_measure(alloc)) {
-            whisper_openvino_encode(wstate.ctx_openvino, mel, cur);
-        }
-#endif
-
         ggml_set_name(cur, "embd_enc");
+        ggml_set_output(cur);
         wstate.embd_enc = cur;
+
+        // TODO: without this op, the "embd_enc" tensor ends up being not allocated
+        //       is there a better fix?
+        cur = ggml_scale(ctx0, cur, 1.0f);
     }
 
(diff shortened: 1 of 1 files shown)


whisper.cpp
```diff
@@ -1659,25 +1659,19 @@ static struct ggml_cgraph * whisper_build_graph_conv(
...
-        }
-#endif
-
         ggml_set_name(cur, "embd_enc");
+        ggml_set_output(cur);
         wstate.embd_enc = cur;
+
+        // TODO: without this op, the "embd_enc" tensor ends up being not allocated
+        //       is there a better fix?
+        cur = ggml_scale(ctx0, cur, 1.0f);
```
slaren, line 1674: Only the tensors used in the nodes of the graph are allocated, but I don't think this is new. Did it work before alloc v3?

whisper.cpp
```diff
@@ -1659,25 +1659,19 @@ static struct ggml_cgraph * whisper_build_graph_conv(
         ggml_set_name(cur, "embd_conv");
         wstate.embd_conv = cur;
     } else {
-#ifdef WHISPER_USE_COREML
-        cur = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, n_state, n_ctx);
-        ggml_allocr_alloc(alloc, cur);
+        // keep the "mel" tensor alive - we will use it to store the input data for the external encoders
+        // TODO: is there a better way to do this
+        mel = ggml_scale(ctx0, mel, 1.0f);
+        ggml_build_forward_expand(gf, mel);
```
ggerganov, line 1665: @slaren Do you have suggestions how to improve this? Without using an explicit op (such as `ggml_scale`, the input `mel` and `embd_enc` tensors will not get allocated by the allocator

The alternative is to not use `ggml` tensors at all when using an external encoder, which makes sense overall. But before I reimplement it, was wondering if there is a neat way to improve on this

slaren: As a workaround, you can add this directly to the graph nodes as a `GGML_OP_NONE`, ie `gf->nodes[gf->n_nodes++] = mel`. At least, this will avoid wasting compute on a useless op.

The problem, as mentioned above, is that only the nodes of the graph are considered in `ggml-alloc`. I have thought of making all the leafs in the graph automatically inputs - this would ensure that they are allocated at the beginning of the graph, and if they are never used they will never be freed, so they will also be suitable as an output or static tensor. My reasoning is that the only reason to use `ggml_new_tensor` on a graph is to create an input tensor, since there are better ways to do everything else that in the past required creating new tensors (ie. use `ggml_cont` and `ggml_cast` instead of `ggml_cpy` with a new tensor). But I am worried that it will waste memory in code that is not written very carefully. What do you think?

whisper.cpp
ggerganov: I guess before it worked because of the explicit `ggml_allocr_alloc` calls

whisper.cpp
slaren: Thinking about this more, I am inclined to favor ease of use over a minor memory usage optimization. It can be very surprising that a tensor added to the graph ends not being allocated, and figuring the reason of that requires knowledge of the internals of ggml-alloc. So I think it would be a good idea to simply allocate all the leafs at the beginning of the graphs as if they were inputs. If this causes increased memory usage, it is always possible to optimize this by removing the calls to `ggml_new_tensor` from the graph.

slaren: https://github.com/ggerganov/ggml/pull/731

ggerganov: Will give this a try now

ggerganov:
superseded by https://github.com/ggerganov/whisper.cpp/pull/1860

PR #1865, opened 2024-02-13
talk and talk-llama: Pass text_to_speak as a file

### LINKED ISSUE #1796: win 11 / linux - talk-llama voice interface is broken. Commit d6b9be2 from 19.01.2204
Under windows 11 / linux no audio output
Commit [d6b9be2](https://github.com/ggerganov/whisper.cpp/commit/d6b9be21d76b91a96bb987063b25e5b532140253) from 19.01.2204

````
.\talk-llama.exe -mw ggml-large-v3.bin -ml marcoroni-7b-v3.Q8_0.gguf -s speak.bat -ngl 99 -t 16
````
Variables $1 and $2 from speak / speak.bat are not passing proper strings 
For instance I changed under windows  speak.ps1 to
````
# Set-ExecutionPolicy -ExecutionPolicy Bypass -Scope CurrentUser
param(
  # voice options are David or Zira
  [Parameter(Mandatory=$true)][string]$voice,
  [Parameter(Mandatory=$true)][string]$text
)

# testing variables
Add-Content -Path C:\AI_music\output.txt -Value $voice
Add-Content -Path C:\AI_music\output.txt -Value $text

Add-Type -AssemblyName System.Speech;
$speak = New-Object System.Speech.Synthesis.SpeechSynthesizer;
$speak.SelectVoice("Microsoft $voice Desktop");
$speak.Rate="0";
$speak.Speak($text);

````

to get variables $1 and $2 to a file

Output the output.txt during talking with talk-llama
````
2
'
2
'
2
'
2
'
2
'
````

Conclusion - audio interface is broken.

### PR DESCRIPTION
**Pull request by tamo**
1. Fix #1796
2. Make speak.sh work without modification in more cases
3. Enable eleven-labs.py to play audio
4. Implement as much information in comments as possible

(Drop some commits if you think I made too many changes in one PR.)

1:
Currently talk and talk-llama pass text_to_speak as an argument to TTS in system() call.
That's a little scary.

True, we can escape dangerous characters confidently if we can focus on bash.
But we have to take care of Windows, whose pwsh once changed the way of handling quotes.
I'm not sure that what is safe today will be safe tomorrow.

So I suggest we output text_to_speak to a file, and then pass it to TTS.

Pipe would be better, but I don't know how to implement piping in a portable way (and I guess we prefer avoiding #ifdefs in cpp).

2:
Testing a number of TTS's is hard because we have to edit speak.sh every time.
Simple "export PATH=somewhere" would be better.
So I suggest we check "command -v" in speak.sh.
You can change the order if you want. (For example, to increase the priority of "say")

3:
The current example of the usage of elevenlabs in speak.sh is, save audio.mp3 and ffplay it.
But elevenlabs itself plays the audio with ffplay if we call "play" instead of "save".
https://github.com/elevenlabs/elevenlabs-python/blob/v0.2.27/elevenlabs/utils.py

Also, when I tested "save", ffplay miscalculated the audio duration. So I think "play" is more stable.

4:
Comments have much useful information, such as how to install TTS programs and voice options.
It would be better to code it than keep it hidden in comments.

### DIFF
--- a/examples/talk-llama/talk-llama.cpp
+++ b/examples/talk-llama/talk-llama.cpp
@@ -75,6 +75,7 @@ struct whisper_params {
     std::string model_wsp   = "models/ggml-base.en.bin";
     std::string model_llama = "models/ggml-llama-7B.bin";
     std::string speak       = "./examples/talk-llama/speak";
+    std::string speak_file  = "./examples/talk-llama/to_speak.txt";
     std::string prompt      = "";
     std::string fname_out;
     std::string path_session = "";       // path to file for saving/loading model eval state
@@ -113,6 +114,7 @@ bool whisper_params_parse(int argc, char ** argv, whisper_params & params) {
         else if (arg == "-mw"  || arg == "--model-whisper")  { params.model_wsp      = argv[++i]; }
         else if (arg == "-ml"  || arg == "--model-llama")    { params.model_llama    = argv[++i]; }
         else if (arg == "-s"   || arg == "--speak")          { params.speak          = argv[++i]; }
+        else if (arg == "-sf"  || arg == "--speak-file")     { params.speak_file     = argv[++i]; }
         else if (arg == "--prompt-file")                     {
             std::ifstream file(argv[++i]);
             std::copy(std::istreambuf_iterator<char>(file), std::istreambuf_iterator<char>(), back_inserter(params.prompt));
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Review by bobqianic — APPROVED** [review]

File: examples/talk-llama/talk-llama.cpp [hunk 1]
```diff
@@ -546,9 +549,16 @@ int main(int argc, char ** argv) {
...
+                    std::ofstream speak_file(params.speak_file.c_str());
+                    if (speak_file.fail()) {
+                        fprintf(stderr, "%s: failed to open speak_file\n", __func__);
+                    } else {
+                        speak_file.write(params.heard_ok.c_str(), params.heard_ok.size());
+                        speak_file.close();
+                        int ret = system((params.speak + " " + std::to_string(voice_id) + " " + params.speak_file).c_str());
+                        if (ret != 0) {
+                            fprintf(stderr, "%s: failed to speak\n", __func__);
+                        }
```
ggerganov, line 561: Let's factor out this as a helper function `speak_with_file()` in `common` and reuse it

ggerganov (approved):

PR #1872, opened 2024-02-16
main : check if input files exist before proceeding

Theldus:
Until the most recent commit (3d42463), the main.cpp sample file does not check whether the input files exist or not. Consequently, the model is loaded first before reporting whether there was a failure or not when processing a file. In environments with HDD, this can take about 50 seconds or more, depending on the loaded model.

This PR addresses this by checking in advance whether the input files exist or not.

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -10,6 +10,8 @@
 #include <vector>
 #include <cstring>
 
+#include <sys/stat.h>
+
 #if defined(_MSC_VER)
 #pragma warning(disable: 4244 4267) // possible loss of data
 #endif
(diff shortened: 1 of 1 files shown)


bobqianic:
A more graceful approach might be to first load the file and then the model. This way, we can leverage `dr_wav` to verify the existence of the file.

Theldus:
@bobqianic Yes, I do agree with you, but I thought about not doing that in this PR, since main.cpp allows an arbitrary number of wav files, and the user could even run out of memory* before loading the model.

Although perhaps this might not be exactly an issue for whisper.cpp users, I'm not sure.

(*From what I understood, in the current code, each wav is loaded and then processed, one at a time.)

ggerganov (approved):

mcdallas:
This needs to account for reading from stdin with `-`

Theldus:
@mcdallas You're right; I apologize for introducing this bug in my last PR. I did not consider the stdin case.

Although simple, I have created a new PR (#1889) addressing this issue.

PR #1897, opened 2024-02-24
Correct syscall name SYS_get_cpu used with old glibc where getcpu() is missing.

### LINKED ISSUE #1894: ggml.c: error: 'SYS_getcpu' undeclared did you mean 'SYS_get_cpu'
On both projects, **whisper.cpp**, and **llama.cpp**, we get this compilation error with latest git when compiling with CUBLAS.

 `WHISPER_CUBLAS=1 make -j`

```
ggml.c: In function 'ggml_numa_init':
ggml.c:2088:26: error: 'SYS_getcpu' undeclared (first use in this function); did you mean 'SYS_get_cpu'?
 2088 |     getcpu_ret = syscall(SYS_getcpu,&current_cpu,&g_state.numa.current_node);
      |                          ^~~~~~~~~~
      |                          SYS_get_cpu
ggml.c:2088:26: note: each undeclared identifier is reported only once for each function it appears in
make: *** [Makefile:317: ggml.o] Error 1
```

The fix is very straightforward. We simply drop in `SYS_get_cpu`.

petterreinholdtsen:
Fixes #1894.

### DIFF
--- a/ggml.c
+++ b/ggml.c
@@ -2084,8 +2084,17 @@ void ggml_numa_init(enum ggml_numa_strategy numa_flag) {
 #if __GLIBC__ > 2 || (__GLIBC__ == 2 && __GLIBC_MINOR__ > 28)
     getcpu_ret = getcpu(&current_cpu, &g_state.numa.current_node);
 #else
-    // old glibc doesn't have a wrapper for this call. Fall back on direct syscall
-    getcpu_ret = syscall(SYS_getcpu,&current_cpu,&g_state.numa.current_node);
+    // old glibc doesn't have a wrapper for this call. Fall back on
+    // direct syscall
+    getcpu_ret = syscall(
+#  if defined(SYS_getcpu)
+			 SYS_getcpu,
+#  elif defined(SYS_get_cpu)
+			 SYS_get_cpu,
+#  else
+#    error "Unable fo find getcpu syscall define"
+#  endif /* SYS_getcpu */
+			 &current_cpu,&g_state.numa.current_node);
 #endif
 
     if (g_state.numa.n_nodes < 1 || g_state.numa.total_cpus < 1 || getcpu_ret != 0) {


ggml.c
```diff
@@ -2084,8 +2084,17 @@ void ggml_numa_init(enum ggml_numa_strategy numa_flag) {
...
+    // direct syscall
+    getcpu_ret = syscall(
+#  if defined(SYS_getcpu)
+			 SYS_getcpu,
+#  elif defined(SYS_get_cpu)
+			 SYS_get_cpu,
+#  else
+#    error "Unable fo find getcpu syscall define"
+#  endif /* SYS_getcpu */
+			 &current_cpu,&g_state.numa.current_node);
```
cebtenzzre, line 2097: I think something like this would be more readable:
```suggestion
#   if !defined(SYS_getcpu) && defined(SYS_get_cpu)
#       define SYS_getcpu SYS_get_cpu
#   endif
    // old glibc doesn't have a wrapper for this call. Fall back on direct syscall
    getcpu_ret = syscall(SYS_getcpu, &current_cpu, &g_state.numa.current_node);
```

bmtwl: @cebtenzzre that looks like an elegant solution.
@petterreinholdtsen are you able to test cebtenzzre's proposed fix and confirm that it works in your case?

petterreinholdtsen:
[bmwl]
> @cebtenzzre that looks like an elegant solution.
> @petterreinholdtsen are you able to test cebtenzzre's proposed fix and
> confirm that it works in your case?

I did not really see the point to test it, as it seemed obvious that his
proposal would work too.  The only difference is how the two proposals
will fail if both of the two defines are missing.  His proposal will
report the unknown symbol, while mine will report the message in the
#error line.

In any case, because you asked and it only take a few minutes, I did a
test build on Debian 12 Bookworm of his proposal and can confirm that it
build just fine here.

-- 
Happy hacking
Petter Reinholdtsen

bmtwl:
> I did not really see the point to test it, as it seemed obvious that his proposal would work too. The only difference is how the two proposals will fail if both of the two defines are missing. His proposal will report the unknown symbol, while mine will report the message in the #error line. In any case, because you asked and it only take a few minutes, I did a test build on Debian 12 Bookworm of his proposal and can confirm that it build just fine here.
> […](#)
> -- Happy hacking Petter Reinholdtsen

True enough. I suppose what we really need is @themayone to test to make sure the regression on his OS is fixed with this patch as its only very old Linux kernels that would be hit with this issue.

bobqianic:
The issue has been resolved with the changes made in https://github.com/ggerganov/llama.cpp/pull/5906
Therefore, this pull request will now be closed.

PR #1902, opened 2024-02-24
Stream.wasm: Fix invalid memory access when no segments are returned

Andrews54757:
No segments may be returned when a smaller sample buffer (EG 2048 samples) is sent to the worker. This causes an invalid memory access error as `whisper_full_get_segment_text(ctx, -1)` is called.

### DIFF
--- a/examples/stream.wasm/emscripten.cpp
+++ b/examples/stream.wasm/emscripten.cpp
@@ -103,11 +103,11 @@ void stream_main(size_t index) {
 
             {
                 const int n_segments = whisper_full_n_segments(ctx);
-                for (int i = n_segments - 1; i < n_segments; ++i) {
-                    const char * text = whisper_full_get_segment_text(ctx, i);
+                if (n_segments > 0) {
+                    const char * text = whisper_full_get_segment_text(ctx, n_segments - 1);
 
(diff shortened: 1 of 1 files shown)


examples/stream.wasm/emscripten.cpp
```diff
@@ -103,11 +103,11 @@ void stream_main(size_t index) {
 
             {
                 const int n_segments = whisper_full_n_segments(ctx);
-                for (int i = n_segments - 1; i < n_segments; ++i) {
-                    const char * text = whisper_full_get_segment_text(ctx, i);
+                if (n_segments > 0) {
```
ggerganov, line 106: This will only process the last segment. We want to process all segments

```suggestion
                for (int i = n_segments - 1; i < n_segments && n_segments > 0; ++i) {
```

Andrews54757: `single_segment` is set to true so it will not output more than one segment. Furthermore, the original code's `for` loop will only iterate once. Because it starts from `n-1`, it will only process the last segment.

ggerganov: Oh, I missed that 👍

ggerganov (approved):

PR #1932, opened 2024-03-06
Add build scripts for bench.cpp to generate bench tool for Android-based device

zhouwg:


zhouwg:
not work on my Xiaomi 14. I'll investigate why?  so close this PR accordingly.

PR #1942, opened 2024-03-08
whisper : document whisper_batch.n_seq_id

josharian:
It is unused and has been since its introduction.

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -413,7 +413,7 @@ struct whisper_batch {
 
     whisper_token  *  token;
     whisper_pos    *  pos;
-    int32_t        *  n_seq_id;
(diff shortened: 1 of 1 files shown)


ggerganov:
This is mostly for consistency with `llama_batch` from the `llama.cpp` project. Prefer to keep it around - it might become useful at some point if we decide to implement some advanced speculative decoding techniques

josharian:
Ack. I'll update this PR to add a few words of documentation instead.

josharian:
> Ack. I'll update this PR to add a few words of documentation instead.

Updated.

Given that `n_seq_id` exists, `seq_id` probably doesn't need to be null-terminated. But I'll resist the temptation to make further gratuitous code changes. :)

ggerganov (approved):

PR #1964, opened 2024-03-16
whisper : allocate encoder results in dedicated buffer

### LINKED ISSUE #1959: whisper : adapt to latest ggml changes
This view in the "encoder" graph of a tensor from the "conv" graph is triggering an assert with the latest build:

https://github.com/ggerganov/whisper.cpp/blob/de4d067f1e162527044498a7ec23a9ee458ac280/whisper.cpp#L1711

The assert is:

https://github.com/ggerganov/whisper.cpp/blob/de4d067f1e162527044498a7ec23a9ee458ac280/ggml-alloc.c#L737-L745

Is this a valid use case and we simply have to remove the assert, or a proper fix is necessary?

cc @slaren

ggerganov:
fix #1959 

Allocate the encoder results tensors (`embd_conv` and `embd_enc`) in dedicated backend buffer. Copy the results into these tensors and used them in follow-up graphs

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -793,6 +793,9 @@ struct whisper_state {
     struct ggml_tensor * embd_conv = nullptr;
     struct ggml_tensor * embd_enc  = nullptr;
 
+    ggml_context * ctx_embd = nullptr;
+    ggml_backend_buffer_t buffer_embd = nullptr;
+
     // helpers for GPU offloading
     std::vector<float> inp_mel;
     std::vector<float> inp_mask;
@@ -1669,15 +1672,9 @@ static struct ggml_cgraph * whisper_build_graph_conv(
             cur = ggml_gelu(ctx0, cur);
         }
 
-        ggml_set_name(cur, "embd_conv");
-        wstate.embd_conv = cur;
+        cur = ggml_cpy(ctx0, cur, wstate.embd_conv);
     } else {
         ggml_build_forward_expand(gf, mel);
-
-        cur = ggml_new_tensor_2d(ctx0, GGML_TYPE_F32, n_state, n_ctx);
-
-        ggml_set_name(cur, "embd_enc");
-        wstate.embd_enc = cur;
     }
 
     ggml_build_forward_expand(gf, cur);
@@ -1708,7 +1705,10 @@ static struct ggml_cgraph * whisper_build_graph_encoder(
 
     ggml_cgraph * gf = ggml_new_graph_custom(ctx0, WHISPER_MAX_NODES, false);
 
-    struct ggml_tensor * cur = ggml_view_tensor(ctx0, wstate.embd_conv);
+    // TODO: this still triggers the assert:
+    //struct ggml_tensor * cur = ggml_view_tensor(ctx0, wstate.embd_conv);
+
+    struct ggml_tensor * cur = wstate.embd_conv;
 
     const float KQscale = 1.0f/sqrtf(float(n_state)/n_head);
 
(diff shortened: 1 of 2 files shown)


whisper.cpp
```diff
@@ -1708,7 +1705,10 @@ static struct ggml_cgraph * whisper_build_graph_encoder(
 
     ggml_cgraph * gf = ggml_new_graph_custom(ctx0, WHISPER_MAX_NODES, false);
 
-    struct ggml_tensor * cur = ggml_view_tensor(ctx0, wstate.embd_conv);
+    // TODO: this still triggers the assert:
+    //struct ggml_tensor * cur = ggml_view_tensor(ctx0, wstate.embd_conv);
```
ggerganov, line 1709: With the `wstate.embd_conv` now pre-allocated, this view still triggers the assert:

```
Assertion failed: (tensor_alloc->offset == SIZE_MAX), function ggml_gallocr_init_tensor, file ggml-alloc.c, line 739.
```

Not sure why

slaren: The code that stores the allocation location of leafs didn't take into account pre-allocated leafs. This should fix it:
```diff
diff --git a/ggml-alloc.c b/ggml-alloc.c
index 8ac1d3e..60b86c2 100644
--- a/ggml-alloc.c
+++ b/ggml-alloc.c
@@ -701,8 +701,13 @@ bool ggml_gallocr_reserve_n(ggml_gallocr_t galloc, struct ggml_cgraph * graph, c
         struct ggml_tensor * leaf = graph->leafs[i];
         struct hash_node * hn = ggml_gallocr_hash_get(galloc, leaf);
         galloc->leaf_allocs[i].buffer_id = hn->buffer_id;
-        galloc->leaf_allocs[i].leaf.offset = hn->offset;
-        galloc->leaf_allocs[i].leaf.size_max = ggml_backend_buft_get_alloc_size(galloc->bufts[hn->buffer_id], leaf);
+        if (leaf->view_src || leaf->data) {
+            galloc->leaf_allocs[i].leaf.offset = SIZE_MAX;
+            galloc->leaf_allocs[i].leaf.size_max = 0;
+        } else {
+            galloc->leaf_allocs[i].leaf.offset = hn->offset;
+            galloc->leaf_allocs[i].leaf.size_max = ggml_backend_buft_get_alloc_size(galloc->bufts[hn->buffer_id], leaf);
+        }
     }

     // reallocate buffers if needed
```

ggerganov: With just this change applied on `master` the code also works (i.e. without pre-allocating the tensors). But if I understand correctly, it is technically not correct because nothing guarantees that the data in these tensors would not be overwritten by some ops. Since they are currently at the end of the computation graphs it seems to produce correct results, but the concern is that this is not very future-proof if we expand the graphs in the future - is that correct?

What if I use `ggml_set_output(ctx0, wstate.embd_conv);` in `whisper_build_graph_conv()` to guarantee that the data would not be overwritten?

slaren: Yes, with `ggml_set_output` it would guarantee that the tensor is never overwritten.

ggerganov: Would pre-allocating the tensors as proposed in this PR have any advantage over the `ggml_set_output` option? Since with this PR we now have to perform extra copy of the data between the graph calls, which otherwise is not needed.

slaren: I don't think there would be any significant advantage. Maybe it would using `ggml_gallocr_reserve` here instead of `ggml_gallocr_alloc_graph`, but the difference would be minimal:

https://github.com/ggerganov/whisper.cpp/blob/66df44b0b7f2b1671e59379ad019df3b71b2547c/whisper.cpp#L496-L502

ggerganov:
Picked the `ggml-alloc` commit directly on `master`

PR #1969, opened 2024-03-17
fedora dependencies needed

Man2Dev:
worte some documentation on for dependencies needed for builing on fedora linux.


ulatekh:
You may want to include how to build GCC 12, which is needed for CUDA compilation under Fedora Core. Find full instructions [here](https://stackoverflow.com/a/61434449/603828).

Man2Dev:
thanks I will take look

Issue #1962, opened 2024-03-16
How to calculate token size from string?


sindresorhus:
I'm using the `params.initial_prompt` setting in my app, and I would like to interactively warn the user if they exceed the prompt limit. Given a string, how can I calculate the token size?

felrock:
You could use `static std::vector<whisper_vocab::id> tokenize(const whisper_vocab & vocab, const std::string & text)` on your initial_prompt and check the length, compare it to your context size.

sindresorhus:
I'm using Whisper.cpp's C API from Swift. But I now see that I could use `whisper_tokenize`. However, it's a bit of a waste to fill up the tokens array when I only care about the size.

So my request is to add something like this:

```c++
int whisper_token_count(struct whisper_context * ctx, const char * text) {
    const auto res = tokenize(ctx->vocab, text);
    return res.size();
}
```

felrock:
Your request is a bit too specific imo. You can create a function like that yourself, just disregard the tokens after you've checked the size.

sindresorhus:
Any app accepting a custom prompt will need this. Of course every app could reinvent the wheel, but I think many app developers don't even realize they have to validate the prompt. I for sure did not. Having a specific function like this makes it more discoverable that they need to.

Issue #1961, opened 2024-03-16
Setting `params.initial_prompt` to too many tokens should make `whisper_full` fail instead of throwing an exception


sindresorhus:
I cannot really catch exceptions in Swift.

There could be some check here instead and then return a status code:

https://github.com/ggerganov/whisper.cpp/blob/de4d067f1e162527044498a7ec23a9ee458ac280/whisper.cpp#L5080

Currently, it throws an exception at this line.

PR #1981, opened 2024-03-21
whisper : improve handling of prompts

### LINKED ISSUE #1960: Document max tokens for `params.initial_prompt`
It's limited to 1024 tokens, but that's not documented anywhere. It should be documented in the API and in the `--prompt` CLI help output.

ggerganov:
fix #1960, #1961, #1962

Changed `whisper_tokenize()` to return the negative number of required tokens when the token buffer is not big enough. This can be used to determine the necessary tokens for a given text:

```c
int n_needed = -whisper_tokenize(ctx, "some text", NULL, 0);
```

Also improve docs by clarifying that the Whisper models can process prompts of up to only `n_text_ctx/2` tokens which is 224 tokens. So there is no point to provide longer prompts as they would be truncated:

https://github.com/ggerganov/whisper.cpp/blob/48a145207ef08d6be1be98121d1034b024f521fd/whisper.cpp#L5474-L5480

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -207,7 +207,7 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
     fprintf(stderr, "  -nt,       --no-timestamps     [%-7s] do not print timestamps\n",                        params.no_timestamps ? "true" : "false");
     fprintf(stderr, "  -l LANG,   --language LANG     [%-7s] spoken language ('auto' for auto-detect)\n",       params.language.c_str());
     fprintf(stderr, "  -dl,       --detect-language   [%-7s] exit after automatically detecting language\n",    params.detect_language ? "true" : "false");
-    fprintf(stderr, "             --prompt PROMPT     [%-7s] initial prompt\n",                                 params.prompt.c_str());
+    fprintf(stderr, "             --prompt PROMPT     [%-7s] initial prompt (max n_text_ctx/2 tokens)\n",       params.prompt.c_str());
     fprintf(stderr, "  -m FNAME,  --model FNAME       [%-7s] model path\n",                                     params.model.c_str());
     fprintf(stderr, "  -f FNAME,  --file FNAME        [%-7s] input WAV file path\n",                            "");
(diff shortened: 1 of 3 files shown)


ggerganov:
@sindresorhus lmk if this change would work for you

sindresorhus:
Thanks for looking into this. That would work, although I still think wrapping it up into a separate function would make for a better API and make it more discoverable for consumers. I would never have thought of using the `whisper_tokenize` like that.

```c++
// Some docs
int whisper_token_count(struct whisper_context * ctx, const char * text) {
   return -whisper_tokenize(ctx, text, NULL, 0);
}
```

sindresorhus:
> fix https://github.com/ggerganov/whisper.cpp/issues/1960, https://github.com/ggerganov/whisper.cpp/issues/1961, https://github.com/ggerganov/whisper.cpp/issues/1962

By the way, you need to make that:

> fix https://github.com/ggerganov/whisper.cpp/issues/1960, fix https://github.com/ggerganov/whisper.cpp/issues/1961, fix https://github.com/ggerganov/whisper.cpp/issues/1962

For GitHub to do the right thing and close them all when this PR is merged.

examples/main/main.cpp
```diff
@@ -207,7 +207,7 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
     fprintf(stderr, "  -nt,       --no-timestamps     [%-7s] do not print timestamps\n",                        params.no_timestamps ? "true" : "false");
     fprintf(stderr, "  -l LANG,   --language LANG     [%-7s] spoken language ('auto' for auto-detect)\n",       params.language.c_str());
     fprintf(stderr, "  -dl,       --detect-language   [%-7s] exit after automatically detecting language\n",    params.detect_language ? "true" : "false");
-    fprintf(stderr, "             --prompt PROMPT     [%-7s] initial prompt\n",                                 params.prompt.c_str());
+    fprintf(stderr, "             --prompt PROMPT     [%-7s] initial prompt (max n_text_ctx/2 tokens)\n",       params.prompt.c_str());
```
sindresorhus, line 210: I would document the 224 number here for quick reference.

Same in https://github.com/ggerganov/whisper.cpp/blob/5c2c07d479163b39fe7337c0c2976bb4834a7cd3/whisper.h#L507

PR #1990, opened 2024-03-24
Swift: Updating Swift package for linux configuration

Sombre-Osmoze:
- We need to ignore Apple platforms frameworks (Accelerate and Metal) when building the project on Linux
- We need to define `_GNU_SOURCE` for certains function declaration (in `sched.h`)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -2,6 +2,46 @@
 
 import PackageDescription
 
+var sources = [
+  "ggml.c",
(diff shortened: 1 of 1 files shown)


sindresorhus:
The Package.swift here closely follows https://github.com/ggerganov/llama.cpp/blob/master/Package.swift, so this should be done there first.

PR #2004, opened 2024-03-27
Create solution folders in the CMake build

ulatekh:


ggerganov:
What is this needed for?

ulatekh:
To make the Visual Studio solution more organized and easier to navigate.
This will also work for Qt Creator, and presumably, whatever Macintosh uses.

PR #2005, opened 2024-03-27
Added a bool fold_lowercase to whisper_context_params

ulatekh:
If true, it folds language-model tokens to lowercase. By default, it's false.
This is intended to make grammar matching more predictable, e.g. no need to account for case in the grammar.

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -1383,6 +1383,11 @@ static bool whisper_model_load(struct whisper_model_loader * loader, whisper_con
                 word = "";
             }
 
+            // If requested, output all text as lowercase.
+            if (wctx.params.fold_lowercase)
+                std::transform(word.begin(), word.end(), word.begin(),
+                    [](unsigned char c) { return std::tolower(c); });
+
             vocab.token_to_id[word] = i;
             vocab.id_to_token[i] = word;
 
@@ -3374,7 +3379,7 @@ struct whisper_context_params whisper_context_default_params() {
     struct whisper_context_params result = {
         /*.use_gpu              =*/ true,
         /*.gpu_device           =*/ 0,
-
+        /*.fold_lowercase       =*/ false,
         /*.dtw_token_timestamps =*/ false,
         /*.dtw_aheads_preset    =*/ WHISPER_AHEADS_NONE,
         /*.dtw_n_top            =*/ -1,
--- a/whisper.h
+++ b/whisper.h
@@ -115,6 +115,9 @@ extern "C" {
         bool  use_gpu;
         int   gpu_device;  // CUDA device
 
+        // Fold language tokens to lowercase
+        bool fold_lowercase;
+
         // [EXPERIMENTAL] Token-level timestamps with DTW
         bool dtw_token_timestamps;
         enum whisper_alignment_heads_preset dtw_aheads_preset;


examples/command/command.cpp
```diff
@@ -44,6 +44,7 @@ struct whisper_params {
     bool print_energy  = false;
     bool no_timestamps = true;
     bool use_gpu       = true;
+    bool model_fold_lc = false;
```
ggerganov, line 47: Change name to `vocab_lc`

whisper.h
```diff
@@ -115,6 +115,9 @@ extern "C" {
         bool  use_gpu;
         int   gpu_device;  // CUDA device
 
+        // Fold language tokens to lowercase
+        bool fold_lowercase;
```
ggerganov, line 119:
```suggestion
        bool vocab_lc;
```

whisper.cpp
```diff
@@ -1383,6 +1383,11 @@ static bool whisper_model_load(struct whisper_model_loader * loader, whisper_con
                 word = "";
             }
 
+            // If requested, output all text as lowercase.
+            if (wctx.params.fold_lowercase)
+                std::transform(word.begin(), word.end(), word.begin(),
+                    [](unsigned char c) { return std::tolower(c); });
+
```
ggerganov, line 1390:
```suggestion
            // If requested, output all text as lowercase.
            if (wctx.params.fold_lowercase) {
                std::transform(word.begin(), word.end(), word.begin(),
                    [](unsigned char c) { return std::tolower(c); });
            }

```

ggerganov (approved):

examples/main/main.cpp
```diff
@@ -205,6 +207,7 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
     fprintf(stderr, "  -pc,       --print-colors      [%-7s] print colors\n",                                   params.print_colors ? "true" : "false");
     fprintf(stderr, "  -pp,       --print-progress    [%-7s] print progress\n",                                 params.print_progress ? "true" : "false");
     fprintf(stderr, "  -nt,       --no-timestamps     [%-7s] do not print timestamps\n",                        params.no_timestamps ? "true" : "false");
+    fprintf(stderr, "  --model-fold-lc                [%-7s] fold all model tokens to lowercase\n",             params.model_fold_lc ? "true" : "false");
```
ggerganov, line 210:
```suggestion
    fprintf(stderr, "  -nt,       --no-timestamps     [%-7s] do not print timestamps\n",                        params.no_timestamps ? "true" : "false");
    fprintf(stderr, "  --vocab-lc                     [%-7s] fold all vocab tokens to lowercase\n",             params.vocab_lc ? "true" : "false");
```

ulatekh:
I have no idea what's wrong with the Java bindings. I loaded them all into Visual Studio Code and fixed all the errors it reported (which didn't seem related to my changes), but still the Java-related tests fail. FYI, I haven't programmed in Java in over 10 years.

ggerganov:
I'm also not good with Java, but I think we are probably observing an issue similar to this one: https://github.com/ggerganov/llama.cpp/pull/1902#issuecomment-1605524391

In short, even though the two structs `whisper_context_params` (C) and `WhisperContextParams` (Java) have the same members, there are likely different paddings between the members which is causing UB in the following code:

https://github.com/ggerganov/whisper.cpp/blob/8f253ef3af1c62c04316ba4afa7145fc4d701a8c/bindings/java/src/main/java/io/github/ggerganov/whispercpp/WhisperCpp.java#L70-L81

The proper solution is to order the members in decreasing size (i.e. keep the bools at the end of the struct). Or maybe avoid `bool` and simply replace them with `int` - this seems much easier change

PR #2043, opened 2024-04-12
build : detect AVX512 in Makefile, add AVX512 option in CMake

didzis:
Adds missing AVX512 detection to Makefile and CMakeLists.txt


CMakeLists.txt
```diff
@@ -59,10 +59,11 @@ option(WHISPER_BUILD_EXAMPLES         "whisper: build examples" ${WHISPER_STANDA
 
 option(WHISPER_SDL2                   "whisper: support for libSDL2" OFF)
 
-option(WHISPER_NO_AVX                 "whisper: disable AVX"  OFF)
-option(WHISPER_NO_AVX2                "whisper: disable AVX2" OFF)
-option(WHISPER_NO_FMA                 "whisper: disable FMA"  OFF)
-option(WHISPER_NO_F16C                "whisper: disable F16c" OFF)
+option(WHISPER_NO_AVX                 "whisper: disable AVX"    OFF)
+option(WHISPER_NO_AVX2                "whisper: disable AVX2"   OFF)
+option(WHISPER_AVX512                 "whisper: enable AVX512"  OFF)
```
przemoc, line 64: I suggest using `WHISPER_NO_AVX512` to be consistent with rest of these settings (until they will be overhauled one day in whisper.cpp to use positive notion instead of negative one), and naturally changing default to `ON` in such case.

Sorry if my rant in side note made you misunderstand the expectation for this PR.

didzis:
Updated PR so that for CMake AVX512 must be enabled explicitly. Also added AVX512 subsets for this case as suggested, however I would argue that this will lead to "bad instruction" error during execution in case a particular subset is used by the code, but not supported by the target CPU. Although in practice this may not happen because it looks like the inference code does not use any other AVX512 instruction subsets than AVX512F.
Updated Makefile to detect AVX512 subsets.

Makefile
```diff
@@ -144,6 +144,42 @@ ifeq ($(UNAME_M),$(filter $(UNAME_M),x86_64 i686 amd64))
...
+		ifneq (,$(AVX512BW_M))
+			CFLAGS   += -mavx512bw
+			CXXFLAGS += -mavx512bw
+		endif
+
+		AVX512VL_M := $(shell $(CPUINFO_CMD) | grep -iw 'AVX512VL')
+		ifneq (,$(AVX512VL_M))
+			CFLAGS   += -mavx512vl
+			CXXFLAGS += -mavx512vl
+		endif
```
przemoc, line 175: I suggest combining them into one block, where we check just for AVX512F, but then add flags related to all 5 AVX-512 subsets (F, CD, VL, DQ, BW). I don't believe Xeon Phi existence is good enough reason check all of these subsets individually. I never suggested checking them individually.

Table at https://en.wikipedia.org/wiki/AVX-512#CPUs_with_AVX-512 explains why.

Let's have a simpler code. If there will be mainstream CPU in the future not fitting into the assumption (F presence means presence of CD, VL, DQ and BW too), then we can start playing again with checking those flags individually.

Makefile
```diff
@@ -144,6 +144,42 @@ ifeq ($(UNAME_M),$(filter $(UNAME_M),x86_64 i686 amd64))
...
+
+		AVX512VNNI_M := $(shell $(CPUINFO_CMD) | grep -iwE 'AVX512_VNNI|AVX512VNNI')
```
przemoc, line 177: AVX512_VNNI is for Linux, so just out of curiosity:
Which OS uses AVX512VNNI?

As you're adding VNNI beside basic AVX-512 subsets, I would suggest adding also VBMI, as llama.cpp does it too. It's nice to have some kind of parity between these projects.

Moreover, if you're adding VNNI and VBMI in Makefile, you should add them in CMakeLists.txt too, via `WHISPER_NO_AVX512_VNNI` and `WHISPER_NO_AVX512_VBMI`, defaulting to `ON`. I suggest using `add_compile_definitions` similarly like llama.cpp does.

CMakeLists.txt
didzis: Are you suggesting to use `WHISPER_NO_AVX512` set to `ON` by default?

przemoc: Yes, to be consistent, because all other ISA extension options in whisper.cpp have such polarity.
<sub>(Maybe we'll change it in the future, but if it will happen, then it should be done separately.)</sub>

Makefile
didzis: Added AVX512VNNI for pure symmetry just in case any system does have it.

przemoc: I asked about AVX512VNNI, didn't request removing. But I see you removed it in dc69460b42.

So I looked around and it seems AVX512VNNI is the name reported by FreeBSD.
http://fxr.watson.org/fxr/source/x86/x86/identcpu.c#L994

Please bring back `grep -iwE 'AVX512_VNNI|AVX512VNNI'` for VNNI detection.

CMakeLists.txt
```diff
@@ -464,16 +467,31 @@ else()
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /utf-8")
         set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /utf-8")
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /utf-8")
-        if(NOT WHISPER_NO_AVX2)
+        if(NOT WHISPER_NO_AVX512)
+            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /arch:AVX512")
+            set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /arch:AVX512")
+            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /arch:AVX512")
+            # from llama.cpp:
```
przemoc, line 474: I woud suggest removing `from llama.cpp` comment line. There are many things shared between llama.cpp and whisper.cpp, in both directions, so if we would be marking all of them, it would be pure noise. :)

przemoc:
So fixing VNNI detection for non-Linux and removing superfluous comment are the only remaining things.
Rest looks good to me. (But I haven't tested locally building and running whisper.cpp w/ AVX512 enabled.)

CMakeLists.txt
didzis: Sure, just wanted to note, that I'm not author of that comment.

Makefile
didzis: Great, my feeling was right then.

didzis:
Tested inference on a Linux machine with AVX512 all five instruction subsets and AVX512 VNNI.

przemoc (approved):

przemoc:
One little thing. I would suggest changing PR title for two reasons:

- it's partially wrong, because there is no x86 ISA extensions detection in CMake,
- it doesn't follow style used in this project.

I would suggest following one:

> build : support enabling use of AVX-512 (detected in make, via WHISPER_NO_AVX512=0 for CMake)

ggerganov (approved):
@didzis Thank you for the contribution and @przemoc thank you for the thorough review!

PR #2054, opened 2024-04-15
ggml: add hw_accel in data structure

zhouwg:
this PR is for multi purpose:
(1) try to fix issue https://github.com/ggerganov/ggml/issues/795

(2) borrow some advantages from PyTorch(the user could specify whether a GGML OP(such as mulmat) is accelerated by a specify backend) 

(3) prepare for submit Qualcomm's QNN backend to upstream GGML from "PoC: Add Qualcomm mobile SoC native backend for GGML,https://github.com/zhouwg/kantv/issues/121 ". whisper.cpp at the first, then llama.cpp, because llama.cpp is much more complicated then whisper.cpp

pls refer to this commit:https://github.com/zhouwg/kantv/commit/ce44da6cf14c0569c8d9e1cbf66b4cc6228ab344

or 

pls refer to this commit: https://github.com/zhouwg/kantv/blob/kantv-poc-with-qnn/core/ggml/llamacpp/ggml.c#L16137

this is a workaround(it breaks OO principle in original GGML) solution/method for this TODO in original ggml: https://github.com/ggerganov/ggml/blob/master/src/ggml-backend.c#L1127

I personally think this member is not redundant(it's NOT same to existing "backend" in "struct ggml_tensor") and it will NOT bring side-effect to existing codes. of course, I understand we should not bring too much "useful codes" into existing implementation of GGML internal and we should keep GGML as compact/clean as possible.

<hr>

updated on 04-17-2024, not essential, there is another better method(workaround) for this TODO in original ggml: https://github.com/ggerganov/ggml/blob/master/src/ggml-backend.c#L1127.

so I'd like to close this PR accordingly.

### DIFF
--- a/ggml.c
+++ b/ggml.c
@@ -2195,6 +2195,7 @@ struct ggml_context {
     bool   mem_buffer_owned;
     bool   no_alloc;
     bool   no_alloc_save; // this is used to save the no_alloc state when using scratch buffers
+    bool   use_hwaccel;
 
     int    n_objects;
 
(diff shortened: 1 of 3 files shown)


hey-shashikant (approved):

zhouwg:
@hey-shashikant, thanks for your time and approve. could you help to take a look another PR(same to this PR)? this PR is helpful for some scenarios. Thanks.

PR #2071, opened 2024-04-17
log: refine log function for Android

zhouwg:
```
__ANDROID__ ||  ANDROID
```

are defined by Android NDK, not self-defined macro. so it will  not bring side-effect to existing codes.

this PR is prepare for submit Qualcomm's QNN backend to upstream whisper.cpp from "PoC: Add Qualcomm mobile SoC native backend for GGML,https://github.com/zhouwg/kantv/issues/121 ". 

the first step is for whisper.cpp, then llama.cpp, because llama.cpp is much more complicated then whisper.cpp

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -130,7 +130,7 @@ extern "C" int __android_log_print(int prio, const char * tag, const char * fmt,
     __attribute__((__format__(printf, 3, 4)));
 #endif
 
-WHISPER_ATTRIBUTE_FORMAT(2, 3)
+WHISPER_ATTRIBUTE_FORMAT(5, 6)
 static void whisper_log_internal        (ggml_log_level level, const char * file, const char * func, int line, const char * format, ...);
 static void whisper_log_callback_default(ggml_log_level level, const char * text, void * user_data);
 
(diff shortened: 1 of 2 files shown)


whisper.cpp
```diff
@@ -117,24 +120,29 @@ static void byteswap_tensor(ggml_tensor * tensor) {
...
 //
+#if (defined __ANDROID__) || (defined ANDROID)
+extern "C" int __android_log_print(int prio, const char * tag, const char * fmt, ...)
+    __attribute__((__format__(printf, 3, 4)));
+#endif
 
-WHISPER_ATTRIBUTE_FORMAT(2, 3)
-static void whisper_log_internal        (ggml_log_level level, const char * format, ...);
+WHISPER_ATTRIBUTE_FORMAT(5, 6)
+static void whisper_log_internal        (ggml_log_level level, const char * file, const char * func, int line, const char * format, ...);
```
imaami, line 134: Why add these three parameters at such a low level?

whisper.cpp
```diff
@@ -7196,14 +7204,19 @@ void whisper_log_set(ggml_log_callback log_callback, void * user_data) {
...
-static void whisper_log_internal(ggml_log_level level, const char * format, ...) {
+GGML_ATTRIBUTE_FORMAT(5, 6)
+static void whisper_log_internal(ggml_log_level level, const char * file, const char * func, int line, const char * format, ...) {
     va_list args;
     va_start(args, format);
     char buffer[1024];
-    int len = vsnprintf(buffer, 1024, format, args);
-    if (len < 1024) {
+    int len_prefix = snprintf(buffer, 1024, "[%s, %d]: ", func, line); // param file not used in this file
+    int len = vsnprintf(buffer + len_prefix, 1024 - len_prefix, format, args);
```
imaami, line 7213: if `snprintf` fails due to an output error then this will write to an address before the buffer's start.

imaami:
The macros `WHISPER_LOG_{ERROR,WARN,INFO,DEBUG}` are used with `__func__`, `__FILE__` or `__LINE__` already all around the codebase. They aren't meant to always be in there.

PR #2075, opened 2024-04-17
Up OpenBLAS and cuda-tookit versions build.yml

royshil:
Upgrading the version of OpenBLAS has performance gains for CPU-bound inference.

Jimver/cuda-toolkit is also outdated and was updated upstream.


royshil:
tests failing because of apt downtime - need to re-run them...

PR #2080, opened 2024-04-19
Fix for fft sizes in whisper.cpp

goldwaving:
`fft_out` needs to be twice the `frame_size`, not the `frame_step`.  It is resized in `fft()` anyway, but this change prevents an unnecessary reallocation.

`n_fft` must match the mel filter size, so it is best not to calculate it from `frame_size`.

We only need to get the magnitudes for half the spectrum since the other half is a mirror and not used in the mel filter loop later.


bobqianic:
LGTM

bobqianic (approved):

Issue #2052, opened 2024-04-14
Model fail to load if path contains Hebrew characters


thewh1teagle:
Whisper.cpp fail to load model if it contains Hebrew characters in the path, happens in Rust bindings as well.

```console
$ main.exe -m ggml-medium.bin -f single_speaker.wav
$ echo %errorlevel% // 0, success...
$ move ggml-medium.bin "ggml-medium-בדיקה.bin"
$ main.exe -m "ggml-medium-בדיקה.bin" single_speaker.wav
whisper_init_from_file_with_params_no_state: loading model from 'ggml-medium-?????.bin'
whisper_init_from_file_with_params_no_state: failed to open 'ggml-medium-?????.bin'
error: failed to initialize whisper context
echo %errorlevel% // 3, fail...
```

https://github.com/thewh1teagle/vibe/issues/42

PR #2086, opened 2024-04-22
fix: model path encoding in windows

thewh1teagle:
Resolve https://github.com/ggerganov/whisper.cpp/issues/2052 by using wide char.

By the way, [examples/main.cpp](https://github.com/ggerganov/whisper.cpp/blob/master/examples/main/main.cpp) also have problem when reading model path / wav path from `argv` in `Windows`.
Maybe it's due to invalid console encoding, I tried to enable `SetConsoleOutputCP(CP_UTF8)` and it didn't helped in main example.


ggerganov (approved):

ggerganov:
The `msys2` CI is failing now:

https://github.com/ggerganov/whisper.cpp/actions/runs/8784802899/job/24718443947?pr=2086#step:4:24

thewh1teagle:
> The `msys2` CI is failing now:
> 
> https://github.com/ggerganov/whisper.cpp/actions/runs/8784802899/job/24718443947?pr=2086#step:4:24

Fixed in https://github.com/ggerganov/whisper.cpp/pull/2086/commits/ab7263cc4d1a52c0386a5f5045a201300e2248cd

PR #2094, opened 2024-04-25
minor: add CMakeSettings.json to gitignore

iboB:
This is a Visual Studio-specific config file


iboB:
I see that some CI jobs are failing, but this PR is definitely not the reason :)

ggerganov (approved):

PR #2095, opened 2024-04-26
Fixed incorrect docker example in readme

lafe:
The docker example used Docker Hub based images that could not be found. This has been replaced with the correct Docker images.


kmoswrk:
Also the executable is /build/bin/whisper-cli

PR #2154, opened 2024-05-15
Update ruby bindings

taf2:
Want to update the ruby bindings and push a new gem.   I've added a Rakefile and would like to also think about making this more maintainable moving forward, but for today having a new gem is a good thing i think.


ggerganov (approved):

ggerganov:
> but for today having a new gem is a good thing i think.

Would you like this to be merged as it is, or do you plan to make more updates in this PR?

taf2:
As is - I’m hoping to get the new gem pushed to rubygems thanks

ggerganov:
The CI does not pass atm

PR #2184, opened 2024-05-25
Openvino audio ctx quantize

dscripka:
Compiling with [OpenVino](https://github.com/ggerganov/whisper.cpp?tab=readme-ov-file#openvino-support) on supported platforms can significantly improve performance, but currently it lacks two other features that can also significantly improve performance:

- Using quantized models (10%+ performance increase)
- Supporting [custom audio context sizes](https://github.com/ggerganov/whisper.cpp/issues/1855) (~3x performance increase for short audio files)

This PR adds both of these as optional arguments to the OpenVino model conversion scripts, enabling performance of quantization, audio context size, and OpenVino to stack.

There are a few important caveats, however:

- The OpenVino encoder (to my knowledge) only supports a fixed audio context size, so the converted model is somewhat more restricted
- This does require monkey patching a method from the `openai-whisper` library in `convert-whisper-to-openvino.py`, which isn't ideal
- Quantization is done with [nncf](https://github.com/openvinotoolkit/nncf) 2.7.0, which currently only supports 4 and 8 bit quantization

Despite these, the performance improvement can be so substantial for certain use cases it may be worth it. For example, on the ~10 second `jfk.wav` file on a Intel(R) Xeon(R) W-2123 CPU:

| Threads | Quant | Model | Encoder Time (s) | Arguments | Build |
| --- | --- | --- | --- | --- | --- |
| 1 | 8 bit | small-en | 8 | -bs 1 -ac 1500 | BLAS = 1 |
| 1 | 8 bit | small-en | 0.8 | -bs 1 -ac 550 | OPENVINO = 1 |

Command to produce the OpenVino model for this PR: `python convert-whisper-to-openvino.py --model small.en -ac 550 -qb 8
`

### DIFF
--- a/models/convert-whisper-to-openvino.py
+++ b/models/convert-whisper-to-openvino.py
@@ -1,16 +1,41 @@
 import argparse
 import torch
 from whisper import load_model
+from whisper.model import AudioEncoder
 import os
-from openvino.tools import mo
+from nncf import compress_weights, CompressWeightsMode
 from openvino.frontend import FrontEndManager
 from openvino.runtime import serialize
 import shutil
+from torch import Tensor
+import torch.nn.functional as F
 
-def convert_encoder(hparams, encoder, mname):
+# Monkey path the whisper AudioEncoder.forward method to accept variable audio context size (n_mels)
+def patched_forward(self, x: Tensor):
+    """
+    x : torch.Tensor, shape = (batch_size, n_mels, n_ctx)
+        the mel spectrogram of the audio
+    """
+    x = F.gelu(self.conv1(x))
+    x = F.gelu(self.conv2(x))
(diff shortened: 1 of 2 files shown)


jason-ni:
Is there a way to improve openvino encoder to support dynamic audio_ctx? That would be much more usefull.

dscripka:
I agree that would be ideal, but it would require more substantial modifications to how the encoder is converted and the actual OpenVino runtime implementation. Possible perhaps, but difficult.

jason-ni:
But CPU and other GPU backends all support dynamic audio_ctx given fixed model encoding size. If we allow converting OpenVino model with different encoding size, it would make model users confused.

PR #2217, opened 2024-06-06
whisper : fixes

ggerganov:
- move `mel_calc` to `whisper_state`
- `mel_calc` uses `backend` of `whisper_state` instead of `whisper_context`
- alloc kv caches using backend of `whisper_state` instead of `whisper_context`
- less C++ in `struct whisper_mel`


ggerganov:
@iboB Could you give this PR a few tests to make sure I don't break something

iboB (approved):
lgtm

PR #2220, opened 2024-06-07
Fixed a failure to free context resulting in a segfault

spschul:
Prior to this change, on my machine, after about 30 seconds or so of running the stream example, the program would segfault. Upon investigation, I found that the segfaulting was happening because the program calls state->mel_calc->calculate which eventually calls ggml_init, which uses a context from the global state. However, that context is never freed. This change frees that context whenever it is about to be replaced. It fixes the segfault, but I don't understand the code enough to say whether it's the right approach.


ggerganov:
Should be fixed with [c2bdb96](https://github.com/ggerganov/whisper.cpp/commit/c2bdb960cdb69af813d3a4bd72a90f852d695127)

PR #2227, opened 2024-06-10
whisper : auto-grow working areas for mel_calc_cuda

iboB:
besides the auto grow additional code was added to only calculate the mel spectrogram on the GPU if the audio is shorter than (the somewhat arbitrary) 5 minutes.

Otherwise these temporary buffers can reach hundreds of megabytes of permanently allocated GPU memory which we consider unwanted.

### DIFF
--- a/whisper-mel-cuda.cu
+++ b/whisper-mel-cuda.cu
@@ -145,17 +145,6 @@ void calc_magnitudes(
 
 constexpr auto LOG_MEL_PREFIX_SIZE = 256;
 
-size_t get_log_mel_temp_storage_size() {
-    constexpr auto maxPaddedSamples = 2 * WHISPER_N_SAMPLES + WHISPER_N_FFT;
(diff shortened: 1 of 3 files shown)


iboB:
The fail on windows is failing because msvc is updated, but not Cuda. It's a combination of silliness between Microsoft and Nvidia.

Expect this on all other msvc-Cuda bulds

iboB:
@slaren and Nvidia's is setting a minor version (1940) as max supported host compiler even though Microsoft swore and promised abi compatibility for all 19xx versions

iboB:
By the way if you manually edit host_config.h from Cuda, as expected everything works. That's for one wanting to get their local builds going. Sadly it can't be done for gh actions workers

slaren:
It may work with the gh actions by passing `--allow-unsupported-compiler`.

ggerganov (approved):

PR #2231, opened 2024-06-11
cuda : fix bounds check for src0 rows in MMVQ kernel

ggerganov:
The non-EN models have a tensor with odd number of rows: 

```
decoder.token_embedding.weight - [ 1024, 51865,     1]
```

With BS > 1 and since https://github.com/ggerganov/whisper.cpp/commit/d7e9f58f7f286b1eea478d54b17360344d1214ba we have been doing out-of-bound writes with quantum models because `rows_per_cuda_block` can be equal to `2`, leading to significantly poor transcription quality:

https://github.com/ggerganov/whisper.cpp/blob/20c542c71334b3e6c422789093a19157b110ca81/ggml-cuda/mmvq.cu#L92-L123

Repro on RTX 2060:

```bash
# any quantum model would reproduce the problem
rm models/ggml-base-q5_1.bin
./models/download-ggml-model.sh base-q5_1

make clean
WHISPER_CUDA=1 make -j && ./main -m models/ggml-base-q5_1.bin -f samples/gb0.wav
```

### DIFF
--- a/ggml-cuda/mmvq.cu
+++ b/ggml-cuda/mmvq.cu
@@ -75,7 +75,7 @@ static __global__ void mul_mat_vec_q(
             tmp[j][i] = warp_reduce_sum(tmp[j][i]);
         }
 
-        if (threadIdx.x < rows_per_cuda_block && row0 + threadIdx.x < nrows_dst) {
+        if (threadIdx.x < rows_per_cuda_block && (rows_per_cuda_block == 1 || row0 + threadIdx.x < nrows_dst)) {
             dst[j*nrows_dst + row0 + threadIdx.x] = tmp[j][threadIdx.x];
         }
     }
--- a/ggml-cuda/mmvq.cu
+++ b/ggml-cuda/mmvq.cu
@@ -75,7 +75,7 @@ static __global__ void mul_mat_vec_q(
             tmp[j][i] = warp_reduce_sum(tmp[j][i]);
         }
 
-        if (threadIdx.x < rows_per_cuda_block) {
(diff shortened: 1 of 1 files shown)


ggerganov:
cc @JohannesGaessler for review

JohannesGaessler:
This is a more general issue (also for llama.cpp) and I think there's a better way to fix it. A PR to which repository would be the least amount of work for you to sync?

ggml-cuda/mmvq.cu
```diff
@@ -75,7 +75,7 @@ static __global__ void mul_mat_vec_q(
             tmp[j][i] = warp_reduce_sum(tmp[j][i]);
         }
 
-        if (threadIdx.x < rows_per_cuda_block) {
+        if (threadIdx.x < rows_per_cuda_block && row0 + threadIdx.x < nrows_dst) {
```
JohannesGaessler, line 78:
```suggestion
        if (threadIdx.x < rows_per_cuda_block && (rows_per_cuda_block == 1 || row0 + threadIdx.x < nrows_dst)) {
```

JohannesGaessler (approved):
The performance difference from this change (for llama.cpp) is negligible and for batch size 1 you can optimize out the check anyways. So I think it's fine to just fix it like this. Note that these out-of-bounds writes are potentially also causing issues for llama.cpp if whoever finetuned the model added an extra token to the vocabulary.

ggerganov:
Thanks, I'll merge it here and will sync back to `llama.cpp` soon

Issue #2230, opened 2024-06-11
Segmentation fault with CUDA on commit (#2206)


themanyone:
Steps to reproduce

`WHISPER_CUDA=1 make -j`
`./main -m ./models/ggml-medium-q5_0.bin samples/jfk.wav`

**Crash.**
whisper_init_from_file_with_params_no_state: loading model from './models/ggml-medium-q5_0.bin'
whisper_init_with_params_no_state: use gpu    = 1
whisper_init_with_params_no_state: flash attn = 0
whisper_init_with_params_no_state: gpu_device = 0
whisper_init_with_params_no_state: dtw        = 0
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51865
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 1024
whisper_model_load: n_audio_head  = 16
whisper_model_load: n_audio_layer = 24
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 1024
whisper_model_load: n_text_head   = 16
whisper_model_load: n_text_layer  = 24
whisper_model_load: n_mels        = 80
whisper_model_load: ftype         = 8
whisper_model_load: qntvr         = 1
whisper_model_load: type          = 4 (medium)
whisper_model_load: adding 1608 extra tokens
whisper_model_load: n_langs       = 99
whisper_backend_init: using CUDA backend
ggml_cuda_init: GGML_CUDA_FORCE_MMQ:   no
ggml_cuda_init: CUDA_USE_TENSOR_CORES: yes
ggml_cuda_init: found 1 CUDA devices:
  Device 0: Quadro M3000M, compute capability 5.2, VMM: yes
whisper_model_load:    CUDA0 total size =   538.59 MB
whisper_model_load: model size    =  538.59 MB
Segmentation fault (core dumped

**What works.** Versions prior to this commit:

$ `git bisect bad`
ffef323c4cfa8596cb91cf92d6f791f01a44335e is the first bad commit
commit ffef323c4cfa8596cb91cf92d6f791f01a44335e (HEAD)
Author: Borislav Stanimirov <b.stanimirov@abv.bg>
Date:   Tue Jun 4 09:32:23 2024 +0300

    whisper : add CUDA-specific computation mel spectrograms (#2206)
    
    * whisper : use polymorphic class to calculate mel spectrogram
    
    * whisper : add cuda-specific mel spectrogram calculation
    
[...]

PR #2235, opened 2024-06-12
cuda : avoid async allocs in CUDA mel code

ggerganov:
fix #2230

PR #2236, opened 2024-06-12
whisper : use ggml-cuda in mel calc, set appropriate device

iboB:
* fix issues where whisper may be set to run on a device which is not the current one for all `mel_calc` calls.
* workaround for #2230 as opposed to #2235 : forbid `mel_calc_cuda` if device cc is below 600
* allow `whisper_mel_calc_create_cuda` to return null safely


themanyone (approved):
Fixes https://github.com/ggerganov/whisper.cpp/issues/2230
Seems plenty fast.

ggerganov (approved):

PR #2238, opened 2024-06-16
sync : ggml-blas

ggerganov:
TODO:

- [x] update CMake / Makefile builds
- [x] try to introduce `ggml_backend_sched`: https://github.com/ggerganov/whisper.cpp/pull/2239


src/ggml-blas.cpp
slaren, line 1: This file and `src/ggml-blas.h` probably shouldn't be here.

PR #2239, opened 2024-06-16
whisper : use ggml_backend_sched

ggerganov:
target branch: https://github.com/ggerganov/whisper.cpp/tree/sync-blas

Using BLAS just in the Encoder since it involves large matrices

### DIFF
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -539,6 +539,9 @@ static bool whisper_allocr_graph_init(struct whisper_allocr & allocr, std::vecto
         WHISPER_LOG_ERROR("%s: failed to allocate the compute buffer\n", __func__);
         return false;
     }
+
+    ggml_backend_sched_reset(sched);
+
     return true;
 }
 
@@ -2402,17 +2405,17 @@ static struct ggml_cgraph * whisper_build_graph_decoder(
 
     struct ggml_tensor * embd = ggml_new_tensor_1d(ctx0, GGML_TYPE_I32, n_tokens);
     ggml_set_name(embd, "embd");
-    //ggml_set_input(embd);
+    ggml_set_input(embd);
 
     struct ggml_tensor * position = ggml_new_tensor_1d(ctx0, GGML_TYPE_I32, n_tokens);
     ggml_set_name(position, "position");
-    //ggml_set_input(position);
+    ggml_set_input(position);
 
     const float KQscale = pow(float(n_state_head), -0.25);
 
     struct ggml_tensor * KQ_mask = ggml_new_tensor_3d(ctx0, GGML_TYPE_F32, n_kv, GGML_PAD(n_tokens, GGML_KQ_MASK_PAD), 1);
     ggml_set_name(KQ_mask, "KQ_mask");
-    //ggml_set_input(KQ_mask);
+    ggml_set_input(KQ_mask);
 
     struct ggml_tensor * KQ_mask_f16 = ggml_cast(ctx0, KQ_mask, GGML_TYPE_F16);
 
--- a/whisper.cpp
+++ b/whisper.cpp
@@ -831,6 +831,7 @@ struct whisper_state {
     whisper_decoder decoders[WHISPER_MAX_DECODERS];
 
     std::vector<ggml_backend_t> backends;
+    std::vector<ggml_backend_t> backends_used;
 
     // ggml-alloc:
     // - stores meta info about the intermediate tensors into the `meta` buffers
@@ -898,7 +899,7 @@ struct whisper_context {
 
     whisper_state * state = nullptr;
 
-    std::vector<ggml_backend_t> backends;
+    ggml_backend_t backend;
 
     std::string path_model; // populated by whisper_init_from_file_with_params()
 };
@@ -1090,13 +1091,13 @@ static uint32_t whisper_kv_cache_get_padding(const struct whisper_context & wctx
     }
 
 #ifdef GGML_USE_METAL
-    if (ggml_backend_is_metal(wctx.backends[0])) {
+    if (ggml_backend_is_metal(wctx.backend)) {
         return 32u;
     }
 #endif
 
 #ifdef GGML_USE_CUDA
-    if (ggml_backend_is_cuda(wctx.backends[0])) {
+    if (ggml_backend_is_cuda(wctx.backend)) {
         return 256u;
     }
 #endif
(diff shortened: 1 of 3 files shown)


whisper.cpp
```diff
@@ -3299,20 +3357,26 @@ static std::string whisper_openvino_get_path_cache(std::string path_bin) {
...
         whisper_free_state(state);
         return nullptr;
     }
 
-    state->mel_calc = whisper_mel_calc_create(state->backend, ctx->model.filters);
+    state->backends_used.push_back(ctx->backend);
+    state->backends_used.insert(state->backends_used.end(), state->backends.begin(), state->backends.end());
+    for (int i = 0; i < (int) state->backends_used.size(); ++i) {
+        printf("state->backends_used[%d] = %p, name = %s\n", i, state->backends_used[i], ggml_backend_name(state->backends_used[i]));
+    }
```
ggerganov, line 3371: The backend instance in `whisper_context` is used to create a read-only buffer that contains all tensors of the model. Do I need to pass this backend together with `whisper_state` backends when initiating the schedulers? Currently it works both ways

whisper.cpp
```diff
@@ -2347,17 +2405,17 @@ static struct ggml_cgraph * whisper_build_graph_decoder(
...
     ggml_set_name(position, "position");
-    ggml_set_input(position);
+    //ggml_set_input(position);
 
     const float KQscale = pow(float(n_state_head), -0.25);
 
     struct ggml_tensor * KQ_mask = ggml_new_tensor_3d(ctx0, GGML_TYPE_F32, n_kv, GGML_PAD(n_tokens, GGML_KQ_MASK_PAD), 1);
     ggml_set_name(KQ_mask, "KQ_mask");
-    ggml_set_input(KQ_mask);
+    //ggml_set_input(KQ_mask);
```
ggerganov, line 2418: When running with the Metal backend, having these `ggml_set_input()` calls in the decoder graph causes the assigned backends to be NULL (nodes 0, 1, 27):

```
## SPLIT #0: Metal # 0 inputs: 
node #  0 (  GET_ROWS):               node_0 (   2K) [Metal   1.wgt0]:               leaf_1 (  50M) [Metal    1.dst]         Metal#embd#0 (   1K) [ NULL    4.cpy]
node #  1 (  GET_ROWS):               node_1 (   2K) [Metal   1.wgt0]:               leaf_3 ( 896K) [Metal    1.dst]     Metal#position#0 (   1K) [ NULL    4.cpy]
node #  2 (       ADD):               node_2 (   2K) [Metal    2.sup]:               node_0 (   2K) [Metal   1.wgt0]               node_1 (   2K) [Metal   1.wgt0]
node #  3 (      NORM):               node_3 (   2K) [Metal    2.sup]:               node_2 (   2K) [Metal    2.sup]
node #  4 (       MUL):               node_4 (   2K) [Metal   1.wgt1]:               node_3 (   2K) [Metal    2.sup]               leaf_5 (   2K) [Metal    1.dst]
node #  5 (       ADD):               node_5 (   2K) [Metal   1.wgt1]:               node_4 (   2K) [Metal   1.wgt1]               leaf_6 (   2K) [Metal    1.dst]
node #  6 (   MUL_MAT):               node_6 (   2K) [Metal   1.wgt0]:               leaf_0 ( 512K) [Metal    1.dst]               node_5 (   2K) [Metal   1.wgt1]
node #  7 (     SCALE):               node_7 (   2K) [Metal    2.sup]:               node_6 (   2K) [Metal   1.wgt0]
node #  9 (       CPY): leaf_7 (view) (copy  (   1K) [Metal   1.vsrc]:               node_7 (   2K) [Metal    2.sup]        leaf_7 (view) (   1K) [Metal   1.vsrc]
node # 10 (   MUL_MAT):              node_10 (   2K) [Metal   1.wgt0]:               leaf_8 ( 512K) [Metal    1.dst]               node_5 (   2K) [Metal   1.wgt1]
node # 11 (       ADD):              node_11 (   2K) [Metal   1.wgt1]:              node_10 (   2K) [Metal   1.wgt0]               leaf_9 (   2K) [Metal    1.dst]
node # 15 (       CPY): leaf_10 (view) (copy (   1M) [Metal   1.vsrc]:  (reshaped) (transpo (   2K) [Metal   4.vsrc]       leaf_10 (view) (   1M) [Metal   1.vsrc]
node # 21 (   MUL_MAT):              node_21 (   2K) [Metal   1.wgt0]:              leaf_19 ( 512K) [Metal    1.dst]               node_5 (   2K) [Metal   1.wgt1]
node # 22 (       ADD):              node_22 (   2K) [Metal   1.wgt1]:              node_21 (   2K) [Metal   1.wgt0]              leaf_20 (   2K) [Metal    1.dst]
node # 23 (     SCALE):              node_23 (   2K) [Metal    2.sup]:              node_22 (   2K) [Metal   1.wgt1]
node # 26 (   MUL_MAT):              node_26 (   0K) [Metal    2.sup]:        leaf_7 (view) (   1K) [Metal   1.vsrc]  (reshaped) (permute (   2K) [Metal   4.vsrc]
node # 27 (  SOFT_MAX):              node_27 (   0K) [Metal    2.sup]:              node_26 (   0K) [Metal    2.sup]      Metal#KQ_mask#0 (   2M) [ NULL    4.cpy]
node # 28 (   MUL_MAT):              node_28 (   2K) [Metal    2.sup]:       leaf_10 (view) (   1M) [Metal   1.vsrc]              node_27 (   0K) [Metal    2.sup]
```

Is this expected? Without Metal it works as expected. Removing the calls also seems to work, but seems incorrect since the `KQ_mask` might get overwritten?

whisper.cpp
ggerganov: I think this is actually not necessary and can have side-effect of not using the BLAS backend if the `whisper_context` backend is CPU. So I'll pass just the `whisper_state` backends to the schedulers

slaren: Backend instances are not necessary at all to create buffers at all, it's just more convenient because you can use the `ggml_backend_get_default_buffer_type` to obtain the preferred buffer type of this backend, rather than having to call a backend-specific function such as `ggml_backend_metal_buffer_type`, which would require a bunch of `#ifdefs` such as in `llama_default_buffer_type_offload`. But the backend instance is not bound to the buffer in any way, and it is not necessary to pass it to the sched.

ggerganov: Got it, will try to remove the `backend` instance from `whisper_context` then to simplify things

whisper.cpp
slaren: I am not sure what's happening. When I uncomment this line, I get `ggml_metal_get_buffer: error: tensor '��M9' buffer is nil`, and the garbled name suggests memory corruption.

ggerganov: It's the same here. I thought the source of the issue is the `NULL` backend assigned for these nodes

slaren: The `NULL` is expected here, it's because split inputs do not have a backend assignment in the hashtable. This is just so that the size of the hash table remains predictable. I think the tensors structs are being overwritten, but I am not sure where.

whisper.cpp
```diff
@@ -1229,34 +1244,76 @@ static ggml_backend_t whisper_backend_init(const whisper_context_params & params
...
+}
+
+static ggml_backend_buffer_type_t whisper_default_buffer_type(const whisper_context_params & params) {
+    if (!params.use_gpu) {
+        return ggml_backend_cpu_buffer_type();
     }
 
-    return ggml_backend_cpu_init();
+#ifdef GGML_USE_CUDA
+    return ggml_backend_cuda_buffer_type(params.gpu_device);
```
slaren, line 1305: This will return `NULL` if there are no CUDA devices. To allow it to continue and fallback to the CPU backend, it should check for that in a similar way the llama.cpp function does this.

PR #2240, opened 2024-06-16
use sched in whisper_allocr

slaren:
The target is https://github.com/ggerganov/whisper.cpp/pull/2239

Not sure if this is better, but it works.

Increasing the number of threads of the BLAS backend seem to results in better performance with BLIS, but setting the same number of threads in the CPU backend results in lower overall performance. For best performance, I think it is necessary to have different numbers of threads depending on the batch size.

### DIFF
--- a/ggml-backend.c
+++ b/ggml-backend.c
@@ -1706,14 +1706,16 @@ static void ggml_backend_sched_split_graph(ggml_backend_sched_t sched, struct gg
 static bool ggml_backend_sched_alloc_splits(ggml_backend_sched_t sched) {
     bool backend_ids_changed = false;
     for (int i = 0; i < sched->graph->n_nodes; i++) {
-        if (sched->node_backend_ids[i] != sched->prev_node_backend_ids[i]) {
+        if (sched->node_backend_ids[i] != sched->prev_node_backend_ids[i] &&
+            sched->bufts[sched->node_backend_ids[i]] != sched->bufts[sched->prev_node_backend_ids[i]]) {
             backend_ids_changed = true;
             break;
         }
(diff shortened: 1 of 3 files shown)


ggerganov:
Nice, very helpful. Currently it fails with Metal enabled:

```bash
make -j && ./main -m models/ggml-base.en.bin -f ./samples/gb0.wav 
```

```
system_info: n_threads = 4 / 10 | AVX = 0 | AVX2 = 0 | AVX512 = 0 | FMA = 0 | NEON = 1 | ARM_FMA = 1 | METAL = 1 | F16C = 0 | FP16_VA = 1 | WASM_SIMD = 0 | BLAS = 1 | SSE3 = 0 | SSSE3 = 0 | VSX = 0 | CUDA = 0 | COREML = 0 | OPENVINO = 0

main: processing './samples/gb0.wav' (2037686 samples, 127.4 sec), 4 threads, 1 processors, 5 beams + best of 5, lang = en, task = transcribe, timestamps = 1 ...

whisper_mel_init: n_len = 15735, n_len_org = 12735, n_mel = 80
ggml_metal_get_buffer: error: tensor '0' buffer is nil
Bus error: 10
```

Looking into this. I also think I might be doing something wrong with the backend allocations both in `whisper_context` and `whisper_state` but when creating the scheduler, I pass only the backends from the `whisper_state`. My guess is I have to allocate only one backend in `whisper_context` instead of the full set of backends (e.g. Metal + BLAS + CPU). And also make sure that the backend from `whisper_context` is also passed to `ggml_backend_sched_new()` together with the backends from `whisper_state`

PR #2291, opened 2024-07-08
Implementing Encoder Begin Callback for golang binding

aderbedr:
Adding in EncoderBeginCallback to the Context's Process callback. This optional callback function returns false if computation should be aborted.


ggerganov (approved):

aderbedr:
Are you alright if I close this PR and open a separate that just contains the Encoder Begin Callback commit? I forgot about this PR and put a slew of other commits on the same branch, ruining this PR, then noticed it was failing because of a test file that needed updating. https://github.com/ggerganov/whisper.cpp/compare/master...aderbedr:whisper.cpp:feat/encoderBeginCallback2 would be the branch just for this change.

ggerganov:
Yes, feel free to open a new PR with just the callback change.

PR #2302, opened 2024-07-14
whisper: use vulkan as gpu backend when available

mstephenson6:
- When built with Vulkan support, attempt to initialize the ggml vk backend as a gpu

### DIFF
--- a/ggml/src/ggml-vulkan.cpp
+++ b/ggml/src/ggml-vulkan.cpp
@@ -2042,9 +2042,9 @@ void ggml_vk_instance_init() {
 }
 
 static void ggml_vk_init(ggml_backend_vk_context * ctx, size_t idx) {
-    GGML_ASSERT(idx < vk_instance.device_indices.size());
     VK_LOG_DEBUG("ggml_vk_init(" << ctx->name << ", " << idx << ")");
     ggml_vk_instance_init();
+    GGML_ASSERT(idx < vk_instance.device_indices.size());
 
     ctx->name = GGML_VK_NAME + std::to_string(idx);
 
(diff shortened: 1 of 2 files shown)


ggerganov:
Hi, have you tested that it works correctly when Vulkan is enabled?

mstephenson6:
@ggerganov - yes, so far just on an Apple M1Pro running Fedora, with the "[Honeykrisp](https://rosenzweig.io/blog/vk13-on-the-m1-in-1-month.html)" Vulkan driver. 

I'm excited about this working, it's the very first GPU-accelerated ML I've been able to do on Linux + M1.

Happy to test other GPU/OS combos too.

ggml/src/ggml-vulkan.cpp
```diff
@@ -2042,9 +2042,9 @@ void ggml_vk_instance_init() {
 }
 
 static void ggml_vk_init(ggml_backend_vk_context * ctx, size_t idx) {
-    GGML_ASSERT(idx < vk_instance.device_indices.size());
     VK_LOG_DEBUG("ggml_vk_init(" << ctx->name << ", " << idx << ")");
     ggml_vk_instance_init();
+    GGML_ASSERT(idx < vk_instance.device_indices.size());
```
ggerganov, line 2047: Is this change necessary? The original code works in https://github.com/ggerganov/llama.cpp, not sure why we would need to change it here

mstephenson6: I changed it so `ggml_vk_instance_init()` would populate `device_indices` right before the assertion, and it would avoid another vk-specific call in whisper.cpp. Without this, the assertion failed every time, no other code had attempted to initialize devices.

I didn't look at how llama.cpp did it, but I can check it out.

ggerganov: I see, seems the change is good. cc @0cc4m to confirm that it's OK to move the assert after the init call

0cc4m: Yeah, that's fine. Makes more sense that way, too.

ggerganov (approved):

PR #2346, opened 2024-08-12
Set MSVC to use UTF-8 on source files

drasticactions:
Although the files in Whisper.cpp are encoded with UTF-8, MSVC seems to take the system's default text encoding into account when compiling.

https://github.com/ggerganov/whisper.cpp/blob/master/src/whisper.cpp#L4982C1-L4987C3

So when MSVC reencodes this into the system encoding (Ex. Shift-JIS), it blows up when compiling. We could change this to use Unicode escape characters, but if we set MSVC to treat all files as UTF-8, it should compile fine on all systems, regardless of the default encoding and whatever text is used here.

As far as I know, this should only affect compiling, and the build output should be the same.


chinshou:
I can confirm that I have the same compile problem in a Japanese Windows.

Issue #2349, opened 2024-08-13
[gobinding] add more methods to context interface


hsinhoyeh:
Hi there,

I am using go binding for whisper and want to add some functions in context interface

```
+       SetMaxContext(n int)            // Set maximum number of text context tokens to store
+       SetBeamSize(n int)              // Set Beam Size
+       SetEntropyThold(t float32)      // Set Entropy threshold
```

just wondering what is the best way to keep it in sync or we just update it manually?

hsinhoyeh:
just fire a PR first, and see if we have better idea (or more functions to add).

PR #2339, opened 2024-08-06
fix go bindings

10Jib:
Builds off of #2330
 
Fixes some issues with go bindings such as:
- Links are updated for whisper and ggml
- Added OpenMP flag for non apple systems
- Updated model path used in the testcases

PR #2330, opened 2024-07-31
fix go bindings

gyorgy1:
This PR fixes missing ggml issues in go bindings

PR #2350, opened 2024-08-13
feat(go binding): add beamsize/entropythold/maxcontext to context interface

hsinhoyeh:
fixes: #2349

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -972,8 +972,7 @@ $(LIB_WHISPER): \
 
 $(LIB_WHISPER_S): \
 	$(OBJ_WHISPER) \
-	$(OBJ_GGML) \
-	$(LIB_GGML)
+	$(OBJ_GGML)
 	ar rcs $(LIB_WHISPER_S) $^
 
 # common


ggerganov (approved):

hsinatfootprintai:
@ggerganov I think the build failed for go building has been fixed. please check it.

hsinhoyeh:
related issue: #2330 #2339

hsinatfootprintai:
@ggerganov seems it didn't re-test automatically, require manually trigger

ggerganov:
Just triggered it again - let's see if it's green

hsinhoyeh:
@ggerganov it is green now. it seems that we need a static library `libwhisper.a` as well as a dynamic library `libggml.so`.
is it intentional? making a clean path for lamma.cpp I guess?

ggerganov:
> it seems that we need a static library libwhisper.a as well as a dynamic library libggml.so.

In what sense we need both?

hsinhoyeh:
> > it seems that we need a static library libwhisper.a as well as a dynamic library libggml.so.
> 
> In what sense we need both?

okey, so it is clear that this is NOT intentional...

hsinhoyeh:
> > it seems that we need a static library libwhisper.a as well as a dynamic library libggml.so.
> 
> In what sense we need both?

@ggerganov this CL makes much more sense now.

Makefile
```diff
@@ -967,11 +967,14 @@ src/whisper.o: \
 
 $(LIB_WHISPER): \
 	$(OBJ_WHISPER) \
+	$(OBJ_GGML) \
```
ggerganov, line 970: I don't think it is necessary to include the `ggml` object files in the `whisper` dynamic lib. We want to link `ggml` dynamically

hsinatfootprintai: ya, removed.

hsinhoyeh:
@ggerganov This issue may require your attention, as similar issues with Go bindings keep recurring.

Makefile
```diff
@@ -971,7 +971,9 @@ $(LIB_WHISPER): \
 	$(CXX) $(CXXFLAGS) -shared -fPIC -o $@ $^ $(LDFLAGS)
 
 $(LIB_WHISPER_S): \
-	$(OBJ_WHISPER)
+	$(OBJ_WHISPER) \
+	$(OBJ_GGML) \
+	$(LIB_GGML)
```
ggerganov, line 976: Is this change necessary?

hsinhoyeh: ya, otherwise it can't find `ggml_xxx` library

error message like:

```
/usr/bin/ld: whisper.cpp:(.text+0x2a82b): undefined reference to `ggml_time_us'
/usr/bin/ld: whisper.cpp:(.text+0x2a881): undefined reference to `ggml_time_us'
/usr/bin/ld: whisper.cpp:(.text+0x2b191): undefined reference to `ggml_time_us'
/usr/bin/ld: whisper.cpp:(.text+0x2b2e1): undefined reference to `ggml_time_us'
/usr/bin/ld: /home/ubuntu/xxxx/libwhisper.a(whisper.o):whisper.cpp:(.text+0x2b5fe): more undefined references to `ggml_time_us' follow

```

PR #2360, opened 2024-08-17
Use colorblind friendly TTY color scheme

jart:
This change updates the -pc flag, so that a new xterm256 color scheme is used. This color scheme is believed to be better for three reasons:

1. It should be friendlier to the colorblind. The scheme was designed by Paul Tol (see: https://personal.sron.nl/~pault/). TensorBoard uses it since 2017, so it's already popular in the machine learning community

2. It should appear to be the same colors as before to people who aren't i.e. it's still a red-green spectrum like before but lightly modified

3. It is readable in both white and black background terminals. The neon colors before were probably a bit too intense for white backgrounds.

### Before

image

### After

image

### DIFF
--- a/examples/common.h
+++ b/examples/common.h
@@ -9,6 +9,7 @@
 #include <thread>
 #include <ctime>
 #include <fstream>
+#include <sstream>
 
 #define COMMON_SAMPLE_RATE 16000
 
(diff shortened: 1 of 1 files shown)


Ghostboo-124 (approved):
Looks good

ggerganov (approved):

espinielli:
Great contribution...a great help for me personally...I can now SEE!

jart:
I'm so happy to hear that @espinielli!

Issue #2395, opened 2024-08-29
[BUG] Invalid flags in python example


UsernamesLame:
In  commit a13a7da, specifically ```whisper_processor.py``` there's this line: ```    full_command = f"./main -m {model} -f {wav_file} -np -nt"```

The flag ```-np``` is a nonexistent flag according to ```README.md```

Issue #2394, opened 2024-08-29
README.md C style API link is broken


UsernamesLame:
The link in README.md for the C style API is broken. Just a heads up!

PR #2396, opened 2024-08-29
Fix broken link in README.md and remove invalid flag from Python example

UsernamesLame:
Fix broken C-style API link in README.md, and remove nonexistent whisper flags from python example.


UsernamesLame:
This PR fixes #2395 and #2394.

ggerganov (approved):

PR #2419, opened 2024-09-16
server: Use OS-generated temp file name for ffmpeg converted files

teejae:
When doing auto convert of audio files, always generate new temporary filenames, rather than a fixed filename.
- Allows running multiple conversions/transcriptions simultaneously without risk of overwriting files

### DIFF
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -677,7 +677,8 @@ int main(int argc, char ** argv) {
         if (sparams.ffmpeg_converter) {
             // if file is not wav, convert to wav
             // write to temporary file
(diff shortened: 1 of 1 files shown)


ggerganov:
I'm not super familiar with `tmpnam`, but I see the compiler giving the following warning:

```txt
/usr/bin/ld: examples/server/server.o: in function `main::{lambda(httplib::Request const&, httplib::Response&)#3}::operator()(httplib::Request const&, httplib::Response&) const':
server.cpp:(.text+0x4d39): warning: the use of `tmpnam' is dangerous, better use `mkstemp'
```

Could you propose a way to fix this warning?

PR #2432, opened 2024-09-26
Add support for decoding using FFmpeg in the `main` example

shih-liang:


kth8:
> Could be interesting to have this in main for those you don't want to install ffmpeg. But you can also simply use directlty ffmpeg to pipe into main. e.g. : ffmpeg -i file -ar 16000 -ac 2 -c:a pcm_s16le -f wav - | main -m model.bin -f -

Thanks for the tip. I wanted to see how many pipes I can use together
```
yt-dlp --quiet -o - https://www.youtube.com/watch?v=1ShSxdgYa7w | \
ffmpeg -v quiet -i - -ar 16000 -ac 2 -c:a pcm_s16le -f wav - | \
bin/main --no-prints --no-timestamps -fa -m models/ggml-large-v3-turbo.bin -f - 2>/dev/null | \
ollama run llama3.2:3b "provide a summary in paragraph format" | \
tee /dev/stderr | \
../piper/piper --model ../piper/en_US-hfc_male-medium.onnx --output_raw 2>/dev/null | \
aplay -r 22050 -f S16_LE -t raw - 2>/dev/null
```

imaami:
The main commit is absolutely flooded with unnecessary and invasive whitespace and brace style changes. Even if such changes were called for, they would belong in a separate commit instead of mixed in with functional changes.

But the problem here isn't simply whitespace fixes polluting a functional commit. These aren't whitespace fixes but changes. One function originally indented with four spaces is changed to tab indentation, and above it a tab-indented function is changed to four-space indentation...

The upstream code *does* suffer from a bad case of mixed indentation, but any fixes need to be in separate PRs or at least in separate commits. And of course they need to actually be consistent to be fixes at all.

Issue #2334, opened 2024-08-03
whisper_full_with_state: failed to decode


MathiasSchindler:
After running successfully for 6 minutes of audio, whisper.cpp (compiled with CUDA support) exits with the following message:

```
[00:06:04.000 --> 00:06:11.000]   Gut, dann kommen wir zur Sache, nämlich zum digitale Dienste Gesetz.
whisper_full_with_state: failed to decode
./main: failed to process audio
```

This error happens using the ggml-tiny.bin as a model and the -bs 8 -bo 8 flags. Using the base model, the same issue happens some minutes later.

Are there any tools available to get a more descriptive error message that could help narrow the source of the issue?

Kostyansa:
Stuck on the same problem but with large-v3 model. `-bs 6 -bo 6` works fine, any -bs larger than 6 - fails.

WilliamTambellini:
same issue here with ggml-large-v2.bin or ggml-large-v3.bin beamsize 8 whispercpp 1.5.4
@ggerganov  any comment?

ggerganov:
Can you provide a sample audio that reproduces the isse?

ggerganov:
Sorry, haven't had much time lately to work on `whisper.cpp`. Maybe sometime next week. Hopefully someone figures it out in the meantime.

WilliamTambellini:
If I could help a bit: seems to be a failure in: 
```
whisper_kv_cache_find_slot: failed to find a slot for 8 tokens. n_tested=1345 n_ctx=1344
whisper_decode_internal: failed to whisper_kv_cache_find_slot
whisper_full_with_state: failed to decode
```

```
        if (n_tested >= n_ctx) {
            WHISPER_LOG_ERROR("%s: failed to find a slot for %d tokens. n_tested=%d n_ctx=%d\n", __func__, n_tokens, n_tested, n_ctx);
            return false;
        }

```

WilliamTambellini:
@ggerganov 
I m now wondering about these 2 lines: 
https://github.com/ggerganov/whisper.cpp/blob/34972dbe221709323714fc8402f2e24041d48213/src/whisper.cpp#L987C9-L991C10

```
 if (cache.head + n_tokens > n_ctx) {
            n_tested += n_ctx - cache.head;
            cache.head = 0;
            continue;
        }
```
Searching at the beginning of the kvcache (head=0) could make sens but why increasing n_tested ?
Best

WilliamTambellini:
[...]

PR #2433, opened 2024-09-26
Add a new param to set the self attn text context factor

WilliamTambellini:
Add an env var to set the text context factor
WHISPER_SELFATTN_CACHE_TEXT_CTX_FACTOR
same default to 3.
Resolve:
https://github.com/ggerganov/whisper.cpp/issues/2334

### DIFF
--- a/examples/main/main.cpp
+++ b/examples/main/main.cpp
@@ -38,6 +38,7 @@ struct whisper_params {
     int32_t max_len       = 0;
     int32_t best_of       = whisper_full_default_params(WHISPER_SAMPLING_GREEDY).greedy.best_of;
     int32_t beam_size     = whisper_full_default_params(WHISPER_SAMPLING_BEAM_SEARCH).beam_search.beam_size;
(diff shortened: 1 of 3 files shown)


amoskahiga:
Thanks for the fix! Looks good to me.

WilliamTambellini:
@aderbedr FYI

aderbedr (approved):
lgtm

matiaslin:
Great! This looks good to me.

WilliamTambellini:
@ggerganov ?

ggerganov:
It's better to avoid the environment variable. Should implement this by adding a `int max_decoders` parameter to `whisper_context_params` and using it for the `factor`.

WilliamTambellini:
ok @ggerganov 
I ve reimplemented and repushed.
Please re-review.
best

ggerganov:
@WilliamTambellini I've pushed an alternative fix. Can you give it a try: https://github.com/ggerganov/whisper.cpp/pull/2443

ggerganov:
Superseded by #2443

PR #2449, opened 2024-10-03
update dr_wav.h to newer version to fix _MSC_VER macro issues

RahulVadhyar:
While using this repo in my project, I have encountered compiling issues with dr_wav.h. Specifically on linux, I have been getting undefined _MSC_VER issues. This has been fixed in the latest versions of dr_wav.h. Hence this PR updates this file to fix the issue.

### DIFF
--- a/examples/dr_wav.h
+++ b/examples/dr_wav.h
@@ -1,6 +1,6 @@
 /*
 WAV audio loader and writer. Choice of public domain or MIT-0. See license statements at the end of this file.
-dr_wav - v0.12.16 - 2020-12-02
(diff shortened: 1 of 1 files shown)


ggerganov:
Thanks, could you point me to the reference file so I can verify that it matches?

RahulVadhyar:
> Thanks, could you point me to the reference file so I can verify that it matches?

Sure, here is the link to the file:

https://github.com/mackron/dr_libs/blob/master/dr_wav.h

ggerganov (approved):

Issue #2450, opened 2024-10-03
Vulkan: failed to allocate memory


SRHMorris:
This is having built whisper.cpp from commit [ccc2547](https://github.com/ggerganov/whisper.cpp/commit/ccc2547210e09e3a1785817383ab770389bb442b) with Vulkan support.

I assume this is related to the rebar stuff that was recently added to ggml? Previously at commit 5236f02 this wasn't an issue, though this machine did still have issues with poor transcription.

```
whisper_init_from_file_with_params_no_state: loading model from '/some/path/ggml-base.bin'
whisper_init_with_params_no_state: use gpu    = 1
whisper_init_with_params_no_state: flash attn = 0
whisper_init_with_params_no_state: gpu_device = 0
whisper_init_with_params_no_state: dtw        = 0
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51865
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 512
whisper_model_load: n_audio_head  = 8
whisper_model_load: n_audio_layer = 6
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 512
whisper_model_load: n_text_head   = 8
whisper_model_load: n_text_layer  = 6
whisper_model_load: n_mels        = 80
whisper_model_load: ftype         = 1
whisper_model_load: qntvr         = 0
whisper_model_load: type          = 2 (base)
whisper_model_load: adding 1608 extra tokens
whisper_model_load: n_langs       = 99
ggml_vulkan: Found 1 Vulkan devices:
Vulkan0: AMD Radeon RX 5700 XT (AMD proprietary driver) | uma: 0 | fp16: 1 | warp size: 64
whisper_model_load: AMD Radeon RX 5700 XT total size =   147.37 MB
whisper_model_load: model size    =  147.37 MB
whisper_backend_init_gpu: using Vulkan backend
whisper_init_state: kv self size  =   18.87 MB
whisper_init_state: kv cross size =   18.87 MB
whisper_init_state: kv pad  size  =    3.15 MB
whisper_init_state: compute buffer (conv)   =   16.26 MB
ggml_vulkan: Device memory allocation of size 130368000 failed.
ggml_vulkan: vk::Device::allocateMemory: ErrorOutOfDeviceMemory
ggml_gallocr_reserve_n: failed to allocate AMD Radeon RX 5700 XT buffer of size 130368000
```

[...]

Issue #2411, opened 2024-09-09
Fallback from Vulkan to CPU


thewh1teagle:
Vulkan has a lot of bugs on Windows / Linux. but when it works, it works much faster than CPU. (10-20x faster)
I'm forced to use Vulkan in the project [vibe](https://github.com/thewh1teagle/vibe) but many users report that it's crash on Windows / Linux.

Some of the errors:

PopOS
https://github.com/thewh1teagle/vibe/issues/269

Ubuntu
```console
ggml_vulkan: Found 1 Vulkan devices:
Vulkan0: Intel(R) HD Graphics 620 (KBL GT2) (Intel open-source Mesa driver) | uma: 1 | fp16: 1 | warp size: 32
2024-09-09T10:58:08.692125Z ERROR whisper_rs::whisper_sys_tracing: whisper_model_load: ERROR not all tensors loaded from model file - expected 947, got 3
2024-09-09T10:58:08.711251Z ERROR whisper_rs::whisper_sys_tracing: whisper_init_with_params_no_state: failed to load model
```

Arch
https://github.com/thewh1teagle/vibe/issues/267

Windows
https://github.com/thewh1teagle/vibe/issues/266

https://github.com/thewh1teagle/vibe/issues/263

Windows

```console
ggml_gallocr_reserve_n: reallocating NVIDIA GeForce GT 730 buffer from size 0.00 MiB to 565.06 MiB
ggml_vulkan: Device memory allocation of size 592512000 failed.
ggml_vulkan: vk::Device::allocateMemory: ErrorOutOfDeviceMemory
ggml_gallocr_reserve_n: failed to allocate NVIDIA GeForce GT 730 buffer of size 592512000
```

thewh1teagle:
@ggerganov 

Do you have any suggestions on how we can improve the stability of ggml and whisper.cpp to reduce crashes (aborts) and ensure they consistently return errors instead?

ggerganov:
Hm, I haven't tested the Vulkan backend with `whisper.cpp` at all, so cannot recommend any way to improve the stability. But looking at the error - this seems like its trying to load an invalid mode, no?

The other error seems like the GPU device runs out of memory. I think your application can check if there is enough available memory before trying to load the Whisper model.

thewh1teagle:
@ggerganov 

[...]

PR #2451, opened 2024-10-04
Retry allocation with fallback flags

SRHMorris:
This fixes issue #2450 

I'm unsure what the exact cause is as I'm not overly familiar with Vulkan. So feel free to reject if you can think of a better solution.


ggerganov:
@0cc4m What do you think about this patch? Should we merge it upstream?

0cc4m:
@ggerganov Yeah, this should go upstream. I haven't tested it yet, but it looks good. It might even fix https://github.com/ggerganov/llama.cpp/pull/9734.

ggerganov (approved):
I just tested it and it also fixes the issue reported in https://github.com/ggerganov/whisper.cpp/issues/2411#issuecomment-2395272966

PR #2506, opened 2024-10-24
ruby: New segment callback

KitaitiMakoto:
Hello, again,

This is rewriting of #2495, which is attempted to add new segment callback feature.

I added methods represent the same to original API as possible such as `Whisper::Params#new_segment_callback=`, etc:

```ruby
params.new_segment_callback = ->(context, _state, n_new, user_data) {
  # ...
}
params.new_segment_callback_user_data = {some: "data"}

```
(`_state` is not implemented because `context` can do the same thing and state implementation brings boring memory management, potentially enbugs.)

Additionally, I added "high level API" such as `Whisper::Segment` class, `Whisper::Context#on_new_segment` and so on on top of APIs above:

```ruby
params.on_new_segment do |segment|
  puts segment.start_time
  puts segment.end_time
  puts segment.text
end

```

Can you consider it? Thanks.

### DIFF
--- a/.github/workflows/bindings-ruby.yml
+++ b/.github/workflows/bindings-ruby.yml
@@ -16,6 +16,9 @@ on:
       - ggml/src/ggml-quants.h
       - ggml/src/ggml-quants.c
       - ggml/src/ggml-cpu-impl.h
+      - ggml/src/ggml-metal.m
+      - ggml/src/ggml-metal.metal
+      - ggml/src/ggml-blas.cpp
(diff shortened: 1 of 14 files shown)


ggerganov (approved):
Thanks. Again, without understanding the details, the implementation seems OK to me, so happy to merge it. Please resolve the conflict and will proceed with the merge.

KitaitiMakoto:
I have rebased. Waiting for GitHub actions complete.

PR #2511, opened 2024-10-25
ci : fix openblas build

### LINKED ISSUE #2508: cmake option `GGML_OPENBLAS` not working
Running cmake with `-DGGML_OPENBLAS` raises a CMake warning "Manually-specified variables were not used by the project: GGML_OPENBLAS" and indeed it does not build with OpenBLAS as reported at runtime with the main example.

You can see it in the latest [CI workflow build log](https://github.com/ggerganov/whisper.cpp/actions/runs/11479618582/job/31946312905#step:6:52).

I see in `ggml/CMakeLists.txt` `option(GGML_BLAS...`. I will try that and report back.

...No, cmake `find_package` is not working, which does not surprise me.  This is why the msys2 [`PKGBUILD`](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-whisper.cpp/PKGBUILD) uses good-old reliable `pkg-config`:
```
CFLAGS+=" $(pkgconf.exe --cflags openblas)" \
...
      -DBLAS_LIBRARIES="`pkgconf --libs-only-l openblas`" \
      -DBLAS_INCLUDE_DIRS="`pkgconf --cflags-only-I openblas`" \
```

ggerganov:
fix #2508

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -278,7 +278,6 @@ jobs:
             mingw-w64-${{matrix.env}}-cmake
             mingw-w64-${{matrix.env}}-SDL2
             mingw-w64-${{matrix.env}}-openblas
-            mingw-w64-${{matrix.env}}-pkgconf
 
       - name: Build using make
         shell: msys2 {0}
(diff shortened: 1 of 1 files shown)


ggerganov:
Possibly, though I'm not sure how to modify the workflow file to apply this.

PR #2515, opened 2024-10-27
When DTW timestamps are enabled, defer new_segment_callback until after DTW compute step

jettoblack:
I'm using the whisper.net wrapper for .NET which uses the new_segment_callback to get all results from whisper.cpp. I'm extending it to support enabling DTW timestamps which I find to be useful. However, new_segment_callback gets called on segments before DTW timestamps are computed, so I can't get the DTW timestamps this way.

This commit changes it so that when dtw_token_timestamps is true, new_segment_callback are deferred until after DTW timestamps are computed. This change has no effect except when dtw_token_timestamps is true AND new_segment_callback is not null.

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -6198,7 +6198,7 @@ int whisper_full_with_state(
                                     n_new = whisper_wrap_segment(*ctx, *state, params.max_len, params.split_on_word);
                                 }
                             }
-                            if (params.new_segment_callback) {
+                            if (params.new_segment_callback && !ctx->params.dtw_token_timestamps) {
                                 params.new_segment_callback(ctx, state, n_new, params.new_segment_callback_user_data);
                             }
                         }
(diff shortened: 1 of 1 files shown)


ggerganov (approved):
I wonder if it makes sense to move the `new_segment_callback` at the end of the decode loop, regardless if DTW is on or off. What do you think - do you see any problems with it?

jettoblack:
> I wonder if it makes sense to move the `new_segment_callback` at the end of the decode loop, regardless if DTW is on or off. What do you think - do you see any problems with it?

I considered that but I wanted to limit the scope of the change just in case. It might increase latency when doing streaming transcription with a small window? I'm not sure if anyone would notice..

ggerganov:
> It might increase latency when doing streaming transcription with a small window?

Hm, yes. I guess it is fine as proposed. Thanks 👍

Issue #2521, opened 2024-10-29
Crash if --processors > 8


jrosener:
I am writing an application that is able to transcribe multiple audio in parallel using the same model.
For that I use one common `whisper_context` for multiple `whisper_state` used by worker threads where transcriptions processing are performed with `whisper_full_with_state()`.
It works perfectly until 8 parallel transcriptions but crashes into `whisper_full_with_state()` if running more transcriptions.

Because this implementation is based on `whisper_full_parallel()` used by the `main` sample application, it is possible to reproduce the issue by running it using more than 8 `--processors`: `./build/bin/main --model ggml-tiny.bin --processors 9 10min_audio_french.wav`
It does not matter if it is running on cpu, openvino, cuda,...: it always crashes.

Results:
```
$ ./build/bin/main --model ggml-tiny.bin --processors 9 10min_audio_french.wav 
whisper_init_from_file_with_params_no_state: loading model from 'ggml-tiny.bin'
whisper_init_with_params_no_state: use gpu    = 1
whisper_init_with_params_no_state: flash attn = 0
whisper_init_with_params_no_state: gpu_device = 0
whisper_init_with_params_no_state: dtw        = 0
whisper_model_load: loading model
whisper_model_load: n_vocab       = 51865
whisper_model_load: n_audio_ctx   = 1500
whisper_model_load: n_audio_state = 384
whisper_model_load: n_audio_head  = 6
whisper_model_load: n_audio_layer = 4
whisper_model_load: n_text_ctx    = 448
whisper_model_load: n_text_state  = 384
whisper_model_load: n_text_head   = 6
whisper_model_load: n_text_layer  = 4
whisper_model_load: n_mels        = 80
whisper_model_load: ftype         = 1
whisper_model_load: qntvr         = 0
whisper_model_load: type          = 1 (tiny)
whisper_model_load: adding 1608 extra tokens
whisper_model_load: n_langs       = 99
whisper_model_load:      CPU total size =    77.11 MB
whisper_model_load: model size    =   77.11 MB
whisper_init_state: kv self size  =    3.15 MB
whisper_init_state: kv cross size =    9.44 MB
whisper_init_state: kv pad  size  =    2.36 MB
whisper_init_state: compute buffer (conv)   =   13.19 MB
whisper_init_state: compute buffer (encode) =   64.79 MB
whisper_init_state: compute buffer (cross)  =    3.88 MB
whisper_init_state: compute buffer (decode) =   95.89 MB

[...]

PR #2525, opened 2024-10-30
whisper : reduce ggml_context usage

ggerganov:
ref https://github.com/ggerganov/whisper.cpp/issues/2521

No need to retain the `ggml_context` instances for the KV caches. With this change, each `whisper_state` now uses 3 less contexts. It still uses 4 though - one for each of the 4 schedulers.

Also, `ggml_init` will now allocate the contexts on the heap instead of using a static pool. Added `ggml_reset(ctx)` for resetting existing contexts.

### DIFF
--- a/ggml/src/ggml.c
+++ b/ggml/src/ggml.c
@@ -308,6 +308,7 @@ void ggml_abort(const char * file, int line, const char * fmt, ...) {
 }
 
 #define GGML_DEBUG 0
+
 #define GGML_GELU_FP16
 #define GGML_GELU_QUICK_FP16
 
@@ -1985,7 +1986,7 @@ static const size_t GGML_OBJECT_SIZE = sizeof(struct ggml_object);
 
 struct ggml_context {
     size_t mem_size;
-    void* mem_buffer;
+    void * mem_buffer;
     bool   mem_buffer_owned;
     bool   no_alloc;
     bool   no_alloc_save; // this is used to save the no_alloc state when using scratch buffers
@@ -3234,7 +3235,6 @@ struct ggml_numa_nodes {
 //
 
 struct ggml_state {
-    struct ggml_context_container contexts[GGML_MAX_CONTEXTS];
     struct ggml_numa_nodes numa;
 };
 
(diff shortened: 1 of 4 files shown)


slaren:
The number of contexts is likely an issue in llama.cpp as well, especially on systems with multiple GPUs. The model loader, cache, loras, control vectors all use a different context for each buffer type. It may be better to remove the limit entirely and allocate them dynamically, maybe with an option to reset the context if allocations during inference are absolutely unacceptable.

ggerganov:
@slaren What do you think about [`2d9c313` (#2525)](https://github.com/ggerganov/whisper.cpp/pull/2525/commits/2d9c313b49b995710b13049e43a6cfb7b2720e38)? `ggml_init` will now use the statically allocated contexts, and when they run out, it will start allocating on the heap.

> maybe with an option to reset the context if allocations during inference are absolutely unacceptable.

I'm not sure if avoiding these allocations during inference is really that important. So if this looks overly-complicated, I'm open to completely switch to dynamically allocated contexts and simplify this logic.

slaren:
I wouldn't be surprised in `malloc` is faster in most of cases than what `ggml_init` does iterating over a list to find a free slot. It's a very small allocation, so it will likely find a free block very quickly, without having to request more memory from the OS. So I think it would be better to remove the old logic entirely, but it's not very important either.

ggml/src/ggml.c
```diff
@@ -3839,24 +3834,11 @@ struct ggml_context * ggml_init(struct ggml_init_params params) {
...
-            g_state.contexts[i].used = true;
-            ctx = &g_state.contexts[i].context;
-
-            GGML_PRINT_DEBUG("%s: found unused context %d\n", __func__, i);
-            break;
-        }
-    }
+    ggml_critical_section_end();
 
+    struct ggml_context * ctx = GGML_ALIGNED_MALLOC(sizeof(struct ggml_context));
```
slaren, line 3839: `GGML_ALIGNED_MALLOC` is likely completely overkill here, I don't think there is a good reason to require more than the alignment of `malloc`. Additionally, `GGML_ALIGNED_MALLOC` uses `vm_allocate` now on macOS, and this may actually be much slower than `malloc`, since it is requesting pages from the OS.

slaren (approved):
This may be tricky to do correctly when freeing, but it may also make sense to use `malloc` for the buffer for `no_alloc` contexts, since these should be small and only need need to allocate tensors. It doesn't matter too much however since applications can provide their own buffer. At some point we will probably want to remove support for allocating tensors on a `ggml_context`, or just allocate the tensor data dynamically, and then that would be a good moment to implement that.

slaren:
On a side note related to this, I was planning to implement support for automatically growing the buffers of `ggml_context`, so if there is not enough space to create a tensor, it would allocate a new buffer instead of failing. I think that would be good to improve the usability of ggml, especially for newcomers or when experimenting.

ggerganov:
Yes, these sound like good improvements. Allocating tensor data in the context is not really useful and applications that really need it can implement their own memory pool if necessary.

PR #2534, opened 2024-11-06
passing samples_padded by ref to the threads.

vinmisra:
I observed substantial memory usage during multithreaded MFCC processing of large audio files. This appears to originate from copying the full `samples_padded` audio sequence to each MFCC processing thread.

This minor optimization switches to a cref to avoid copying. MFCC extraction doesn't modify the samples so I believe this is safe.


ggerganov (approved):
Oh, good one! Thanks

Josscii:
@ggerganov I tested this change, and it introduce a new memory leak, after `whisper_free`, it has more memory usage than before this change.

ggerganov:
I don't see why it would cause more memory usage.

PR #2543, opened 2024-11-08
fix ffmpeg v5 build

stsydow:
Remove call to `av_register_all()` from the transcode example,
which does not exist in ffmpeg v5 anymore.

This will break ffmpeg v3, if this needs to be supported still, we  need a different approach.

### DIFF
--- a/examples/ffmpeg-transcode.cpp
+++ b/examples/ffmpeg-transcode.cpp
@@ -204,8 +204,6 @@ static int decode_audio(struct audio_buffer *audio_buf, s16 **data, int *size)
     const size_t errbuffsize = 1024;
     char errbuff[errbuffsize];
 
-    av_register_all(); // from avformat. Still a must-have call for ffmpeg v3! (can be skipped for later versions)
-
     fmt_ctx = avformat_alloc_context();
     avio_ctx_buffer = (u8*)av_malloc(AVIO_CTX_BUF_SZ);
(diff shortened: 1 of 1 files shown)


ggerganov (approved):

examples/ffmpeg-transcode.cpp
```diff
@@ -204,8 +204,6 @@ static int decode_audio(struct audio_buffer *audio_buf, s16 **data, int *size)
     const size_t errbuffsize = 1024;
     char errbuff[errbuffsize];
 
-    av_register_all(); // from avformat. Still a must-have call for ffmpeg v3! (can be skipped for later versions)
```
WilliamTambellini, line 207: naive and bad modif: breaks the ffmpeg v3 build.
Proper modif should have been:
#ifdef FFMPEG_VERSION < 5
 av_register_all();
#endif

WilliamTambellini (commented):
-1

examples/ffmpeg-transcode.cpp
stsydow: If you want this to happen just do a PR with:

```
#ifdef FFMPEG_VERSION < 4
av_register_all();
#endif
```

In anyway this is an example for the current state and meant to be simple.

I find the tone quite harsh and not inviting for help.
The caveat for ffmpeg 3.0 was clearly mentioned in the description though.
The call is deprecated since ffmpeg 4.0 in 2018.

PR #2548, opened 2024-11-10
Fix the instructions on the Ruby binding

wilsonsilva:
`#prompt` doesn't exist but `#initial_prompt` does


KitaitiMakoto:
Oops. You're right. Thanks!

PR #2550, opened 2024-11-12
Added infinite conversation feature to talk-llama

pseelam02:
Currently the kv-cache is managed like this in the talk-llama.cpp assistant file: 

Let's say you have a prompt: 1 2 3 4 and your context length is 16. You could say the context looks like this initially:

|................|

Then the prompt gets fed to the model:

|1234............|

Then the model starts generating tokens:

|1234ABCD........|

until it reaches the end of available context space:

|1234ABCDEFGHIJKL|

Once the context space is full then the voice assistant exits and sends a "failed to decode" message. The lines of code I incorporated into the talk-llama.cpp file allows for an additional parameter -inf that can be used to have an infinite conversation with the voice assistant!

### DIFF
--- a/examples/talk-llama/talk-llama.cpp
+++ b/examples/talk-llama/talk-llama.cpp
@@ -66,6 +66,7 @@ struct whisper_params {
     bool verbose_prompt = false;
     bool use_gpu        = true;
     bool flash_attn     = false;
+    bool inf = false;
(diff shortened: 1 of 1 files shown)


pseelam02:
Opened a newer pull request that further improves on this PR.

PR #2546, opened 2024-11-10
whisper.swiftui : add model download list & bench methods

jhen0409:
This PR fixes the build of whisper.swiftui, and then add model download list (from [ggerganov/whisper.cpp](https://huggingface.co/ggerganov/whisper.cpp/resolve/main) HF repo) and bench methods.

The bench method output format is following format from `scripts/bench-all.sh`:

https://github.com/ggerganov/whisper.cpp/blob/31aea563a83803c710691fed3e8d700e06ae6788/scripts/bench-all.sh#L61-L62

New Buttons:
- View Models: Navigate to model list (We can download model & load downloaded model)
- Bench (bench_full for current loaded model)
- Bench All (bench_full for downloaded models)
- Copy Logs

The memcpy and ggml_mul_mat methods are added but commented.

The demo looks like in iPhone 15 Pro:

Benchmark result in iPhone 15 Pro (A17 Pro):

memcpy and ggml_mul_mat:

```
Running memcpy benchmark
memcpy:   14.97 GB/s (heat-up)
memcpy:   21.46 GB/s ( 1 thread)
memcpy:   20.66 GB/s ( 1 thread)
memcpy:   22.52 GB/s ( 2 thread)
memcpy:   22.57 GB/s ( 3 thread)
memcpy:   22.46 GB/s ( 4 thread)
sum:    -3072001183.000000

Running ggml_mul_mat benchmark with 4 threads
  64 x   64: Q4_0     4.1 GFLOPS (128 runs) | Q4_1     3.7 GFLOPS (128 runs)
  64 x   64: Q5_0     4.5 GFLOPS (128 runs) | Q5_1     4.6 GFLOPS (128 runs) | Q8_0     4.0 GFLOPS (128 runs)
  64 x   64: F16      4.4 GFLOPS (128 runs) | F32      4.7 GFLOPS (128 runs)
 128 x  128: Q4_0    28.2 GFLOPS (128 runs) | Q4_1    25.8 GFLOPS (128 runs)
 128 x  128: Q5_0    20.4 GFLOPS (128 runs) | Q5_1    26.4 GFLOPS (128 runs) | Q8_0    28.9 GFLOPS (128 runs)
 128 x  128: F16     16.7 GFLOPS (128 runs) | F32     31.5 GFLOPS (128 runs)
 256 x  256: Q4_0    89.1 GFLOPS (128 runs) | Q4_1    84.1 GFLOPS (128 runs)
[...]

PR #2562, opened 2024-11-15
whisper.swiftui : switch Mac dest to Mac (Designed for iPad)

jhen0409:
After #2546, which used some API only for iOS like `UIKit`, `UIPasteboard` or some navigation styles, the build of the Mac target is broken.

To keep it simple, I just changed to Mac dest to Mac (Designed for iPad).


ggerganov (approved):

ggerganov:
Thank you!

PR #2569, opened 2024-11-17
docs: Update README.md for whisper.objc app

tomer-schles:
I struggled with using the objectivec iphone app with coreml, so I updated the readme to make things a bit clearer.


danbev:
Thanks for the PR and sorry about this late response. We updated the [whisper.swiftui](https://github.com/ggerganov/whisper.cpp/tree/master/examples/whisper.swiftui#core-ml-support) example with something similar to this. If you are still interested it would be great the add something similar so they are consistent.

danbev (approved):

PR #2576, opened 2024-11-20
examples : try to fix objc CI

ggerganov:


danbev:
@ggerganov Would it be alright to close this now that we have updated the objc example?

ggerganov:
Yup, closing.

PR #2577, opened 2024-11-21
openvino : Pass CPU threads parameter

jkarthic:


jkarthic:
@ggerganov Any comments?

PR #2613, opened 2024-12-07
Fix vulkan Makefile paths

viric:
Otherwise it doesn't build, paths wrong after the ggml updates.


ggerganov:
The new recommended way to build the project is with CMake. The Makefile is now a simple "shim" of CMake.

viric:
cmake didn't have vulkan.

ggerganov:
It has now: https://github.com/ggerganov/whisper.cpp#vulkan-gpu-support

Let us know if you encounter any problems with the build.

PR #2625, opened 2024-12-13
Implement no_speech_thold

jkarthic:
no_speech_thold functionality is on par with OpenAI's whisper

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -867,6 +867,7 @@ struct whisper_state {
     whisper_token tid_last;
 
     std::vector<float> energy; // PCM signal energy
+    float no_speech_prob = 0.0f;
 
     // [EXPERIMENTAL] Token-level timestamps with DTW
     whisper_aheads_masks aheads_masks;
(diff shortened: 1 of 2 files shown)


src/whisper.cpp
```diff
@@ -5647,6 +5648,35 @@ int whisper_full_with_state(
...
+
+                    // Normalize
+                    for (int i = 0; i < n_vocab; ++i) {
+                        probs[i] /= sum_exp;
+                    }
+
+                    // Get probability of no_speech token
+                    state->no_speech_prob = probs[whisper_token_nosp(ctx)];
+                }
+
```
ggerganov, line 5679: This likely has to be done inside `whisper_process_logits` in order to avoid computing the softmax again just for this probability.

src/whisper.cpp
```diff
@@ -6038,7 +6068,8 @@ int whisper_full_with_state(
             if (it != (int) temperatures.size() - 1) {
                 const auto & decoder = state->decoders[best_decoder_id];
 
-                if (decoder.failed || decoder.sequence.avg_logprobs < params.logprob_thold) {
+                if (decoder.failed ||
+                    (decoder.sequence.avg_logprobs < params.logprob_thold && state->no_speech_prob < params.no_speech_thold)) {
```
ggerganov, line 6072: - The log message is no longer correct
- The comparison for the speech prob is wrong

src/whisper.cpp
jkarthic: Unfortunately we cannot reuse the softmax computed inside `whisper_process_logits` since `no_speech_prob` has to be calculated before any logits filtering. Otherwise we get some wrong `no_speech_prob` values. The same method is followed in openai's whisper as well. https://github.com/openai/whisper/blob/main/whisper/decoding.py#L689-L703
Since this `no_speech_prob` calculation is only for the first token in the sequence, it will not cause a big performance impact.
On a related note, I have now modularized the probs calculation and now reusing the same code as `whisper_process_logits`

src/whisper.cpp
jkarthic: - Thanks. I have corrected the log message to print the `no_speech_prob` and `no_speech_thold` values as well.
- The comparison logic is on par with the openai implementation. avg_logprobs being lesser than the threshold is considered as a failure only for speech segment. If it is a non-speech, then it is considered as a successful prediction of "silence". Here is the relavant code from openai. I have just merged it into one condition by inverting the comparison. But the logic is the same.
https://github.com/openai/whisper/blob/main/whisper/transcribe.py#L209-L220

ggerganov (approved):
Thank you, good PR.

PR #2649, opened 2024-12-20
feat(server): Add option to suppress non-speech tokens

sachaarbonel:
This PR adds the ability to suppress non-speech tokens in the whisper transcription output. Users can enable this feature through:
- CLI flag: `-sns` or `--suppress-non-speech`
- HTTP request parameter: `suppress_non_speech`

This feature is useful for obtaining cleaner transcriptions by removing non-speech markers (like [LAUGH], [SIGH], etc.) from the output text.


ggerganov (approved):

tannisroot:
Thank you so much for adding this!

PR #2670, opened 2024-12-25
ruby : fix bug on `log_set` and more flexible argument

KitaitiMakoto:
Hello,

`Whisper::Context.log_set` included a bug. I fixed it. And, I made `Whisper::Context.new` accept model URI in addition to model name and path.

Thank you.


KitaitiMakoto:
Sorry, my patches are wrong.

PR #2683, opened 2024-12-30
ruby : Fix of C++ header guard name, model URI support, type signature and more

### LINKED ISSUE #2672: reserved identifier violation
I would like to point out that an identifier like “[`__RUBY_WHISPER_H`](https://github.com/ggerganov/whisper.cpp/blob/7d55637f0bc8ff1b405f6f18c5c04fe39e4ad389/bindings/ruby/ext/ruby_whisper.h#L1 "Update candidate")” [does not fit](https://wiki.sei.cmu.edu/confluence/display/cplusplus/DCL51-CPP.+Do+not+declare+or+define+a+reserved+identifier#DCL51CPP.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29 "Do not declare an identifier which is reserved for the compiler implementation.") to the expected naming convention of the C++ language standard.
Would you like to adjust your selection for unique names?

KitaitiMakoto:
Hello,

This pull request

* fixes #2672,
* makes `Whisper::Context.new` accept URI for model file,
* adds type signature file, and
* includes some slight fixes.

Thank you.

### DIFF
--- a/bindings/ruby/.gitignore
+++ b/bindings/ruby/.gitignore
@@ -1,5 +1,3 @@
 LICENSE
 pkg/
-lib/whisper.so
-lib/whisper.bundle
-lib/whisper.dll
+lib/whisper.*
--- a/bindings/ruby/README.md
+++ b/bindings/ruby/README.md
@@ -60,10 +60,10 @@ You also can use shorthand for pre-converted models:
 whisper = Whisper::Context.new("base.en")
 ```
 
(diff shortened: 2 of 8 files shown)


ggerganov (approved):

elfring:
Thanks for [your adjustment of an affected include guard](https://github.com/ggerganov/whisper.cpp/pull/2683/commits/381e2c52e7ce61400434f21f6c2140775382b2ca "Rename header guard to conform to C++ naming convention").

PR #2699, opened 2025-01-03
sync : ggml

ggerganov:


ggerganov:
I'm not sure how to fix the failing `arm64` and `arm/v7` builds. Any ideas? cc @slaren @angt

PR #2734, opened 2025-01-13
Disable GL_KHR_cooperative_matrix Vulkan extension if not available

### LINKED ISSUE #2733: Detect coopmat support
It'd be great if we could have the analog of https://github.com/ggerganov/llama.cpp/pull/11117

I'll try to put something together when I get the time if needed.

pepijndevos:
fixes #2733


ggerganov:
This is synced now on `master`.

pepijndevos:
Huh do you have some automated system to synchronise changes from llama.cpp?

ggerganov:
It's semi-automated - running Bash scripts manually: https://github.com/ggerganov/ggml/discussions/1066

pepijndevos:
Could I have another sync for https://github.com/ggerganov/llama.cpp/pull/11406 to fix a segfault in https://github.com/ggerganov/llama.cpp/issues/11308

Alternatively, do you accept PRs for synchronization or you'd rather just run it yourself?

ggerganov:
Will sync today. Ideally, I would like to change the repositories so that no synchronization is required and all of them would use a single `ggml` source. But I haven't figured out the best way to do it yet.

ggerganov:
There is something weird going on with the CI and the builds are failing - https://github.com/ggerganov/whisper.cpp/pull/2779. Will check later today if the issue disappears.

slaren:
May be the same issue that we had with the ARM docker images:
https://github.com/ggerganov/llama.cpp/pull/11434#discussion_r1932909270
https://github.com/docker/build-push-action/issues/1309

PR #2749, opened 2025-01-18
ruby : Make context accept initial parameters, API to retrieve a segment and more

KitaitiMakoto:
Hello,

This pull request

* makes `Whisper::Context.new` accept initial parameters,
* adds to retrieve to a segment to `Whisper::Context`,
* splits a C++file into multiple files,
* and includes tiny refinements

Thanks.


bindings/ruby/ext/ruby_whisper_transcribe.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+    // convert to mono, float
+    pcmf32.resize(n);
+    if (wav.channels == 1) {
+      for (uint64_t i = 0; i < n; i++) {
+        pcmf32[i] = float(pcm16[i])/32768.0f;
+      }
+    } else {
+      for (uint64_t i = 0; i < n; i++) {
+        pcmf32[i] = float(pcm16[2*i] + pcm16[2*i + 1])/65536.0f;
```
ggerganov, line 113: This can overflow `int16_t`

ggerganov (approved):

bindings/ruby/ext/ruby_whisper_transcribe.cpp
KitaitiMakoto: Thanks for reviewing! I fixed it. I appreciate if you tell me better way because I'm a newbie to C++.

ggerganov: This is OK.

KitaitiMakoto: Thanks!

PR #2756, opened 2025-01-22
Updated models download URL

AMDphreak:
https://huggingface.co/ggerganov/whisper.cpp/blob/main/

Example
https://huggingface.co/ggerganov/whisper.cpp/blob/main/ggml-large-v3-turbo.bin

Docs are out of date. AI keeps generating the wrong URL when I ask it.


bindings/go/examples/go-model-download/go-model-download.exe
ggerganov, line 1: This executable should be removed

whisper_build.log
ggerganov, line 1: Remove this log file.

bindings/go/examples/go-model-download/go-model-download.exe
AMDphreak: I didn't notice that. Why is .exe not in the gitignore? I had assumed I would be dropping .exe into all of my gitignores so I put it in my global gitignore too.

whisper_build.log
AMDphreak: Same for .log files. I assume we would want to put those in gitignore. Can I update gitignore with these both?

PR #2777, opened 2025-02-03
cmake : fix compile assumptions for power9/etc

midnightmagic:
A regression was introduced in the move to CMake and with the
patch for POWER10 support.

This patch corrects CMake build logic to correctly detect and define
-cpu=power9 during build, which turns back on all the VSX codepaths
for POWER9 machines.

Results especially with -DGGML_BLAS=1 restore whisper.cpp performance
to faster-than-realtime transcription with the medium-q5 model on a
Talos II underclocked to 3.3GHz.


ggerganov:
Could you rebase on latest `master` - I think the CI builds should work now.

README.md
```diff
@@ -14,7 +14,7 @@ High-performance inference of [OpenAI's Whisper](https://github.com/openai/whisp
 - Plain C/C++ implementation without dependencies
 - Apple Silicon first-class citizen - optimized via ARM NEON, Accelerate framework, Metal and [Core ML](#core-ml-support)
 - AVX intrinsics support for x86 architectures
-- VSX intrinsics support for POWER architectures
+- VSX intrinsics support for POWER architectures(#power-vsx-intrinsics)
```
fitzsim, line 17: Does this link work without `[...]` to indicate the link text?

fitzsim:
Without this patch on my system `jfk.wav` with `ggml-base.en.bin` took `7362.89 ms` which represents a regression from the `Makefile` build.  With this patch, performance is restored to `2614.46 ms`.  I reviewed the patch over email with @midnightmagic.  I think it looks good, other than the Markdown question I asked as a patch comment.

README.md
midnightmagic: It does not; corrected.

midnightmagic:
As requested—force-pushed a rebase on current master as of a few minutes ago.

ggerganov (approved):
Thanks, we can merge if the CI is green.

PR #2779, opened 2025-02-03
sync : ggml

ggerganov:


ggerganov:
Huh, `git` disappeared from the Ubuntu images?

https://github.com/ggerganov/whisper.cpp/actions/runs/13115281453/job/36588000327?pr=2779#step:4:1769

PR #2790, opened 2025-02-05
Fixes for Windows

foldl:
* MSVC default to utf-8 without BOM.
* Console output code page changed to utf-8.

### DIFF
--- a/cmake/build-info.cmake
+++ b/cmake/build-info.cmake
@@ -42,6 +42,7 @@ endif()
 if(MSVC)
     set(BUILD_COMPILER "${CMAKE_C_COMPILER_ID} ${CMAKE_C_COMPILER_VERSION}")
     set(BUILD_TARGET ${CMAKE_VS_PLATFORM_NAME})
+    add_compile_options("/utf-8")
 else()
     execute_process(
         COMMAND sh -c "$@ --version | head -1" _ ${CMAKE_C_COMPILER}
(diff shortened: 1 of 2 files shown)


ggerganov (approved):
Not sure why the CUDA build failed - restarted it just now.

cmake/build-info.cmake
```diff
@@ -42,6 +42,8 @@ endif()
 if(MSVC)
     set(BUILD_COMPILER "${CMAKE_C_COMPILER_ID} ${CMAKE_C_COMPILER_VERSION}")
     set(BUILD_TARGET ${CMAKE_VS_PLATFORM_NAME})
+    add_compile_options("$<$<COMPILE_LANGUAGE:C>:/utf-8>")
+    add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:/utf-8>")
```
ggerganov, line 46: Btw, which source files require this flag? I am wondering because in `llama.cpp` we don't seem to need this.

foldl: `whisper.cpp` requires this flag:

https://github.com/ggerganov/whisper.cpp/blob/46d07b9c8506246bea4f2939470acfbbc63ce675/src/whisper.cpp#L4849

There are such flags in `llama.cpp`:

https://github.com/ggerganov/llama.cpp/blob/d774ab3acc4fee41fbed6dbfc192b57d5f79f34b/CMakeLists.txt#L51

ggerganov (approved):
Thanks. Merge at will.

PR #2797, opened 2025-02-08
Update jni.c for enabling word level timestamps in whisper.cpp android

ranjitsingha:
Updated jni.c for enabling word level timestamps in whisper.cpp android


ranjitsingha:
Results
Screenshot_20250208-130520_1

ggerganov:
It should be an option, not the default behavior.

PR #2800, opened 2025-02-08
Update jni.c for enabling word timestamp in Android

ranjitsingha:
Update jni.c for enabling word level timestamps in whisper.cpp Android 

Set to optional.

Default is set to off.

If anyone wants can turn it to true for enabling word Level Timestamps

### DIFF
--- a/examples/whisper.android/app/src/main/java/com/whispercppdemo/ui/main/MainScreenViewModel.kt
+++ b/examples/whisper.android/app/src/main/java/com/whispercppdemo/ui/main/MainScreenViewModel.kt
@@ -143,7 +143,7 @@ class MainScreenViewModel(private val application: Application) : ViewModel() {
             printMessage("${data.size / (16000 / 1000)} ms\n")
             printMessage("Transcribing data...\n")
             val start = System.currentTimeMillis()
-            val text = whisperContext?.transcribeData(data)
+            val text = whisperContext?.transcribeDataWithParams(data, tokenTimestamps = false, splitOnWord = false, maxLen = 1)
             val elapsed = System.currentTimeMillis() - start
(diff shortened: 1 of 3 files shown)


foldl:
This does not solve the problem.

Maybe you can add another API, with 3 additional params: `token_timestamps`, `split_on_word` and `max_len`.

ranjitsingha:
> This does not solve the problem.
> 
> Maybe you can add another API, with 3 additional params: `token_timestamps`, `split_on_word` and `max_len`.

Ok i updated with api

examples/whisper.android/app/src/main/java/com/whispercppdemo/ui/main/MainScreenViewModel.kt
```diff
@@ -143,7 +143,7 @@ class MainScreenViewModel(private val application: Application) : ViewModel() {
             printMessage("${data.size / (16000 / 1000)} ms\n")
             printMessage("Transcribing data...\n")
             val start = System.currentTimeMillis()
-            val text = whisperContext?.transcribeData(data)
+            val text = whisperContext?.transcribeDataWithParams(data, tokenTimestamps = false, splitOnWord = false, maxLen = 1)
```
foldl, line 146: Is this `transcribeDataWithParams` or `transcribeData`? I can't find the definition of `transcribeDataWithParams`, but `transcribeData` is updated with new parameters.

ranjitsingha: Sorry i mistakely used `transcribeDataWithParams` insted of `transcribeData`
I have updated it now

ranjitsingha:
What happened here

PR #2821, opened 2025-02-16
Added WER 

harvestingmoon:
Made a WER script that is used to calculate WER via the Google's Voice Commands Dataset


harvestingmoon:
ignore this pull request, will create another script for this

PR #2816, opened 2025-02-14
Restore big endian support

fitzsim:
Restore big endian support, which worked with the `Makefile`, to `CMake`, eliminate `C++23` reliance, fix some whitespace.  It might make sense to move this logic into `ggml` at some point.

PR #2822, opened 2025-02-16
Adjusted stream example to stop on ^C when no audio is received.

petterreinholdtsen:
Add check for ctrl-c in potentially endless loop while calling audio.get() to receive sound.


petterreinholdtsen:
CI on arm64 seem to be completely broken, similar on #2816.  The failing checks do not have anything to do with my code change.

ggerganov (approved):

PR #2824, opened 2025-02-17
new WER script

harvestingmoon:
WER testing based off speaker 6097 of the HiTTS Dataset.  Audio carried is ~ 10mb and contains dozens of short 10 second audio. WER_Scripting.py would then calculate the WER via DP.


foldl:
Are those Java class files are redundant?

test.py
foldl, line 1: This file is not used.

wer_testing/Readme.md
```diff
@@ -0,0 +1,16 @@
+# WER Testing For Whisper Cpp 
+
+## Instructions: (Take note, this must be only after you have downloaded and made the model, else you can reference to either YT or the main dir)
+
+1. Place the WER_Script in the same directory as main directory
+2. Place the folder ``` 6097_5_min ``` as the same directory as whisper
+3. Run WER_Script.py
```
foldl, line 7: Can you re-organize directory structure so that step 1 & 2 can be eliminated?

wer_testing/Readme.md
foldl, line 1: To be consistent with other files: `Readme.md` -> `README.md`

wer_testing/WER_Script.py
```diff
@@ -0,0 +1,173 @@
...
+import contextlib
+import argparse
+import json
+
+# Custom action to handle comma-separated list
+class ListAction(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        setattr(namespace, self.dest, [int(val) for val in values.split(",")])
+
+
```
foldl, line 15: Please delete extra blank lines. A single blank line is OK. And elsewhere.

test.py
harvestingmoon: Fixed under new script

wer_testing/Readme.md
harvestingmoon: Removed

harvestingmoon:
have also added some preview to cli:

``` printing out command:
../build/bin/whisper-cli -m ../models/ggml-base.en.bin -t 4 -p 1 -f ./6097_5_mins/audio/presentpictureofnsw_02_mann_0083.wav
Word transcribed is : ['and all boats to be moored within the hospital waff and hulk.']
Actual word is: and all boats to be moored within the hospital wharf and hulk
wer for audio/presentpictureofnsw_02_mann_0083.wav is 0.17 ```

this is what will be viewed when it is iterating every audio loop

foldl:
> * The audio files should not be committed in the repo, but should be downloaded instead

Downloadable datasets are huge with several hundreds of hours of audio. While I think this work item is to create some light-weight tests on WER performance which can be integrated into github workflows.

If there is a tiny dataset contained in repo, then WER benchmarking will work just out of the box and on-the-fly. A selected tiny dataset can also cover English/non-English, clean or noisy, which will be handy & useful. For example, if some sort of noise cancellation is added, then some noisy audio files can be added to the dataset and get benchmark easily & quickly.

Anyway, I suggested to add audio files to this repo. My apologies to @harvestingmoon.

harvestingmoon:
> * Should handle short audio inputs (i.e. less than 1s)

I have tried short audio inputs with the Google Command Dataset which contains audio input files approx 1s each. However, the problem with this is that whisper.cpp is unable to capture any words at all (I believe it is because the audio inputs are just too short) so there is difficulty in calculating WER. Hence, I switched over to the HiTTS dataset. 

Also, no worries @foldl ! am glad to try to help / contribute 😄

PR #2832, opened 2025-02-22
ci : fix (again) arm64 build fails

fitzsim:
* docker : attempt fixing arm64 build on ci

* qemu v7.0.0-28

[imported from
https://github.com/ggml-org/llama.cpp
/commit/818a340ea8be55b3706e1772527cb8738e90a8c7
(#11895)]

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,6 +30,8 @@ jobs:
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
+        with:
+          image: tonistiigi/binfmt:qemu-v7.0.0-28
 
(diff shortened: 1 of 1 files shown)


fitzsim:
Submitted per the discussion in #2816.  I tried pushing this patch on that pull request, but it did not take; I suspect workflow changes need to be reviewed/approved/merged before taking effect, so I submitted this separately.  I will rebase the big endian branch if this goes in.

foldl:
I have tested the change from llama.cpp. It doesn't solve the problem. @fitzsim You can test it on your own fork.

ggerganov:
I think we have to apply it also to `.github/workflows/build.yml`

PR #2838, opened 2025-02-24
Add C language binding for grammar-parser to core.

philipag:
The main idea is to create a C facade for grammar loading so that it can easily be use to load/allocate (and later free) grammars by any type of language binding.

This facade is extracted from a different project and not yet part of the whisper.cpp build.


philipag:
I now realize that this really belongs in ggml so someone familiar with ggml build files could repurpose this code to there...

PR #2840, opened 2025-02-24
Update convert-h5-to-ggml.py

fltman:
improved handling of missing max_length


fltman:
fix errors such as:

convert-h5-to-ggml.py", line 119, in <module> fout.write(struct.pack("i", hparams["max_length"])) struct.error: required argument is not an integer (base)

PR #2842, opened 2025-02-26
Install some example programs in user PATH.

petterreinholdtsen:
This allow the key programs whisper-bench, whisper-quantize, whisper-server and whisper-talk-llama to be available for users in addition to the shared library.

### DIFF
--- a/examples/bench/CMakeLists.txt
+++ b/examples/bench/CMakeLists.txt
@@ -5,4 +5,4 @@ include(DefaultTargetOptions)
 
 target_link_libraries(${TARGET} PRIVATE whisper ${CMAKE_THREAD_LIBS_INIT})
 
-install(TARGETS ${TARGET} RUNTIME)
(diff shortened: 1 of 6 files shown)


NeoShadow890 (approved):

petterreinholdtsen:
I resolved the conflict and updated the patch to adjust the new and existing install statements instead of introducing them.

foldl:
I don't like this. Why pollute users' path?

petterreinholdtsen:
[Judd]
> I don't like this. Why pollute users' path?

Providing useful programs to users is to me making it easier in a
multiuser installation for users to find the tool when the system
administrator installed it.  I thus believe your framing ("pollute") is
not useful.

-- 
Happy hacking
Petter Reinholdtsen

foldl:
There are ones who like to **install** things into their systems. There are also ones who like portable programs. All have their own tastes and preferences. The point is that if things have been done in one way, then it is not necessary to be changed into another way.

PR #2843, opened 2025-02-26
whisper : support GGML_BACKEND_DL

slaren:


slaren:
I didn't notice any issues with the basic functionality, but I am not sure how to test the DTW stuff. The CI failures seem unrelated.

ggerganov (approved):
The DTW feature hasn't been exercised for a long time and it seems it broke at some point. The original command that used to work when the feature was added was:

```bash
./main -m models/ggml-small.bin -f samples/gb0.wav -dtw small -ojf
```

https://github.com/ggerganov/whisper.cpp/pull/1485#issuecomment-2009985308

Pinging @denersc if they can take a look after merging. But I think it is not super important, since the feature is marked as experimental anyway and I am not sure if it got any significant usage.

slaren:
I had introduced issues in the DTW code, but it should be fixed now.

The `ios-xcode-build` CI is failing, it looks like that's because I added `ggml-cpp.h` header in `whisper.cpp`, and it has its own build system, but I don't know how to fix that.

The Ruby CI is also failing, but don't know why or if it is related to the changes here.

ggerganov:
I am not sure why the Ruby workflow is failing too, but it's ok to ignore for now. @KitaitiMakoto will probably take a look later.

PR #2759, opened 2025-01-23
Use miniaudio for direct decoding flac, mp3, ogg and wav

data-man:
`miniaudio.h` taken from https://github.com/mackron/miniaudio/commit/6d5efde254a81e42294f3baf4a4376a76492c192. And `stb_vorbis.c` from https://github.com/nothings/stb/commit/5c205738c191bcb0abc65c4febfa9bd25ff35234.
I think `miniaudio` can be used in many other purposes in the future.
Note: tested in Linux only.

Thank you for the awesome project!

PR #2851, opened 2025-02-27
ruby : follow audio library change

KitaitiMakoto:
Hello,

This pull request makes Ruby bindings follow audio library change introduced at #2759 and then makes CI fixed.

Thanks.

### DIFF
--- a/.github/workflows/bindings-ruby.yml
+++ b/.github/workflows/bindings-ruby.yml
@@ -19,7 +19,12 @@ on:
       - ggml/**/*.m
       - ggml/**/*.metal
       - scripts/get-flags.mk
-      - examples/dr_wav.h
+      - examples/common.h
+      - examples/common.cpp
+      - examples/common-whisper.h
+      - examples/common-whisper.cpp
+      - examples/stb_vorbis.c
+      - examples/miniaudio.h
   pull_request:
     paths:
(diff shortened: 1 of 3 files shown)


ggerganov (approved):
Btw, it would be better if the Ruby bindings start using CMake to build the ggml/whisper.cpp, instead of replicating the build with Makefiles.

KitaitiMakoto:
I think so, though it will take several weeks for me to learn CMake.

ggerganov:
Just to clarify, I mean that most likely the Makefile that you generate can be replaced by a single call to a `cmake` command. Not that you need to write CMake scripts - just use the existing CMake build system for the project.

KitaitiMakoto:
Ah, I see. Thanks!

houyuanjie:
> I think so, though it will take several weeks for me to learn CMake.

Hi @KitaitiMakoto, 
I'm sorry if I bothered you, please take a look at this template repository [houyuanjie/learn_cmake_ruby](https://github.com/houyuanjie/learn_cmake_ruby), it may give you some help.

KitaitiMakoto:
@houyuanjie 
Great! I didn't know
```ruby
spec.extensions = ["ext/learn_cmake_ruby/CMakeLists.txt"]
```
is effective. Thank you!

PR #2853, opened 2025-02-28
add wip wyoming cli

### LINKED ISSUE #2738: Wyoming / stdio server
I have been looking at https://github.com/rhasspy/wyoming-whisper-cpp/ which is used for Home Assistant speech to text. It uses the deprecated `main` binary to pipe audio samples into whisper.cpp.

It appears that `whisper-cli` does not retain this functionality, and all the examples I've seen of eg piping ffmpeg into whisper.cpp use this main binary.

What would be the correct approach to implement a Wyoming server?

It does appear there is a https://github.com/ser/wyoming-whisper-api-client proxy server which seems a bit wasteful but might work.

pepijndevos:
Should fix #2738 once it's tested, polished, and approved

I basically applied https://gist.github.com/jaharon/b18536bd28455ab961ce5e13712f30d9 and deleted all the code for reading and writing various types of files.

So what this does is basically read a wyoming json+wav chunk, and spit out text

Next my plan is to update https://github.com/rhasspy/wyoming-whisper-cpp to build whisper.cpp as a submodule and then test it on my raspberry pi.


pepijndevos:
I have confirmed this works in https://github.com/rhasspy/wyoming-whisper-cpp/pull/4

If there is no interest in having it upstream, another option would be to build it as a standalone binary downstream, but I wasn't able to figure out how to reuse the whisper.cpp cmake infrastructure to build an out of tree example with all the right includes and libraries.

ggerganov:
It seems that the new example that you propose is very similar to the existing `cli` example. Maybe it would be simpler to add the changes directly in the `cli` example?

pepijndevos:
I tried that initially, but the way each chunk of json+wav on stdin is essentially processed as a separate file seemed a bit at odds with how the cli example works. So it seemed simpler to split off the example than to complicate the cli option parsing and main loop.

ggerganov:
I see. I think you can introduce a CLI flag `-wyoming` that if provided, then the input loop will switch from files to json+wav.

I would like to avoid having a new example that is so similar to an existing one - it will be more difficult to maintain. Let me know if the above propose would work.

pepijndevos:
Sure I can give it a go.

Another alternative is to just maintain this as part of the python package but I'd have to figure out the build system

ggerganov:
Yes, it would make more sense. Note that the example apart from using `libwhisper`, also uses functionality from `common` which you would need to re-implement downstream. It's mostly utility stuff such as argument parsing, grammar (which probably you don't need), etc. Let me know if you encounter any problems.

PR #2876, opened 2025-03-13
Fix linking of examples on Ubuntu 20

WilliamTambellini:
Just link some examples with DL (dlopen, dlclose, ...).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,8 @@
 
 build/
 build-*/
+debug/
+release/
 
 # SPM
 .build/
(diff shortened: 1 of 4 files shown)


danbev:
@WilliamTambellini I've merged #2875 which does the same as your solution here. but does this for the `libcommon`. I used your solution of using `CMAKE_DL_LIBS` which I was not aware of and was much nicer that what I originally had :+1:   

Sorry if this wasted some of your time and please open an new PR for the other changes in this PR if you like, and I'll make sure to review it. Thanks!

WilliamTambellini:
Sure tks

danbev:
Closed in favor of #2875.

PR #2914, opened 2025-03-20
Fixed cooperative matrix shaders on Vulkan builds when CROSS_COMPILE ON

sandrohanea:
# Cross-Compilation System for GGML Vulkan

### glslc Executable Handling:

 - The system now finds the glslc executable on the host system first (before loading Vulkan) using `find_program(Vulkan_GLSLC_EXECUTABLE glslc NO_CMAKE_FIND_ROOT_PATH)`
 - This ensures that we use a host version of glslc even in cross-compilation scenarios
 - Users can override this by setting the Vulkan_GLSLC_EXECUTABLE cache variable directly (`-DVulkan_GLSLC_EXECUTABLE=/path/to/glslc`)

### Cooperative Matrix Support Detection:

 - In normal builds, the system tests if the host glslc supports cooperative matrices by compiling test shaders
 - In cross-compilation, it makes intelligent guesses based on the target platform:
      - Android targets default to COOPMAT=ON and may enable COOPMAT2 for newer API/NDK versions
      - ARM64 platforms default to COOPMAT=ON but keep COOPMAT2=OFF
      - Other platforms default to both OFF
  - These defaults can be overridden with explicit CMake options:
     - `-DGGML_VULKAN_COOPMAT_GLSLC_SUPPORT=ON `
     -  `-DGGML_VULKAN_COOPMAT2_GLSLC_SUPPORT=ON`
     
### Shader Compilation Process:
  - During cross-compilation, the build system creates a host-native version of the vulkan-shaders-gen tool
  - This tool receives the glslc path from the main CMake and embeds it as a default at compile-time
  - The tool can still accept a runtime override via the --glslc command-line parameter
  - This approach ensures that shader compilation always uses a compatible host glslc
    
### Android NDK Compatibility:
  - The system now properly handles the case where different Android NDKs might have different glslc versions
  - Users can explicitly select which glslc to use, allowing them to pick a newer version if needed
  - This solves potential compatibility issues with older NDK glslc versions
  
  
 ## CI (in Whisper.net)
 
 As whisper.cpp doesn't have some CI for Vulkan, I'm attaching my CI runs for whisper.net (only building, no tests yet :( ):
 
 With Cross-Compile:
 https://github.com/sandrohanea/whisper.net/actions/runs/14092352637/job/39471912246
 
 Without Cross-Compile:
 https://github.com/sandrohanea/whisper.net/actions/runs/14092371745/job/39471978181

### DIFF
--- a/ggml/src/ggml-vulkan/CMakeLists.txt
+++ b/ggml/src/ggml-vulkan/CMakeLists.txt
@@ -1,15 +1,15 @@
 cmake_minimum_required(VERSION 3.19)
 cmake_policy(SET CMP0114 NEW)
 
-find_package(Vulkan COMPONENTS glslc REQUIRED)
-
-# Allow explicitly specifying glslc executable for cross-compilation scenarios
-if(DEFINED GGML_VULKAN_GLSLC_EXECUTABLE)
-    set(VULKAN_GLSLC_EXECUTABLE_OVERRIDE ${GGML_VULKAN_GLSLC_EXECUTABLE})
-    message(STATUS "Using explicitly provided glslc: ${VULKAN_GLSLC_EXECUTABLE_OVERRIDE}")
-else()
-    set(VULKAN_GLSLC_EXECUTABLE_OVERRIDE ${Vulkan_GLSLC_EXECUTABLE})
+# First find glslc on the host system (not target) to ensure we have control over which one is used
+find_program(Vulkan_GLSLC_EXECUTABLE glslc NO_CMAKE_FIND_ROOT_PATH)
+if(NOT Vulkan_GLSLC_EXECUTABLE)
+    message(WARNING "glslc not found on host system. Will attempt to find via Vulkan package")
 endif()
+
+# Find Vulkan package with required components
+find_package(Vulkan REQUIRED)
+message(STATUS "Using glslc: ${Vulkan_GLSLC_EXECUTABLE}")
 
 function(detect_host_compiler)
     if (CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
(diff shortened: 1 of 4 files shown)


slaren:
@0cc4m @jeffbolznv @netrunnereve Can you take a look at this?

ggml/src/ggml-vulkan/vulkan-shaders/CMakeLists.txt
```diff
@@ -4,8 +4,22 @@ if(NOT GLSLC_EXECUTABLE)
...
 
+option(GGML_VULKAN_COOPMAT_GLSLC_SUPPORT "Enable coopmat shaders" OFF)
+option(GGML_VULKAN_COOPMAT2_GLSLC_SUPPORT "Enable coopmat2 shaders" OFF)
+
+message(STATUS "GGML_VULKAN_COOPMAT_GLSLC_SUPPORT: ${GGML_VULKAN_COOPMAT_GLSLC_SUPPORT}")
+message(STATUS "GGML_VULKAN_COOPMAT2_GLSLC_SUPPORT: ${GGML_VULKAN_COOPMAT2_GLSLC_SUPPORT}")
+
 set(TARGET vulkan-shaders-gen)
 add_executable(${TARGET} vulkan-shaders-gen.cpp)
+if (GGML_VULKAN_COOPMAT_GLSLC_SUPPORT)
```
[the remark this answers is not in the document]
sandrohanea: I'm not 100% sure if we need additional version to be used (and in my case it was working as same version as the host system was used). I can confirm that only providing the compile definitions is working for my build => https://github.com/sandrohanea/whisper.net/actions/runs/13977409227/job/39134498314

ggml/src/ggml-vulkan/CMakeLists.txt
```diff
@@ -22,33 +22,54 @@ if (Vulkan_FOUND)
                              ggml-vulkan.cpp
                              ../../include/ggml-vulkan.h
                             )
+    if (CMAKE_CROSSCOMPILING)
+        # Make this configurable for cross builds
```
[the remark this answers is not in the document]
jeffbolznv: Does it work if you remove this condition? i.e. detect whether it's supported and pass it to the external project?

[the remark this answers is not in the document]
sandrohanea: Yes, that should work:
 - if we remove this condition
 - detect in the main project if we have the COOPMAT and which version
 - pass it to the `vulkan-shaders-gen` external project
 - in the CMakeLists for that one, we should still check the option and call `target_compile_definitions`
 
 The reason why I didn't use this approach was that I was thinking that it is clearer to not check the build host for this version when CROSS_COMPILING as we're not building for this system => but indeed, if this system won't have that specific version, we won't be able to build anyway.
 
 I'm happy to simplify it like that if you think it would be better :)

ggml/src/ggml-vulkan/CMakeLists.txt
```diff
@@ -48,6 +50,7 @@ if (Vulkan_FOUND)
         message(STATUS "GL_NV_cooperative_matrix2 not supported by glslc")
```
jeffbolznv, line 50: Should there be an "OFF" here like there is for coopmat1 above?

ggml/src/ggml-vulkan/CMakeLists.txt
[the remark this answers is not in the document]
sandrohanea: Link to CI in Whisper.net after the change: https://github.com/sandrohanea/whisper.net/actions/runs/14087208133/job/39454313299

ggml/src/ggml-vulkan/CMakeLists.txt
sandrohanea: Good catch!
I got too fast on it, thanks for catching it!

It would not be consistent otherwise => won't cause issues as default will be OFF anyway, but I think it is better to have it explicit.

sandrohanea:
CI in Whisper.net for:
 - https://github.com/sandrohanea/whisper.net/actions/runs/14087392764/job/39454949037 => CMAKE_CROSSCOMPILE
 - https://github.com/sandrohanea/whisper.net/actions/runs/14087392764/job/39454949037 => NOT CMAKE_CROSSCOMPILE

jeffbolznv (approved):
The change looks good to me.

I'm still concerned about a case where the cross-compile picks up a different glslc in the host environment vs the target environment, and if those glslcs don't have the same level of support then we may still have the same problem. I think this would be most likely to happen for android builds, where the android NDK glslc is very old. If that happens, we could fix it by also passing the glslc executable through cmake variables. I'd also be OK with you doing that now, if you want. It should be OK to use a newer glslc, since the code it generates is backward compatible.

0cc4m:
@bandoti is the most knowledgeable about cmake and cross compiling, I think.

bandoti:
There are a couple of PRs related to this in the Llama codebase:

Fix for the GGML_VULKAN_COOPMAT_GLSLC_SUPPORT:
https://github.com/ggml-org/llama.cpp/pull/12272

And I am working on adding cross-compile to CI here (also Llama):
https://github.com/ggml-org/llama.cpp/pull/12428

@sandrohanea Here is the [Github CI build](https://github.com/ggml-org/llama.cpp/pull/12428/files#diff-245fd2c5accd266a35983ed2891af1c8f8b41af027aa393075f15a00b38ff817R42) that does not require the Vulkan SDK on Linux, which should work with Ubuntu 24. This is possible by installing `glslc` and, in this case, `libvulkan-dev:riscv64`. What I am seeing at the moment is the same error message, but I am not yet convinced this is due to mismatched glslc as there is only one installed in this case—still need to test/verify that however.

ggml/src/ggml-vulkan/CMakeLists.txt
```diff
@@ -3,6 +3,14 @@ cmake_policy(SET CMP0114 NEW)
 
 find_package(Vulkan COMPONENTS glslc REQUIRED)
 
+# Allow explicitly specifying glslc executable for cross-compilation scenarios
+if(DEFINED GGML_VULKAN_GLSLC_EXECUTABLE)
```
bandoti, line 7: I am not convinced that providing a command-line override is the right solution here. When cross-compiling the standard behavior is to separate build/target utilities via the find-root path. For example, the line above is stating that it requires glslc, but it seems this should not in fact be required but found separately using a appropriate paths.

For example, what we are basically saying is:
1. We need to find Vulkan as defined in the cross-compile toolchain.
2. We need the build system glslc _no matter what_.

As such, it would be more consistent with general cross-compiling to instead use `find_program(VULKAN_GLSLC_EXECUTABLE glslc REQUIRED NO_CMAKE_FIND_ROOT_PATH)` and change the Vulkan find command to `find_package(Vulkan)` (not specifying the glslc component there).

bandoti: Thinking about the Android case though, where there are multiple NDKs, and each might have its own glslc might be worthwhile to support both modes. So, the default behavior could be to use find_program, but also allow the command-line option to ensure we have our bases covered. There have been a couple issues surrounding this so probably best to play it safe.

bandoti: Okay, so I just took a quick look in the `FindVulkan.cmake` file on my system and quickly checked the docs. Based on the docs, the search for glslc will not be repeated if it is cached first:

> This command is used to find a program. A cache entry, or a normal variable if NO_CACHE is specified, named by <VAR> is created to store the result of this command. If the program is found the result is stored in the variable and the search will not be repeated unless the variable is cleared. If nothing is found, the result will be <VAR>-NOTFOUND.

One possibility is to set the cache variable `Vulkan_GLSLC_EXECUTABLE` before calling `find_package(Vulkan COMPONENTS glslc REQUIRED)`. So, first running `find_program(Vulkan_GLSLC_EXECUTABLE glslc REQUIRED NO_CMAKE_FIND_ROOT_PATH)` would override the search. Similarly, the command-line approach overriding the Vulkan_GLSLC_EXECUTABLE cache variable would override that, and then we don't need a separate variable for the same thing. Something to consider. 🙂

jeffbolznv: > Thinking about the Android case though, where there are multiple NDKs, and each might have its own glslc might be worthwhile to support both modes. 

IMO we should just always use the glslc installed on the build system. The NDK glslc is egregiously out of date, it's really not what we want to be using.

sandrohanea: Updated the description of the PR, let me know what do you think

ggml/src/ggml-vulkan/CMakeLists.txt
```diff
@@ -22,33 +30,97 @@ if (Vulkan_FOUND)
                              ggml-vulkan.cpp
                              ../../include/ggml-vulkan.h
                             )
+    if (CMAKE_CROSSCOMPILING)
+        # Check if user has explicitly set these options
```
bandoti, line 34: I am not sure what is the need for these Android checks? Perhaps I confused matters mentioning Android, but I believe it was working correctly with the feature detection—unless I am missing something. Feature detection by attempting to compile the shaders should be able to work on all systems.

ggml/src/ggml-vulkan/vulkan-shaders/CMakeLists.txt
```diff
@@ -1,11 +1,36 @@
...
-find_program(GLSLC_EXECUTABLE glslc)
-if(NOT GLSLC_EXECUTABLE)
-    message(FATAL_ERROR "glslc not found.")
+
+# Use glslc from the parent CMake context or find it on the host system
+if(DEFINED Vulkan_GLSLC_EXECUTABLE)
+    set(GLSLC_EXECUTABLE ${Vulkan_GLSLC_EXECUTABLE})
+    message(STATUS "Using glslc from parent CMake: ${GLSLC_EXECUTABLE}")
+    # Also define this at compile time to set the default value in the shader generator
+    add_compile_definitions(VULKAN_GLSLC_EXECUTABLE="${GLSLC_EXECUTABLE}")
```
bandoti, line 8: Actually, I don't think we need this functionality at all. The vulkan-shaders-gen binary has a command-line switch which receives a path to the specified glslc. There is really no need to provide a statically compiled path to glslc, since it is already being passed in at runtime.

ggml/src/ggml-vulkan/vulkan-shaders/vulkan-shaders-gen.cpp
```diff
@@ -35,7 +36,16 @@
 std::mutex lock;
 std::vector<std::pair<std::string, std::string>> shader_fnames;
 
-std::string GLSLC = "glslc";
+// Default glslc path, can be overridden at compile time or runtime 
+#ifdef VULKAN_GLSLC_EXECUTABLE
+    // If VULKAN_GLSLC_EXECUTABLE is defined at compile time, use it
+    #define DEFAULT_GLSLC_EXECUTABLE VULKAN_GLSLC_EXECUTABLE
```
bandoti, line 42: See note above. I don't think there's any benefit to bake an executable path into the binary when it is already supplied as a command-line switch to vulkan-shaders-gen.

bandoti (commented):
After looking over this a bit more (see review notes) I think we should consider the simplicity of [llama.cpp #12272](https://github.com/ggml-org/llama.cpp/pull/12272), which directly addresses the need to fix _only_ the coopmat shader generation. I think it is in our best interest to resolve that PR first (since it came first anyhow) and circle back to this one for handling the user-specified glslc.

Note: I applied a quick patch to the [cross-build CI PR](https://github.com/ggml-org/llama.cpp/pull/12428) I have been working on to see if it can fix the issues. Had some sync issues with the CI image (unrelated to the change) so I might have to wait a bit before re-testing, but I will post a status update here once it completes.

cc: @Icenowy

sandrohanea:
I see that https://github.com/ggml-org/llama.cpp/pull/12272 was merged 🚀
That should be enough for the main issue of the CROSS_COMPILE. I was trying to allow various versions of glslc as suggested in the https://github.com/ggerganov/whisper.cpp/pull/2914#pullrequestreview-2717816538 but not sure if that is needed. Definetly not for my case in whisper.net (where only linux build is cross-compiled, not Android, for now).

Will close this PR and if different glslc versions are needed, we can create another one as a follow-up.

bandoti:
Sounds good—please feel free to loop me in if that scenario comes up! I'm happy to help.

PR #2925, opened 2025-03-23
whisper : enhance model download scripts functionality and resolve compiler warning

peter277:
* Improve whisper-cli executable path detection in model download shell scripts. If whisper-cli is found on the path, do not suggest invoking from build directory
* Enhance Windows model download batch script to have comparable functionality and behaviour as shell scripts
* Resolve compiler warning by removing duplicate definition of NOMINMAX in whisper-cli code

### DIFF
--- a/examples/cli/cli.cpp
+++ b/examples/cli/cli.cpp
@@ -13,7 +13,9 @@
 #include <cstring>
 
 #if defined(_WIN32)
+#ifndef NOMINMAX
 #define NOMINMAX
+#endif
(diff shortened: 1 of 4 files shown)


danbev (approved):
I've not been able to try the updated windows script, only the linux script (`download-coreml-model.sh currently exists early with `whisper.cpp: this script hasn't been maintained and is not functional atm`.

ggerganov (approved):

PR #2941, opened 2025-03-25
support progress_callback API for addon.node

buxuku:
## feat

- Support Inside nodejs, the compiled addon.node file provides the progress_callback API for easy tracking of processing progress
image

## fix
- In the unit test, the max_len parameter is missing, causing the unit test to fail

### DIFF
--- a/examples/addon.node/addon.cpp
+++ b/examples/addon.node/addon.cpp
@@ -130,223 +130,223 @@ void cb_log_disable(enum ggml_log_level, const char *, void *) {}
 
 class ProgressWorker : public Napi::AsyncWorker {
  public:
-  ProgressWorker(Napi::Function& callback, whisper_params params, Napi::Function progress_callback, Napi::Env env)
-      : Napi::AsyncWorker(callback), params(params), env(env) {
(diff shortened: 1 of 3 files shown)


examples/addon.node/addon.cpp
```diff
@@ -128,192 +128,227 @@ void whisper_print_segment_callback(struct whisper_context * ctx, struct whisper
...
+ private:
+  whisper_params params;
+  std::vector<std::vector<std::string>> result;
+  Napi::Env env;
+  Napi::ThreadSafeFunction tsfn;
+
+  // Custom run function with progress callback support
+  int run_with_progress(whisper_params &params, std::vector<std::vector<std::string>> &result) {
     if (params.no_prints) {
-        whisper_log_set(cb_log_disable, NULL);
```
danbev, line 133: I think I would be good to stick with 4 space indentation for C++ code here to be consistent with the rest of the code in this file and the rest of the code base. (I realize Node.js uses 2 space indentation but I think it is better not to mix)

danbev (approved):

PR #2948, opened 2025-03-26
Adding in NewContextWithStrategy for go binding

aderbedr:
Adding in NewContextWithStrategy, a function to allow you to specify the strategy when initializing the context. NewContext will use this "helper" function so that users don't need to think about what strategy they're specifying by default. I'm also doing it this way (instead of a function that sets the strategy) because the basis of what parameters we populate the context with is determined by the strategy we are using, so it needs to be done in the beginning of creating the context.


aderbedr:
I'm not sure if you want this, but I need to swap what strategy we use easily and figured it was useful, so I'd push it up and let you decide.

bindings/go/pkg/whisper/model.go
```diff
@@ -82,12 +89,17 @@ func (model *model) Languages() []string {
 }
 
 func (model *model) NewContext() (Context, error) {
+	// By default, specify the greedy strategy
+	return model.NewContextWithStrategy(SAMPLING_GREEDY)
+}
+
+func (model *model) NewContextWithStrategy(strategy SamplingStrategy) (Context, error) {
```
danbev, line 96: Perhaps instead of passing in just the strategy to this function, it could accepts a parameter of type `whisper.Params`?

bindings/go/pkg/whisper/model.go
```diff
@@ -20,6 +20,13 @@ type model struct {
 // Make sure model adheres to the interface
 var _ Model = (*model)(nil)
 
+type SamplingStrategy whisper.SamplingStrategy
```
danbev, line 23: I think having this and the [globals](https://github.com/ggerganov/whisper.cpp/blob/f28bf5d1864fe039060a9fa84ffe72433d20153e/bindings/go/whisper.go#L77-L80) might be a little confusing.

danbev (commented):
I'm not sure about these changes but perhaps we need to make the API more flexible. One thing I noticed is that [Whisper_init](https://github.com/ggerganov/whisper.cpp/blob/f28bf5d1864fe039060a9fa84ffe72433d20153e/bindings/go/whisper.go#L102) does not take a `whisper.Params` which the [java binding](https://github.com/ggerganov/whisper.cpp/blob/f28bf5d1864fe039060a9fa84ffe72433d20153e/bindings/java/src/main/java/io/github/ggerganov/whispercpp/WhisperCpp.java#L46) does. Perhaps that would be an option here as well.

bindings/go/pkg/whisper/model.go
aderbedr: Yeah, the other option was to have clients import both versions, but figured that'd be more of a headache. Alright, no worries!

danbev:
Closing this for now as I'm not sure about adding this at the moment, but feel free to re-open.

Issue #2965, opened 2025-03-30
[Vulkan] StackOverflow when running ggml-vulkan on NVidia GPU on Windows


sandrohanea:
It seems that latest `ggml-vulkan` is causing Nvidia's `nvoglv64.dll` to stackoverflow.

This issue was initially reported here: https://github.com/sandrohanea/whisper.net/issues/372

GPU used: both NVIDIA RTX 4090 laptop and NVIDIA GeForce RTX 4060 Ti

Details about ggml compilation:
 - It used cooperative_matrix shaders but not cooperative_matrix2
 - You can see the build logs here: https://github.com/sandrohanea/whisper.net/actions/runs/14146821774/job/39635137868

It can be reproduced 100% of the times (not transient error).

Compiled it in debug and the Stack overflow originates in:

`CREATE_MM2(GGML_TYPE_F16, pipeline_matmul_f16, matmul_f16, wg_denoms, warptile, vk_mat_mat_push_constants, 3, );` for `VK_KHR_cooperative_matrix`

For that, `ggml_vk_create_pipeline_func` will crash:

![Image](https://github.com/user-attachments/assets/c3d5954c-fe4f-4c72-8632-1f2d9f16f3f4)

After that, the rest of the stack, indicates some infinite recursive call (inside `nvoglv64`):

![Image](https://github.com/user-attachments/assets/c5282c3d-d2fa-41f2-9780-4d5b44b9094c)

Full call stack attached:
[vulkan-issue-callstack.txt](https://github.com/user-attachments/files/19526548/vulkan-issue-callstack.txt)

Any idea what could be the issue @0cc4m @jeffbolznv @bandoti?

Thanks!

sandrohanea:
Actually, it is `CREATE_MM2(GGML_TYPE_F16, pipeline_matmul_f16, matmul_f16, wg_denoms, warptile, vk_mat_mat_push_constants, 3, );` on line :2010 :

![Image](https://github.com/user-attachments/assets/edbe4ff8-3411-404e-a171-c90e3bc30495)

This means that it will reproduce no matter if the `GGML_VULKAN_COOPMAT2_GLSLC_SUPPORT` or `GGML_VULKAN_COOPMAT_GLSLC_SUPPORT` is defined (it is for the shaders that are compiled and loaded all the time).

jeffbolznv:
The log you linked includes this:

```
-- Vulkan found
-- GL_NV_cooperative_matrix2 not supported by glslc
[...]

PR #2966, opened 2025-03-30
cmake: improve Vulkan cooperative matrix support checks

sandrohanea:
This PR addresses a small issue if `GGML_VULKAN_COOPMAT_GLSLC_SUPPORT` or `GGML_VULKAN_COOPMAT2_GLSLC_SUPPORT` is defined as option, after https://github.com/ggml-org/llama.cpp/pull/12272

The current behaviour will forward it to the `vulkan-shaders-gen` cmake process but it won't add the compile definition for the main cmake process.

If some client will send it as parameter with `-DGGML_VULKAN_COOPMAT_GLSLC_SUPPORT=ON` will result in runtime issues for *some hardware*, see https://github.com/ggerganov/whisper.cpp/issues/2965 for more details.

CC: @jeffbolznv , @bandoti


ggml/src/ggml-vulkan/CMakeLists.txt
```diff
@@ -36,9 +36,14 @@ if (Vulkan_FOUND)
             set(GGML_VULKAN_COOPMAT_GLSLC_SUPPORT OFF CACHE INTERNAL "Whether coopmat is supported by glslc")
         else()
             message(STATUS "GL_KHR_cooperative_matrix supported by glslc")
-            add_compile_definitions(GGML_VULKAN_COOPMAT_GLSLC_SUPPORT)
             set(GGML_VULKAN_COOPMAT_GLSLC_SUPPORT ON CACHE INTERNAL "Whether coopmat is supported by glslc")
         endif()
+    else()
+        message(STATUS "GL_KHR_cooperative_matrix support already defined: ${GGML_VULKAN_COOPMAT_GLSLC_SUPPORT}")
```
bandoti, line 42: Are these status messages required? I am okay with keeping it if you think this is useful. 🙂

bandoti (approved):
Looks good to me.

ggml/src/ggml-vulkan/CMakeLists.txt
sandrohanea: I think they are good indicators. Not strictly required, but they would have made identifying the issue in https://github.com/ggerganov/whisper.cpp/issues/2965 a lot easier. It was strange that no GGML_VULKAN_COOPMAT_GLSLC_SUPPORT indicator existed if it was defined externally.

Let me know if you want to remove it (if you think it is too verbose)

bandoti: Okay sounds good.

ggerganov:
This is ready to merge? I will then sync it upstream.

bandoti (approved):

bandoti:
Yep looks good to merge.

PR #2873, opened 2025-03-12
whisper: add xcframework build script

mdestagnol:
This is adding xcframework support for Whisper; I followed the same strategy as the work done by @danbev on Llama https://github.com/ggml-org/llama.cpp/pull/11996

~~I didn't yet ported the apple validation scripts; I'll have to check that part tomorrow but shouldn't be hard to add.~~

UPDATE: the Apple Validation scripts are added. We should be good to go.

PR #2994, opened 2025-04-02
release : v1.7.5

ggerganov:
- Update the version numbers to the new version
- Do not squash-merge this commit to avoid appending the (1234) suffix to the commit title

After that, we'll create a release from the Github UI, selecting this commit and creating a tag  `v1.7.5`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.5) # for add_link_options and implicit target directories.
 project("whisper.cpp" C CXX)
-project("whisper.cpp" VERSION 1.7.4)
+project("whisper.cpp" VERSION 1.7.5)
 include(CheckIncludeFileCXX)
 
 set(SOVERSION 1)
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 > [!NOTE]
 > New maintenance roadmap: https://github.com/ggerganov/whisper.cpp/discussions/2788
 
-Stable: [v1.7.4](https://github.com/ggerganov/whisper.cpp/releases/tag/v1.7.4) / [Roadmap | F.A.Q.](https://github.com/ggerganov/whisper.cpp/discussions/126)
+Stable: [v1.7.5](https://github.com/ggerganov/whisper.cpp/releases/tag/v1.7.5) / [Roadmap | F.A.Q.](https://github.com/ggerganov/whisper.cpp/discussions/126)
 
 High-performance inference of [OpenAI's Whisper](https://github.com/openai/whisper) automatic speech recognition (ASR) model:
 
--- a/bindings/javascript/package.json
+++ b/bindings/javascript/package.json
@@ -1,6 +1,6 @@
 {
   "name": "whisper.cpp",
-  "version": "1.7.4",
+  "version": "1.7.5",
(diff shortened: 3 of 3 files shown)


danbev (approved):

itsthisjustin:
Are there instructions for building the swift package? i added the repo but noticed the framework isn't listed when i go to add it to the target

danbev:
@itsthisjustin The Package.swift file was [removed](https://github.com/ggerganov/whisper.cpp/pull/2873) in favor of using the xcframework. For a swift project the xcframework can be used like this:
```swift
// swift-tools-version: 5.10
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "Whisper",
    targets: [
        .executableTarget(
            name: "Whisper",
            dependencies: [
                "WhisperFramework"
            ]),
        .binaryTarget(
            name: "WhisperFramework",
            url: "https://github.com/ggerganov/whisper.cpp/releases/download/v1.7.5/whisper-v1.7.5-xcframework.zip",
            checksum: "c7faeb328620d6012e130f3d705c51a6ea6c995605f2df50f6e1ad68c59c6c4a"
        )
    ]
)
```
I'll make sure we document this somewhere as I don't think we have. Perhaps this should also be included in the release notes moving forward, providing the correct checksum as well.

itsthisjustin:
> @itsthisjustin The Package.swift file was [removed](https://github.com/ggerganov/whisper.cpp/pull/2873) in favor of using the xcframework. For a swift project the xcframework can be used like this:
> 
> ```swift
> // swift-tools-version: 5.10
> // The swift-tools-version declares the minimum version of Swift required to build this package.
> 
> import PackageDescription
> 
> let package = Package(
>     name: "Whisper",
>     targets: [
>         .executableTarget(
>             name: "Whisper",
>             dependencies: [
>                 "WhisperFramework"
>             ]),
>         .binaryTarget(
>             name: "WhisperFramework",
>             url: "https://github.com/ggerganov/whisper.cpp/releases/download/v1.7.5/whisper-v1.7.5-xcframework.zip",
>             checksum: "c7faeb328620d6012e130f3d705c51a6ea6c995605f2df50f6e1ad68c59c6c4a"
>         )
>     ]
> )
> ```
> 
> I'll make sure we document this somewhere as I don't think we have. Perhaps this should also be included in the release notes moving forward, providing the correct checksum as well.

This is super helpful. I'd say a little guide for us Swift folks could be super useful. This isn't a "standard" way to use a package like this so it's slightly outside of my existing knowledge compared to how it used to work.

itsthisjustin:
Also not sure if unique to me but i couldn't build the script with xros as the visionOS command. I had to change it visionos and then it build the binaries.

n374:
> Also not sure if unique to me but i couldn't build the script with xros as the visionOS command. I had to change it visionos and then it build the binaries.

Same here, got a `unknown platform: xros` error when running `./build-xcframework.sh` directly, works after replace `xros` with `visionos` in line 347

- Xcode: Version 16.3 (16E140)
- macOS: 15.3.2 (24D81)
- Chip: Apple M4
- Machine: Mac mini 2024

danbev:
@itsthisjustin @n374 Thanks for reporting this. I'll take a closer look.

PR #2999, opened 2025-04-03
tests : add script to benchmark whisper.cpp on LibriSpeech corpus

fujimotos:
LibriSpeech is a widely-used benchmark dataset for training and
testing speech recognition models.

* For more details about LibriSpeech, see https://www.openslr.org/12
 
This adds a set of scripts to measure the recognition accuracy of
whisper.cpp models, following the common benchmark standards.


tests/librispeech/normalizers/LICENSE
```diff
@@ -0,0 +1,25 @@
+Code in this directory is adapted from OpenAI Whisper project
+(https://github.com/openai/whisper) and carries the following
+copyright and license.
```
fujimotos, line 3: As I mentioned above, the normalizer implementation in the `tests/normalizer/`
subfolder was ported from the upstream project.

- We need this to get a comparable WER score. See [this notebook](https://github.com/openai/whisper/blob/main/notebooks/LibriSpeech.ipynb)
  about how OpenAI evaluate their speech recognition models.

- The reason why I commited these files to this reposjitory is to minimize the
  dependencies we need to run the benchmark script. 

In particular, `pip install openai-whisper` package requires a full PyTorch
libraries, so it's heavy.

tests/librispeech/README.md
```diff
@@ -0,0 +1,45 @@
...
+### How to change the inferece parameters
+
+Create `eval.conf` and override variables.
+
+```
+WHISPER_MODEL = large-v3-turbo
+WHISPER_FLAGS = --no-prints --threads 8 --language en --output-txt
+```
+
+Check out `eval.mk` for more details.
```
fujimotos, line 45: This `README` file describes how to perform the benchmark tests.

Confirmed to work on Ubuntu 24.04 and Amazon Linux 2023.

tests/librispeech/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+[LibriSpeech](https://www.openslr.org/12) is a standard dataset for
+training and evaluating automatic speech recognition systems.
+
+This directory contains a set of tools to evaluate the recognition
+performance of whisper.cpp on LibriSpeech corpus.
+
+## Quick Start
+
+1. (Pre-requirement) Compile `whisper-cli` and prepare the Whisper
```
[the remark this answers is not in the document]
fujimotos: @danbev  Based on your feedback, I added a short command
example in commit de329b78.

tests/librispeech/eval.mk
```diff
@@ -0,0 +1,39 @@
+PYTHON = ./venv/bin/python3
```
ggerganov, line 1: It would be better to call a generic `python` command:

```suggestion
PYTHON = python
```

For example, I use `pyenv` to manage my python environments and this is not compatible with that workflow, while simply using `python` I believe it should work in both cases.

tests/librispeech/Makefile
```diff
@@ -0,0 +1,22 @@
...
+
+eval:
+	$(MAKE) -f eval.mk
+
+clean:
+	$(MAKE) -f eval.mk clean
+
+setup-venv:
+	python3 -m venv venv
+	./venv/bin/pip install -r requirements.txt
```
ggerganov, line 13: I'm not very experienced with Python so I might be wrong, but isn't it required to source the virtual environment like this:

```suggestion
setup-venv:
	python3 -m venv venv
	source ./venv/bin/activate
	pip install -r requirements.txt
```

fujimotos: > but isn't it required to source the virtual environment like this:

Since Makefile executes each line in a separate shell, the third 
line (`pip install`) would be executed outside of the virtualenv.

*Result on Ubuntu 24.04 (with `python3-venv`)*

```console
$ make setup-venv
python3 -m venv venv
source ./venv/bin/activate
pip install -r requirements.txt
bash: line 1: pip: command not found
make: *** [Makefile:16: setup-venv] Error 127
```

Calling `./venv/bin/pip` is sort-of a common workaround to
handle this Makefile issue (this is why I called `./venv/bin/python`
in another part of the code).

> For example, I use pyenv to manage my python environments and this is not compatible with that workflow

I'm going to set up a `pyenv` environment myself, and make sure
that the Python part works both on `pyenv` and `python+virtualenv`.
Wait for me!

ggerganov: I see.

Btw, there is also the solution to leave the python environment configuration out of the scripts and just assume that `python` is available and the user has done the necessary steps to install the dependencies.

ggerganov (approved):
This is really great!

There are 2 things that we can improve on:

- The dataset seems to contain only relatively short speech segments. I think it would be good to have a dataset with a bit longer samples (i.e. a few minutes) in order to exercise the rolling window transcription that Whisper does

- The current implementation loads and unloads the entire model for each sample. This is very inefficient. Instead, it should utilize the `whisper-server` to start it once and send all the samples via HTTP request. This will make the benchmark much faster.

For now we can merge and improve on these later.

tests/librispeech/README.md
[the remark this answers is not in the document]
danbev: That looks great, thanks!

danbev (approved):

fujimotos:
@ggerganov @danbev Thank you! I'm glad that it helps this project.

PR #3016, opened 2025-04-08
Update uri.rb

Olli:
Bugfix ... without this Pathname the "/" operator wouldn't work and will throw an error


KitaitiMakoto (approved):
Good catch!

PR #3027, opened 2025-04-10
Added whisper_get_system_info_json

peardox:
Added a modified version of whisper_print_system_info that returns JSON. This is desirable if porting to a different language
There may be a bug-in-waiting as the values are unquoted (which was always the case in testing). If strings are likely values then a modification will be required to suit

### DIFF
--- a/include/whisper.h
+++ b/include/whisper.h
@@ -437,6 +437,7 @@ extern "C" {
 
     // Print system information
     WHISPER_API const char * whisper_print_system_info(void);
+	WHISPER_API const char * whisper_get_system_info_json(void);
 
     ////////////////////////////////////////////////////////////////////////////
 
--- a/include/whisper.h
+++ b/include/whisper.h
@@ -437,7 +437,6 @@ extern "C" {
 
     // Print system information
     WHISPER_API const char * whisper_print_system_info(void);
-	WHISPER_API const char * whisper_get_system_info_json(void);
 
     ////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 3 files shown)


include/whisper.h
```diff
@@ -437,6 +437,7 @@ extern "C" {
 
     // Print system information
     WHISPER_API const char * whisper_print_system_info(void);
+	WHISPER_API const char * whisper_get_system_info_json(void);
```
danbev, line 440: Nit: can you use spaces instead of tabs here.

src/whisper.cpp
```diff
@@ -4346,6 +4346,48 @@ const char * whisper_print_system_info(void) {
...
+
+    for (size_t i = 0; i < ggml_backend_reg_count(); i++) {
+        auto * reg = ggml_backend_reg_get(i);
+        auto * get_features_fn = (ggml_backend_get_features_t) ggml_backend_reg_get_proc_address(reg, "ggml_backend_get_features");
+        if (get_features_fn) {
+            ggml_backend_feature * features = get_features_fn(reg);
+            s += ",\"";
+            s += ggml_backend_reg_name(reg);
+            s += "\":{";
+			auto first = true;
```
danbev, line 4371: Nit: same here with tab and a few of the lines below as well.

danbev (approved):

src/whisper.cpp
peardox: Fixed the tabs. Also added two helpful state-related functions.
Moved everything to end of source files to make it easy to play with.
Also changed JSON to always return string values leaving it up to the caller to handle typing. This ensures that the returned JSON is always valid.

peardox:
Oh, tabs were Notepad++ (I write code in it). I've changed the default to always use spaces as a result (as I have more PRs upcoming)

src/whisper.cpp
```diff
@@ -7590,3 +7548,80 @@ static void whisper_log_callback_default(ggml_log_level level, const char * text
...
+    return s.c_str();
+}
+
+// whisper_get_state_from_context
+// Returns state from supplied context pointer
+// This is mainly a helper for non-C++ language bindings as whisper_context
+// has embedded C++ specific types (e.g. maps and vectors)
+struct whisper_state * whisper_get_state_from_context(struct whisper_context * ctx) {
+    if (!ctx->state) {
+        whisper_free(ctx);
```
danbev, line 7607: Should this perhaps just return a `nullptr` and not free the `whisper_context`? Just thinking that I would not expect anything else to happen in a function named `whisper_get_state_from_context` than either the state being returned or not.

src/whisper.cpp
danbev: Thanks!

src/whisper.cpp
peardox: !000% that's what it should do - apologies for very stupid error (blaming cut + paste used to create it...)

peardox:
Intend to move this to #3068 to keep main code clean as its only real use is for other languages

PR #3028, opened 2025-04-10
whisper : reduce delta_min from 1000ms to 100ms

ggerganov:
fix https://github.com/ggml-org/whisper.cpp/issues/2065

Should re-run https://github.com/ggml-org/whisper.cpp/pull/2999 WER benchmark just in case to make sure something did not break from this.


danbev (approved):
I'm running the `librispeech` test now, and it looks good so far. Completed this on macos :heavy_check_mark:

Issue #2963, opened 2025-03-29
GGML_CPU_ALL_VARIANTS causes a silent exception


peardox:
Building on Win11 using VC like this (CMAKE_TOOLCHAIN_FILE is for SDL2)

> cmake -B build -DCMAKE_TOOLCHAIN_FILE=C:\git\vcpkg\scripts\buildsystems\vcpkg.cmake -DGGML_CCACHE=OFF -DGGML_CUDA=1 -DGGML_VULKAN=1 -DWHISPER_SDL2=ON -DGGML_BACKEND_DL=ON -DBUILD_SHARED_LIBS=ON -DGGML_CPU_ALL_VARIANTS=ON

This should (and does) build CPU/CUDA/Vulkan for all x64 CPUs back to Haswell, the aim, of course, being to allow for older CPU architectures so the same apps will run on a wide range of end-user configs

However, when it comes to loading GGML it fails silently. Looking into this but my C++ is rusty

peardox:
PR #3042 

Finally traced down the issue.

If e.g. whisper_init_from_file_with_params is called in a GGML_BACKEND_DL build the examples have no backend yet but something within the init is calling the backend

I presume there's a list of backends which is empty when called via GGML_BACKEND_DL as a null pointer is eventually returned where it shouldn't be. 

Not sure why it works without GGML_BACKEND_DL but getting it working was the big goal for me.

I've modified bench, cli and stream to call whisper_load_backends()  which I have added to whisper.h

If whisper_load_backends() is called without GGML_BACKEND_DL it does nothing (existing code) so the extra header function is benign

peardox:
PR failed so made the whisper.h change conditional as well so as not to confuse things

peardox:
As suggested tried ggml_backend_load_all then added ggml_backend_load_best.
The benefit of adding a load_best is that it may be preferable to use e.g. vulkan over CUDA in very specific circumstances - namely one of the new AMD APUs with a Radeon 8060S Graphics subsystem as it MAY be faster than CUDA if in a machine with both (8060S = 4060 so they say)
A side-effect of load_best is that a command-line switch could specify desired backend - cpu/vulkan/cuda etc

PR #3042, opened 2025-04-14
Expose whisper_load_backends to make GGML_BACKEND_DL functional

peardox:
As per #2963 
Added whisper_load_backends to whisper.h
Modified 3 examples (bench, cli and stream) to \#ifdef GGML_BACKEND_DL call the function

### DIFF
--- a/examples/bench/bench.cpp
+++ b/examples/bench/bench.cpp
@@ -60,6 +60,9 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
 
 static int whisper_bench_full(const whisper_params & params) {
     // whisper init
+    #ifdef GGML_BACKEND_DL
+    whisper_load_backends();
+    #endif
 
     struct whisper_context_params cparams = whisper_context_default_params();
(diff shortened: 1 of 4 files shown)


slaren:
In retrospect it was a mistake to include the call to load the backends (`ggml_backend_load_all`) in whisper.cpp. This is something that should be the responsibility of applications, so that they also have the freedom to choose what backends to load, and from which locations.

peardox:
ggml_backend_load_all may be slight overkill but the multiple architectures for CPU is a benefit if doing a release is a definite boon especially in the case of x64

slaren:
My suggestion would be:
- Remove `whisper_load_backends`
- Add a call to `ggml_backend_load_all` during initialization to all the examples

peardox:
whisper_load_backends can only be run once while ggml_backend_load_all has no such prevention

peardox:
OK, made this PR significantly better with ggml_backend_load_all and ggml_backend_load_best.
Will redo ASAP

PR #3050, opened 2025-04-15
cli : Support "-" for stdout like stdin

### LINKED ISSUE #3048: Can't output to file descriptor
I would like `whisper-cli` to have the generality of being able to output to a file descriptor. We can already input from a fd using `/dev/fd/3`, but it doesn't work with `-of /dev/fd/4`. Accepting an output file descriptor could allow me the flexibility of using sockets and pipes.

Here is an simplified example of the current error:
```console
$ exec 3<path/to/wav
$ exec 4>/path/to/vtt
$ ~/path/to/whisper-cli -l auto -ovtt -m path/to/model /dev/fd/3 -of /dev/fd/4
[...]
whisper_model_load: [...]
[...]

main: processing '/dev/fd/3' [...]

[00:00:00.000 --> 00:00:10.000]   [...]
[...]

output_vtt: failed to open '/dev/fd/4.vtt' for writing

whisper_print_timings: [...]
```

danielzgtg:
This changes `examples/cli/cli.cpp` to be like `examples/common-whisper.cpp`. `-of -` can be specified (or this can be inferred from `-` as the input file) to output to stdout. This is useful for piping to other applications.

Also fixed `output_score` and `output_wts` not outputting some logging messages that `output_json` does.

Closes #3048

### DIFF
--- a/examples/cli/cli.cpp
+++ b/examples/cli/cli.cpp
@@ -9,6 +9,7 @@
 #include <cstdio>
 #include <string>
 #include <thread>
+#include <utility>
 #include <vector>
 #include <cstring>
 
@@ -379,15 +380,7 @@ static void whisper_print_segment_callback(struct whisper_context * ctx, struct 
     }
 }
 
-static bool output_txt(struct whisper_context * ctx, const char * fname, const whisper_params & params, std::vector<std::vector<float>> pcmf32s) {
-    std::ofstream fout(fname);
-    if (!fout.is_open()) {
-        fprintf(stderr, "%s: failed to open '%s' for writing\n", __func__, fname);
-        return false;
-    }
-
-    fprintf(stderr, "%s: saving output to '%s'\n", __func__, fname);
-
+static void output_txt(struct whisper_context * ctx, std::ofstream & fout, const whisper_params & params, std::vector<std::vector<float>> pcmf32s) {
     const int n_segments = whisper_full_n_segments(ctx);
     for (int i = 0; i < n_segments; ++i) {
         const char * text = whisper_full_get_segment_text(ctx, i);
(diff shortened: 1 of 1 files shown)


danielzgtg:
It's strange that the unrelated `examples/talk-llama/llama-mmap.h` had an error in https://github.com/ggml-org/whisper.cpp/actions/runs/14477612749/job/41345196960?pr=3050 when I only touched `examples/cli/cli.cpp`.

danbev:
> It's strange that the unrelated `examples/talk-llama/llama-mmap.h` had an error

This is unrelated to this PR, it was fixed by https://github.com/ggml-org/whisper.cpp/pull/3084.

examples/cli/cli.cpp
```diff
@@ -1066,8 +1067,45 @@ int main(int argc, char ** argv) {
     }
 
     for (int f = 0; f < (int) params.fname_inp.size(); ++f) {
-        const auto fname_inp = params.fname_inp[f];
-		const auto fname_out = f < (int) params.fname_out.size() && !params.fname_out[f].empty() ? params.fname_out[f] : params.fname_inp[f];
+        const auto & fname_inp = params.fname_inp[f];
+        class FnameOutFactory {
```
danbev, line 1071: Could you rename this class to follow the [naming conventions](https://github.com/ggml-org/llama.cpp/blob/master/CONTRIBUTING.md#naming-guidelines). In this case something like `fname_out_factory` perhaps.

examples/cli/cli.cpp
```diff
@@ -1066,8 +1067,45 @@ int main(int argc, char ** argv) {
...
+            bool used_stdout;
+
+        public:
+            const bool is_stdout;
+            decltype(whisper_print_segment_callback) * const print_segment_callback;
+
+            FnameOutFactory(const std::string & fname_out, const std::string & fname_inp, whisper_params & params) :
+                    buf{!fname_out.empty() ? fname_out : fname_inp},
+                    basename_length{buf.size()},
+                    is_stdout{buf == "-"}, used_stdout{params.diarize},
```
danbev, line 1083: Perhaps place `used_stdout` on a separate line for consistency with the other fields.

danbev (approved):
I tried this using json output format and stdout using the following command:
```console
../build/bin/whisper-cli -m models/ggml-tiny.en.bin -f samples/jfk.wav -oj -of - -np 2> /dev/null | jq -c '.transcription[].text'
" And so my fellow Americans ask not what your country can do for you, ask what you can do for your country."
```
Without redirecting `stderr` the following will be printed as output to the console:
```console
output_json: saving output to '/dev/stdout'
" And so my fellow Americans ask not what your country can do for you, ask what you can do for your country."
``` 
It might be nice to not have to specify the redirect off `stderr` when using `stdout`. This was not something that was introduced by this PR but I think it would nice to perhaps have a check and not output this log message in this case.

examples/cli/cli.cpp
danielzgtg: Renamed to follow the requested naming conventions. It was to `fname_out_factory` before changes to "not output this log message" further renamed it to `fout_factory`. Note that now the struct and variable are homonyms.

`build/bin/whisper-cli -m models/ggml-tiny.en.bin -f samples/jfk.wav -oj -of - -np` no longer outputs `output_json: saving output to '/dev/stdout'`.

danielzgtg:
Squashed and rebased on top of latest master to see if that fixes CI

PR #3068, opened 2025-04-23
Add base framework for flat bindings 

peardox:
This PR adds a new option to cmake - the BINDINGS_FLAT option

If undefined everything is the same as if it didn't exist making it completely benign

If defined it includes ggml-flat.cpp and whisper-flat.cpp to the build of the dynamic libraries whisper and ggml. These are intended to allow extra functionality to be exposed in the shared libs. This is intended to allow other languages to add stuff they need to the dynamic libraries.

There's only one function added to the libraries ATM - whisper_flat_backends_load_all() which just calls ggml_backends_load_all()
There are no functions in the GGML version at all (not worked out what's needed yet), this is, however, likely to have some entries eventually.


peardox:
My recent addtions were illuminating.

If you load vulkan and cuda then your chosen device ends up being whichever was loaded first (using ggml_backend_load_best)

At least that's what happens when my 4070 is both cuda and vulkan device

Oddly vulkan was faster - that may be down to NVIDIA's current diver problems though

peardox:
Arggh - Mac needs a few fixups

peardox:
Mac now happy

PR #3097, opened 2025-04-30
removed stdc++fs

JaredTweed:
### **Why drop `-lstdc++fs` from *whisper.cpp*’s CMakeLists?**

1. **The extra library vanished from modern GCC** – beginning with GCC 13 the standalone `libstdc++fs` archive was removed; the symbols are now built into the normal C++ runtime, so the linker flag causes hard failures on any tool-chain ≥ 13 ( `/usr/bin/ld: cannot find -lstdc++fs` ).  

2. **C++17 made it obsolete** – since GCC 9, `std::filesystem` is a fully-standard header in `<filesystem>` and no longer lives in `<experimental/filesystem>`, so no explicit link option is required.  

3. **Keeping the flag blocks Flatpak and cross-arch builds** – many Flatpak runners (e.g. GNOME 48) already ship GCC 14; the flag currently breaks both x86-64 and aarch64 sandbox builds of *whisper.cpp*.  

4. **Upstream consensus** – other OSS projects (LLVM, CMake, Dolphin Emu, etc.) have removed the flag for exactly the same reason. 

5. **No functional downside** – when using older GCC (< 9) the flag is *still* unnecessary because those versions fall back to the static archive automatically when `std::filesystem` symbols are unresolved.

### DIFF
--- a/ggml/src/CMakeLists.txt
+++ b/ggml/src/CMakeLists.txt
@@ -214,7 +214,6 @@ add_library(ggml
 target_link_libraries(ggml PUBLIC ggml-base)
 
 if (CMAKE_SYSTEM_NAME MATCHES "Linux")
-    target_link_libraries(ggml PRIVATE dl stdc++fs)
 endif()
 
 function(ggml_add_backend_library backend)
--- a/ggml/src/CMakeLists.txt
+++ b/ggml/src/CMakeLists.txt
@@ -214,6 +214,7 @@ add_library(ggml
 target_link_libraries(ggml PUBLIC ggml-base)
 
 if (CMAKE_SYSTEM_NAME MATCHES "Linux")
+    target_link_libraries(ggml PRIVATE dl)
 endif()
 
 function(ggml_add_backend_library backend)


ggerganov:
The flag was added in https://github.com/ggml-org/llama.cpp/pull/12094

cc @ericcurtin We should recommend to upgrade the compiler when the issue appears again.

ericcurtin:
@hbuxiaofei does this work for you?

Maybe you can put in a little conditional to add the flag for your centos8 case

I wouldn't block on @hbuxiaofei responding, we should favour the newest distros, centos8 is a bit long in the tooth for llama.cpp

ericcurtin:
Maybe the centos8 solution is upgrade the compiler like @ggerganov says, I think you can do that by installing one of the Developer Toolsets

ericcurtin (approved):

ericcurtin:
Ignore my approval, need the dl

danielzgtg:
> hard failures on any tool-chain ≥ 13 ( /usr/bin/ld: cannot find -lstdc++fs ).
> Keeping the flag blocks Flatpak

Can you provide logs and the exact steps to get the error? It links fine on `g++ (Ubuntu 14.2.0-4ubuntu2) 14.2.0` / `-- The CXX compiler identification is GNU 14.2.0`, which is "≥ 13".

The linker command used is:

```
$ cd build; rg 'stdc\+\+fs'
ggml/src/CMakeFiles/ggml.dir/link.txt
1:/usr/bin/c++ -fPIC -O3 -DNDEBUG -shared -Wl,-soname,libggml.so -o libggml.so "CMakeFiles/ggml.dir/ggml-backend-reg.cpp.o"  -Wl,-rpath,/home/home/CLionProjects/whisper.cpp/build/ggml/src: -ldl -lstdc++fs libggml-cpu.so libggml-base.so
```

Aside from the lack of logs, I'm not opposed to this change as the build still succeeds with it.

JaredTweed:
> > hard failures on any tool-chain ≥ 13 ( /usr/bin/ld: cannot find -lstdc++fs ).
> > Keeping the flag blocks Flatpak
> 
> Can you provide logs and the exact steps to get the error? It links fine on `g++ (Ubuntu 14.2.0-4ubuntu2) 14.2.0` / `-- The CXX compiler identification is GNU 14.2.0`, which is "≥ 13".
> 
> The linker command used is:
> 
> ```
> $ cd build; rg 'stdc\+\+fs'
> ggml/src/CMakeFiles/ggml.dir/link.txt
> 1:/usr/bin/c++ -fPIC -O3 -DNDEBUG -shared -Wl,-soname,libggml.so -o libggml.so "CMakeFiles/ggml.dir/ggml-backend-reg.cpp.o"  -Wl,-rpath,/home/home/CLionProjects/whisper.cpp/build/ggml/src: -ldl -lstdc++fs libggml-cpu.so libggml-base.so
> ```
> 
> Aside from the lack of logs, I'm not opposed to this change as the build still succeeds with it.

danbev (approved):

PR #3102, opened 2025-04-30
docs : Update cli documentation

antimatter96:
This updates the documentation of cli based on the actual output

In the longterm this should ideally be auto generated to prevent mismatch

### DIFF
--- a/examples/cli/README.md
+++ b/examples/cli/README.md
@@ -6,7 +6,8 @@ It can be used as a reference for using the `whisper.cpp` library in other proje
 ```
 ./build/bin/whisper-cli -h
 
-usage: ./build-pkg/bin/whisper-cli [options] file0.wav file1.wav ...
+usage: ./build-pkg/bin/whisper-cli [options] file0 file1 ...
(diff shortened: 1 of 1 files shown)


examples/cli/README.md
```diff
@@ -6,7 +6,8 @@ It can be used as a reference for using the `whisper.cpp` library in other proje
 ```
 ./build/bin/whisper-cli -h
 
-usage: ./build-pkg/bin/whisper-cli [options] file0.wav file1.wav ...
+usage: ./build-pkg/bin/whisper-cli [options] file0 file1 ...
```
danbev, line 9: I think this might be incorrect in the current README.md and that the correct path should  be `./build/bin/whisper-cli` and not `./build-pkg`. For example, this is the output for me locally:
```console
$ ./build/bin/whisper-cli --help

usage: ./build/bin/whisper-cli [options] file0 file1 ...
supported audio formats: flac, mp3, ogg, wav
...
```

antimatter96: Nice catch. Fixed

danbev (approved):

PR #3104, opened 2025-04-30
Flat bindings for non C++ use

peardox:
As it's only of use to people using other languages the code is only used if -DBINDINGS_FLAT is passed to cmake

If BINDINGS_FLAT is not present then the end ressult is same as master

It was necessary to disable the recent MSVC warnings commit as this broke GGML_CPU_ALL_VARIANTS


danbev:
> It was necessary to disable the recent MSVC warnings commit as this broke GGML_CPU_ALL_VARIANTS

Sorry about that, it should be addressed by https://github.com/ggml-org/whisper.cpp/pull/3103.

peardox:
re-done as #3114

PR #3114, opened 2025-05-02
Flat fix

peardox:
Fixingcommit conflicts


peardox:
@danbev your #3090 and others are now in master so will this get thru CI now?

I just merged and everything looks good

KitaitiMakoto:
@peardox  Hi, I sent a pull request to fix Ruby problem: https://github.com/WhisperBooks/whisper.cpp/pull/1

src/whisper-flat.cpp
```diff
@@ -0,0 +1,60 @@
+#include "whisper.h"
```
danbev, line 1: Could the number of includes be reduces here to only `whisper-flat.h` and `ggml-backend.h` perhaps?

src/whisper-flat.cpp
```diff
@@ -0,0 +1,60 @@
...
+#ifdef BINDINGS_FLAT
+void whisper_flat_backend_load_all(void) {
+    ggml_backend_load_all();
+}
+
+const char * whisper_flat_get_system_info_json(void) {
+    return whisper_get_system_info_json();
+}
+
+struct whisper_state * whisper_flat_get_state_from_context(struct whisper_context * ctx) {
```
danbev, line 41: Should this really be exposed? My reason for asking is that `whisper_state` contains C++ constructs like `std::vector` and my understanding is that the purpose of this is avoid them as not all languages can deal with them in an easy way.

src/whisper-flat.h
```diff
@@ -0,0 +1,36 @@
...
+#include "whisper.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+    #ifdef BINDINGS_FLAT
+    WHISPER_FLAT_API void whisper_flat_backend_load_all(void);
+    WHISPER_FLAT_API struct whisper_activity * whisper_flat_get_activity_with_state(struct whisper_state * state);
+    WHISPER_FLAT_API struct whisper_state * whisper_flat_get_state_from_context(struct whisper_context * ctx);
```
danbev, line 26: Same question as above, this should really be exposed?

src/whisper.cpp
```diff
@@ -1309,14 +1309,19 @@ static size_t aheads_masks_nbytes(struct whisper_aheads_masks & aheads_masks) {
 static ggml_backend_t whisper_backend_init_gpu(const whisper_context_params & params) {
     ggml_log_set(g_state.log_callback, g_state.log_callback_user_data);
 
+    #ifndef BINDINGS_FLAT
```
danbev, line 1312: Nit: Macros are mostly placed at the beginning of the line in the code base, even if the code they guard is indented. This is done in a few other places in this PR as well.

CMakeLists.txt
```diff
@@ -70,6 +70,10 @@ option(WHISPER_ALL_WARNINGS_3RD_PARTY "whisper: enable all compiler warnings in
 option(WHISPER_FATAL_WARNINGS  "whisper: enable -Werror flag"               OFF)
 option(WHISPER_USE_SYSTEM_GGML "whisper: use system-installed GGML library" OFF)
 
+# flat bindings
+option(BINDINGS_FLAT      "Add extra flat definitions to Whisper + GGML" OFF)
```
danbev, line 74: Perhaps this should be `WHISPER_BINDINGS_FLAT` to be consistent with the others?

ggml/include/ggml-backend.h
```diff
@@ -348,6 +348,7 @@ extern "C" {
     // CPU buffer types are always available
     GGML_API ggml_backend_buffer_t      ggml_backend_cpu_buffer_from_ptr(void * ptr, size_t size);
     GGML_API ggml_backend_buffer_type_t ggml_backend_cpu_buffer_type(void);
+
```
danbev, line 351: Can you remove this extra line.

src/whisper-flat.h
peardox: They exist precisely owing to the fact that c++ std:vector aren't available in other languages
Externally whisper_state can be treated as an opaque object as long as you have its address so whisper_flat_get_state_from_context can pluck it out of a whisper_context then plug it into whisper_flat_get_activity_with_state to get timing data directly rather than via JSON

CMakeLists.txt
peardox: Yeah - I had that idea after deciding GGML_BINDINGS_FLAT was a good idea

src/whisper-flat.cpp
peardox: Ahh, yeah - probably overkill there :)

I just copied what was in whisper.cpp and added to it (lazy me, sorry)

src/whisper-flat.cpp
peardox: Addressed in next reply - also whisper state is used in other functions - not just the ones I create (I'd have to look em up)

CMakeLists.txt
peardox: Be u0dating everything in a bit - in the middle of a huge all CUDA architecture build _ SYCL test (PCs going slow)

peardox:
Removed superfluous line in ggml-backend.h
Added documentation via comments at end of whisper.cpp explaining placement
Renamed BINDING_FLAT -> WHISPER_BINDINGS_FLAT (in multiple files)
Reduced included in whisper-flat.h
Added params.use_gpu trigger to server + wchess examples if WHISPER_BINDING_FLAT
Checked both with and without -DWHISPER_BINDING_FLAT library exports to ensure extra exposed functions only if enabled
Tested both with and without -DWHISPER_BINDING_FLAT (7 examples modify execution paths with WHISPER_BINDING_FLAT enabled)

peardox:
@danbev Made changes - can you stick it thru CI just to check pls

peardox:
Arggh - ruby failed again
I altered @KitaitiMakoto fix to suit new naming of WHISPER_BINGINGS_FLAT

KitaitiMakoto:
Seems a backup file ( `ggml/include/ggml-backend.h.bak` ) is commited and causes Ruby bindings' test failure.

peardox:
> Seems a backup file ( `ggml/include/ggml-backend.h.bak` ) is commited and causes Ruby bindings' test failure.

Dunno where that came from :(

peardox:
> What is the goal of this change?
> 
First it should be mentioned that WHISPER_BINDINGS_FLAT is specifically designed to only do anything if passed to the build command.
It is entirely conditional, if built without WHISPER_BINDINGS_FLAT the code doesn't exist

The main purpose is to make -DBUILD_SHARED_LIBS=ON -DGGML_BACKEND_DL=ON -DGGML_CPU_ALL_VARIANTS=ON work
If built with those switches no examples will work as the backends are not available when the model is loaded (try it).
This results in null pointers being used to obtain device informtion.

If devices are loaded via e.g. whisper_flat_backend_load_all() or ggml_backend_try_load_best("cpu", nullptr) before the model is loaded this situation is resolved and examples all work.
However, currently, whisper_load_backends also tries to load backends.
This causes double loading of backends or in the case of only loading e.g. cpu forcing the load of all others as well (which may be undesirable)

The rest of the stuff is for non C++ language support. 
C++ specific structures such as std::vector are used freely throughout which causes issues for other languages.
In the case of whisper_init_from_file_with_params state is assigned but is unavailable to non-C++ languages - this is resolved by whisper_flat_get_state_from_context as whisper_state is required extensively
whisper_flat_get_activity_with_state uses a whisper_state to obtain timing (not available as data otherwise)
The whisper_flat...backend functions all deal with getting information about the backends assigned to the current model (so you can tell which ones are actually being used) - this is also unavailable otherwise

PR #3134, opened 2025-05-09
examples : add HEAPU8 to all of the exported runtime methods

enesgrahovac:
This commit adds HEAPU8 to the list of exported methods.

The motivation for this commit is that currently this is causing an error on Window systems where HEAPU8 in undefined, which results in the following error message in the web console:

```
main.js:1 Uncaught TypeError:
Cannot read properties of undefined (reading 'buffer') at __emval_get_property
(main.js:1:1363125) at 003a453a:0xc4a47 at 003a453a:0xc51cd at
Object.full_default (eval at craftInvokerFunction (main.js:1:1347011),
<anonymous>:9:10) at whisper.cpp/:647:42
```

danbev originall fixed this for whisper.wasm, but the issue still exists on the other examples which I patch in this code.

Resolves: https://github.com/ggml-org/whisper.cpp/issues/3059


danbev:
@enesgrahovac Thanks!

PR #3151, opened 2025-05-14
Allow Vino + CoreML files to be in different directories, also optional switch to disable coreml for testing

peardox:
All the data files live in one directory - this gets confusing and hard to manage

This PR enables the openvino and coreml data files to be in other places than the same location as the main model

The parameters may be passed via whisper_context_params as path_coreml and path_openvino. If these are not supplied the current normal all-in-one-place scheme remains but if passed then these directories can be anywhere.

Also I added a disable_coreml parameter. This allows you to build with coreml but disable it on demand - useful for testing and debugging. This is a very simple bypass that merely skips over the coreml init function which has the same practical effect of not having coreml at all

### DIFF
--- a/examples/bench/bench.cpp
+++ b/examples/bench/bench.cpp
@@ -61,7 +61,7 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
     fprintf(stderr, "                           %-7s  2 - ggml_mul_mat\n",                            "");
     fprintf(stderr, "  -ng,      --no-gpu      [%-7s] disable GPU\n",                                 params.use_gpu ? "false" : "true");
     fprintf(stderr, "  -fa,      --flash-attn  [%-7s] enable flash attention\n",                      params.flash_attn ? "true" : "false");
-    fprintf(stderr, "  -ml,      --disable-ml  [%-7s] disable CoreML\n");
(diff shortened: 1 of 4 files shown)


peardox:
Note - the disable_coreml switch -noml in whisper-bench is due to some testing I did yesterday on a M1 Mac 8Gb
Rather usurprisingly ggml-large-v3 was a bit of a strech for that poor low-memory device. The M1 was a remote device - it hung during bench as a result od ANECompiler grabbing all the system resources.

disable_coreml will at least allow selectively switching off CoreML when using a large model with a small device

examples/bench/bench.cpp
```diff
@@ -11,6 +11,9 @@ struct whisper_params {
     int32_t what = 0; // what to benchmark: 0 - whisper encoder, 1 - memcpy, 2 - ggml_mul_mat
 
     std::string model = "models/ggml-base.en.bin";
+    std::string coreml_dir = "";
+    std::string openvino_dir = "";
+    bool disable_coreml = false;
```
ggerganov, line 16: This is not a good change - we should not pass backend-specific info through the `libwhisper` API.

peardox: --ov-e-device in Cli appears rather backend specific?

PR #3163, opened 2025-05-16
Fix CMakeLists.txt to handle deprecated gpu Warnings

jugal-sheth:
Added -Wno-deprecated-gpu-targets instead of -Wno-deprecated


CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ if (EMSCRIPTEN)
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -s TOTAL_STACK=5242880")
     set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -s TOTAL_STACK=5242880")
 
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-deprecated-gpu-targets") # modified flag to avoid deprecated GPU targets warnings
```
danbev, line 47: I think we should keep `-Wno-deprecated` as it is for general language features and not for CUDA specific warnings. 

And then we could add a check for `GGML_CUDA` and set the `-Wno-deprecated-gpu-targets` flag. The reason for this is that otherwise we will get warnings like the following on systems that don't support nvcc, for example on macos I get a number of warnings like the following:
```console
warning: unknown warning option '-Wno-deprecated-gpu-targets'; did you mean '-Wno-deprecated-pragma'? [-Wunknown-warning-option]
1 warning generated.
```

CMakeLists.txt
```diff
@@ -119,6 +119,11 @@ whisper_option_depr(WARNING     WHISPER_SYCL                GGML_SYCL)
 whisper_option_depr(WARNING     WHISPER_SYCL_F16            GGML_SYCL_F16)
 whisper_option_depr(WARNING     WHISPER_CCACHE              GGML_CCACHE)
 
+if (GGML_CUDA)
```
danbev, line 122: I think we also need to exclude MSVC from this as it will otherwise fail with:
```console
 error D8021: invalid numeric argument '/Wno-deprecated-gpu-targets'
 ```
 Some something like this should work:
 ```
 if (GGML_CUDA AND NOT MSVC)
    #GGML_CUDA enabled, add the necessary compile options -Wno-deprecated-gpu-targets
    add_compile_options(-Wno-deprecated-gpu-targets)
endif()
 ```

danbev (approved):

PR #3195, opened 2025-05-28
Install shared libs when using GGML_BACKEND_DL

peardox:
cmake install doesn't install the shared libraries

this fixes most (not all) of them


danbev (approved):

ggerganov:
@danbev This change is currently under discussion: https://github.com/ggml-org/ggml/pull/1240. It might be OK, but there might be an alternative proposed, so keep an eye on that PR.

PR #3200, opened 2025-05-28
Disable log and switch flash_attn to go binding. 

vlegio:
Add disable_log to WHISPER_API, add disable_log and flash_attn switch to go binding

### DIFF
--- a/bindings/go/whisper.go
+++ b/bindings/go/whisper.go
@@ -57,6 +57,7 @@ static struct whisper_full_params whisper_full_default_params_cb(struct whisper_
 	params.progress_callback_user_data = (void*)(ctx);
 	return params;
 }
+
 */
 import "C"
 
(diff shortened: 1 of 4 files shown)


bindings/go/whisper.go
```diff
@@ -99,9 +101,14 @@ var (
 
 // Allocates all memory needed for the model and loads the model from the given file.
 // Returns NULL on failure.
-func Whisper_init(path string) *Context {
+func Whisper_init(path string, flashAttn, disableLog bool) *Context {
```
danbev, line 104: I'm not sure about adding these context parameters like this. Perhaps instead there should be a method that takes `ContextParams` as an argument similar to the [Java binding](https://github.com/ggml-org/whisper.cpp/blob/7fd6fa809749078aa00edf945e959c898f2bd1af/bindings/java/src/main/java/io/github/ggerganov/whispercpp/WhisperCpp.java#L46)?

danbev:
Closing this PR as it seems to have gone "stale". @vlegio Feel free to reopen if this is something you'd like to revisit.

PR #3222, opened 2025-06-03
ruby : non-parallel support

KitaitiMakoto:
Hello,

This pull request adds an option to transcribe in non-parallel.

Thanks.

### DIFF
--- a/bindings/ruby/ext/ruby_whisper_transcribe.cpp
+++ b/bindings/ruby/ext/ruby_whisper_transcribe.cpp
@@ -13,6 +13,7 @@ extern const rb_data_type_t ruby_whisper_params_type;
 
 extern ID id_to_s;
 extern ID id_call;
+extern ID transcribe_option_names[1];
 
 extern void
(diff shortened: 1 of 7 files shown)


bindings/ruby/ext/ruby_whisper_transcribe.cpp
```diff
@@ -66,8 +72,13 @@ ruby_whisper_transcribe(int argc, VALUE *argv, VALUE self) {
 
   prepare_transcription(rwp, &self);
 
-  if (whisper_full_parallel(rw->context, rwp->params, pcmf32.data(), pcmf32.size(), 1) != 0) {
-    fprintf(stderr, "failed to process audio\n");
+  int result;
+  if (parallel) {
+    result = whisper_full_parallel(rw->context, rwp->params, pcmf32.data(), pcmf32.size(), 1);
```
danbev: Should this perhaps be passing a configurable value instead of 1 for the `n_processors` argument to `whisper_full_parallel`? 
It looks like currently this would not cause a parallel invocation.

danbev (approved):

PR #3223, opened 2025-06-03
get_rows & dequantize function implementation for repacked weights of type q4_0 (q4_0x8)

swetha097:
- This implements the GGML_OP_GET_ROWS operation specifically for repacked (block interleaved) 4-bit quantized format (q4_0x8) defined within ggml-cpu-aarch64.cpp

- The following gains were observed by the changes made in the PR - The changes allow for increased usage of the GEMM function (ggml_gemm_q4_0_8x8_q8_0) for q4_0 type defined within ggml-cpu-aarch64.cpp

image

master branch commit - [82f461eaa4e6a1ba29fc0dbdaa415a9934ee8a1d](https://github.com/ggml-org/whisper.cpp/commit/82f461eaa4e6a1ba29fc0dbdaa415a9934ee8a1d)
development (get_rows) branch commit - [d6cc466bd48dd27474ecb00c3baba2e8a887f6c4](https://github.com/swetha097/whisper.cpp/commit/d6cc466bd48dd27474ecb00c3baba2e8a887f6c4)

The PR was tested in AMD Raphael 7600X which supports the following flags :

> system_info: n_threads = 4 / 12 | WHISPER : COREML = 0 | OPENVINO = 0 | CPU : SSE3 = 1 | SSSE3 = 1 | AVX = 1 | AVX2 = 1 | F16C = 1 | FMA = 1 | BMI2 = 1 | AVX512 = 1 | AVX512_VBMI = 1 | AVX512_VNNI = 1 | AVX512_BF16 = 1 | LLAMAFILE = 1 | OPENMP = 1 | AARCH64_REPACK = 1 |  

Model for performance tests Downloaded from : https://huggingface.co/ggerganov/whisper.cpp/blob/main/ggml-base.en.bin and quantized to q4_0

This patch of the code was also tested with llama.cpp repository & the perplexity of Q4_0 models were ensured to be the same before and after the changes made : 

> Final estimate: PPL = 5.9625 +/- 0.03348

Model used for perplexity test quantized from -  https://huggingface.co/meta-llama/Llama-2-7b

### DIFF
--- a/ggml/src/ggml-cpu/ggml-cpu-aarch64.cpp
+++ b/ggml/src/ggml-cpu/ggml-cpu-aarch64.cpp
@@ -6046,6 +6046,9 @@ template <typename BLOC_TYPE, int64_t INTER_SIZE, int64_t NB_COLS, ggml_type PAR
                 size = GGML_PAD(size, sizeof(int64_t));  // + padding for next bloc.
                 size += sizeof(int64_t) * (1+op->src[0]->ne[2]) * op->src[1]->ne[2];
                 return true;
+            case GGML_OP_GET_ROWS:
+                size = 0;  // GET_ROWS (standard and repacked) doesn't need a work buffer
+                return true;
             default:
                 // GGML_ABORT("fatal error");
                 break;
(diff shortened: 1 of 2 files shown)


ggml/src/ggml-cpu/ggml-cpu-aarch64.cpp
```diff
@@ -6061,13 +6064,142 @@ template <typename BLOC_TYPE, int64_t INTER_SIZE, int64_t NB_COLS, ggml_type PAR
...
             default:
                 // GGML_ABORT("fatal error");
                 break;
         }
         return false;
     }
 
+    void forward_get_rows(const ggml_compute_params *params,
+                          ggml_tensor *dst) {
+        const ggml_tensor *src0 = dst->src[0];
```
danbev:
```suggestion
    void forward_get_rows(const ggml_compute_params * params,
                          ggml_tensor * dst) {
        const ggml_tensor * src0 = dst->src[0];
```

ggml/src/ggml-cpu/ggml-cpu-aarch64.cpp
```diff
@@ -6061,13 +6064,142 @@ template <typename BLOC_TYPE, int64_t INTER_SIZE, int64_t NB_COLS, ggml_type PAR
...
+                GGML_ABORT("fatal error");
+                break;
+        }
+    }
+
+    static void ggml_compute_forward_get_rows_q4_0x8(
+        const ggml_compute_params *params,
+        ggml_tensor *dst) {
+        const ggml_tensor *src0 = dst->src[0];
+        const ggml_tensor *src1 = dst->src[1];
```
danbev:
```suggestion
    static void ggml_compute_forward_get_rows_q4_0x8(
        const ggml_compute_params * params,
        ggml_tensor * dst) {
        const ggml_tensor * src0 = dst->src[0];
        const ggml_tensor * src1 = dst->src[1];
```

ggml/src/ggml-cpu/ggml-cpu-aarch64.cpp
```diff
@@ -6061,13 +6064,142 @@ template <typename BLOC_TYPE, int64_t INTER_SIZE, int64_t NB_COLS, ggml_type PAR
...
+     * Dequantizes a single logical row from data repacked with quant interleaving.
+     *
+     * @param p_repacked_group_column_blocks Pointer to the start of 'block_q4_0x8' for the row group.
+     * @param y                              Output buffer for the dequantized float values.
+     * @param k                              Total number of elements (columns) in the logical row.
+     * @param row_idx_in_group               Index (0-7) of the logical row to dequantize.
+     */
+    static void dequantize_row_q4_0x8(
+        const block_q4_0x8 *GGML_RESTRICT p_repacked_group_column_blocks,
+        float *GGML_RESTRICT y,
```
danbev:
```suggestion
    static void dequantize_row_q4_0x8(
        const block_q4_0x8 * GGML_RESTRICT p_repacked_group_column_blocks,
        float * GGML_RESTRICT y,
```

danbev: There are few more of these right leaning pointer in this function which would be nice make make them consistently non-learning (if that is the correct term for this).

ggml/src/ggml-cpu/ggml-cpu-aarch64.cpp
```diff
@@ -6398,12 +6530,23 @@ class extra_buffer_type : ggml::cpu::extra_buffer_type {
             //if (op->src[1]->type == GGML_TYPE_Q8_0) {
             //    return true;
             //}
+        } else if (op->op == GGML_OP_GET_ROWS 
+            && op->src[0]->buffer 
+            && (ggml_n_dims(op->src[0]) == 2) 
+            && op->src[0]->buffer->buft == ggml_backend_cpu_aarch64_buffer_type()
```
danbev:
```suggestion
        } else if (op->op == GGML_OP_GET_ROWS
            && op->src[0]->buffer
            && (ggml_n_dims(op->src[0]) == 2)
            && op->src[0]->buffer->buft == ggml_backend_cpu_aarch64_buffer_type()
```

src/whisper.cpp
```diff
@@ -1437,24 +1437,25 @@ static bool weight_buft_supported(const whisper_hparams & hparams, ggml_tensor *
...
-                ggml_context * ctx = ctx_ptr.get();
+        ggml_context_ptr ctx_ptr { ggml_init(params) };
+        if (!ctx_ptr) {
+            throw std::runtime_error("failed to create ggml context");
+        }
+        ggml_context * ctx = ctx_ptr.get();
 
-                ggml_tensor * op_tensor = nullptr;
+        ggml_tensor * op_tensor = nullptr;
+
```
danbev: Remove learning whitespaces on this line.

danbev (approved):

danbev:
@swetha097 Would you be able to take a look at resolving the conflict reported here?

swetha097:
@danbev   Resolved the conflicts. 
 

- Performance with the gcc compiler is stable and matches the previous results. 
- To see similar performance gains on clang compiler, we still need to pull in the latest updates from the llama.cpp's ggml library.

ggerganov:
@swetha097 Hm, I am a bit confused about why this would make a difference for the performance. The only place that we use `ggml_get_rows` is here:

https://github.com/ggml-org/whisper.cpp/blob/1e72e4bb20b916fe6ec0582d32a7529a39fb0c19/src/whisper.cpp#L2543-L2548

How does this change lead to increased usage of the `ggml_gemm_q4_0_8x8_q8_0`?

swetha097:
@ggerganov 
As noted in whisper-arch.h, 
```
    // Note: ASR_TENSOR_DEC_TOKEN_EMBD_WEIGHT is also used by GGML_OP_MAT_MUL. Need to figure out a way how to handle
    // weight tensors that are used by multiple different operators when extra_buffer_type implementations accelerate
    // more than just GGML_OP_MUL_MAT.
    {ASR_TENSOR_DEC_TOKEN_EMBD_WEIGHT, GGML_OP_GET_ROWS},
```
 
this tensor is used by both GGML_OP_MAT_MUL and GGML_OP_GET_ROWS. Because the original get_rows operation could not handle repacked weights, the tensor remained in a standard format, preventing the GGML_OP_MAT_MUL operation from using its most optimized kernel (ggml_gemm_q4_0_8x8_q8_0), instead it was using gemm4xN previously.

ggerganov:
@swetha097 Thanks, got it.

@eddnjjn Could you help reviewing this change?

ggml/src/ggml-cpu/repack.cpp
```diff
@@ -1405,6 +1413,132 @@ template <typename BLOC_TYPE, int64_t INTER_SIZE, int64_t NB_COLS, ggml_type PAR
...
+
+            GGML_ASSERT(i01 >= 0 && i01 < ne01);
+
+            int row_group_idx = i01 / nrows_interleaved;
+            const int row_idx_in_group = i01 % nrows_interleaved;
+
+            const char * base_ptr_for_higher_dims_in_src0 = (const char *)src0->data + i11 * nb02 + i12 * nb03;
+
+            // Pointer to the first block_q4_0x8 of the identified row_group_idx
+            const block_q4_0x8 * p_first_repacked_block_of_group_x8 = (const block_q4_0x8 *)(base_ptr_for_higher_dims_in_src0 + row_group_idx * stride_between_actual_row_groups);
```
slaren: Wouldn't this fail if the type of this template is actually `block_q4_0x4`?

I think we should add support for testing extra buffer types to `test-backend-ops` before adding more complexity to this code.

ggerganov: Yes, I agree. We'll have to add tests before merging more stuff to the repacking.

ggerganov:
The main concern here is that we don't have a way to test these changes as noted earlier by slaren. The PR should wait until we extend `test-backend-ops` to work with extra buffer types so that we can verify modifications.

PR #3233, opened 2025-06-07
fix compilation with MSVC for arm64 with __ARM_NEON defined 

divideconcept:
These 2 commits fix compilation errors when trying to build with MSVC arm64 with __ARM_NEON defined

### DIFF
--- a/ggml/src/ggml-cpu/ggml-cpu-impl.h
+++ b/ggml/src/ggml-cpu/ggml-cpu-impl.h
@@ -81,7 +81,7 @@ struct ggml_compute_params {
 #define ggml_vld1q_u32(w,x,y,z) { (w), (x), (y), (z) }
 #endif // _MSC_VER
 
-#if !defined(__aarch64__)
+#if !defined(__aarch64__) && !defined(_MSC_VER)
 
 // 32-bit ARM compatibility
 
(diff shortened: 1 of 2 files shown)


ggml/src/ggml-impl.h
```diff
@@ -326,7 +326,7 @@ GGML_API void ggml_aligned_free(void * ptr, size_t size);
 // for old CUDA compilers (<= 11), we use uint16_t: ref https://github.com/ggml-org/llama.cpp/pull/10616
 // for     MUSA compilers        , we use uint16_t: ref https://github.com/ggml-org/llama.cpp/pull/11843
 //
-#if defined(__ARM_NEON) && !(defined(__CUDACC__) && __CUDACC_VER_MAJOR__ <= 11) && !defined(__MUSACC__)
+#if defined(__ARM_NEON) && !(defined(__CUDACC__) && __CUDACC_VER_MAJOR__ <= 11) && !defined(__MUSACC__) && !defined(_MSC_VER)
```
danbev, line 329: Perhaps also update the `#endif` on l line 495 for consistency.

ggml/src/ggml-cpu/ggml-cpu-impl.h
```diff
@@ -81,7 +81,7 @@ struct ggml_compute_params {
 #define ggml_vld1q_u32(w,x,y,z) { (w), (x), (y), (z) }
 #endif // _MSC_VER
 
-#if !defined(__aarch64__)
+#if !defined(__aarch64__) && !defined(_MSC_VER)
```
danbev, line 84: Same here, perhaps this should also update then `#endif` on line 302.

danbev (approved):

ggerganov:
This is a bit complicated to explain, but the `ggml` generally does not support Arm + MSVC. The recommended alternative is Arm + Clang. In short, this fix won't work.

See https://github.com/ggml-org/ggml/pull/1177 and references there.

mirh:
Idk which different instructions couldn't be used here, but https://github.com/ggml-org/llama.cpp/pull/28362 reports whatever MSVC is still missing shouldn't be a big deal.

PR #3234, opened 2025-06-07
Codex/implement macos transcription overlay app

arronjukas:


danbev:
Thanks for your effort here and this does sound like an interesting feature, but I'm afraid I don't think it should live in this project, but instead should live as a separate example, at least for the time being.

I've spent some time trying to get this to work locally but have not been able to get it fully functional. I needed to make the following [changes](https://github.com/danbev/whisper.cpp/tree/macos-transcription-overlay) to get it running, but I was still not able to get transcription to work.

PR #3239, opened 2025-06-10
sync : whisper.cpp

ggerganov:


danbev (approved):

ggerganov:
@xctan Could you recommend a fix for this Windows build:

https://github.com/ggml-org/whisper.cpp/actions/runs/15554099989/job/43790646211?pr=3239#step:6:170

xctan:
> @xctan Could you recommend a fix for this Windows build:
> 
> https://github.com/ggml-org/whisper.cpp/actions/runs/15554099989/job/43790646211?pr=3239#step:6:170

Because apple targets don't have the weak alias feature, all quant kernels are implemented in x86 and arm architectures to avoid linking issues. Thus, for MSVC builds on Windows, weak aliases are not necessary. Maybe we should only enable this macro on other architectures?

xctan:
For the `__cdecl` calling convention of 32-bit x86, a leading underscore is prefixed to the symbol name ([MSVC doc](https://learn.microsoft.com/en-us/cpp/build/reference/decorated-names?view=msvc-170)).

ggerganov:
Adding the underscores seem to resolve the build.

> Thus, for MSVC builds on Windows, weak aliases are not necessary. Maybe we should only enable this macro on other architectures?

I'm not sure - whatever you think is better, let's do that.

xctan:
> all quant kernels are implemented in x86 and arm

Not true for kernels in `repack.cpp`. I think we should keep the current weak alias implementation after examining the failing CI logs.

PR #3251, opened 2025-06-15
let people select the model the wanna use with yt-wsp.sh using a second argument

w1redch4d:
as the title says probably not much "priority feature" but felt can be handy

### DIFF
--- a/examples/yt-wsp.sh
+++ b/examples/yt-wsp.sh
@@ -33,7 +33,7 @@
 #   git clone https://github.com/ggml-org/whisper.cpp
 #   cd whisper.cpp
 #   make
-#   ./examples/yt-wsp.sh https://www.youtube.com/watch?v=1234567890
+#   ./examples/yt-wsp.sh https://www.youtube.com/watch?v=1234567890 <model>
 #
 
 set -Eeuo pipefail
@@ -49,7 +49,7 @@ SCRIPT_DIR="${SCRIPT_PATH%/*}"
 # note: unless a multilingual model is specified, WHISPER_LANG will be ignored
 # and the video will be transcribed as if the audio were in the English language
 ################################################################################
-MODEL_PATH="${MODEL_PATH:-${SCRIPT_DIR}/../models/ggml-base.en.bin}"
+MODEL_PATH="${MODEL_PATH:-${SCRIPT_DIR}/../models/$2}"
 
 ################################################################################
 # Where to find the whisper.cpp executable.  default to the examples directory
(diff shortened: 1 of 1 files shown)


examples/yt-wsp.sh
```diff
@@ -49,7 +49,7 @@ SCRIPT_DIR="${SCRIPT_PATH%/*}"
 # note: unless a multilingual model is specified, WHISPER_LANG will be ignored
 # and the video will be transcribed as if the audio were in the English language
 ################################################################################
-MODEL_PATH="${MODEL_PATH:-${SCRIPT_DIR}/../models/ggml-base.en.bin}"
+MODEL_PATH="${MODEL_PATH:-${SCRIPT_DIR}/../models/$2}"
```
danbev: I think the original intention is correct here as this is a default value for `MODEL_PATH` if the environment variable for it has not been set (or is empty).

examples/yt-wsp.sh
```diff
@@ -33,7 +33,7 @@
 #   git clone https://github.com/ggml-org/whisper.cpp
 #   cd whisper.cpp
 #   make
-#   ./examples/yt-wsp.sh https://www.youtube.com/watch?v=1234567890
+#   ./examples/yt-wsp.sh https://www.youtube.com/watch?v=1234567890 <model>
```
danbev: I think the intention here is that environment variables should be used to set the model and not that the model be passed as a command line option.

w1redch4d: then would it be better if i set an example using the environment variable commands for anyone to just glance at and understand whats going on?

examples/yt-wsp.sh
w1redch4d: Then i guess i should revert it back and add a comment as an example of how to use the script cuz for most people opening a script to see the configurable environment variables might be an hassel so maybe should add the usage with the env vars in the `print_help` function?

danbev: Yes, I don't think this should change. 
But clarifying with an example in the `print_help` sounds like a good idea :+1:

examples/yt-wsp.sh
```diff
@@ -33,7 +33,7 @@
 #   git clone https://github.com/ggml-org/whisper.cpp
 #   cd whisper.cpp
 #   make
-#   ./examples/yt-wsp.sh https://www.youtube.com/watch?v=1234567890 <model>
+#   MODEL_PATH=<model> WHISPER_EXECUTABLE=<path_to_whisper-cli> WHISPER_LANG=en WHISPER_LANG=en WHISPER_THREAD_COUNT=<threads> ./examples/yt-wsp.sh https://www.youtube.com/watch?v=1234567890
```
danbev: I don't think it is necessary to include all the environment variable options where now that the usage/print_help includes them.

PR #3255, opened 2025-06-16
backport CoreML features to macos < 14

glaszig:
i'm on 12.7 and `[MLModel predictionFromFeatures:]` versions with completion handler are [only available on macos 14](https://developer.apple.com/documentation/coreml/mlmodel/predictionfromfeatures:completionhandler:?language=objc) which is a little too cutting edge for me.

before this i'd get:
```
whisper-encoder-impl.m:183:17: error: no visible @interface for 'MLModel'
declares the selector 'predictionFromFeatures:completionHandler:'
    [self.model predictionFromFeatures:input completionHandler:^(id<MLFeatureProvider> prediction, NSError *predictionError) {
     ~~~~~~~~~~ ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -58,6 +58,7 @@ if (WHISPER_COREML)
     set(TARGET whisper.coreml)
 
     add_library(${TARGET}
+        coreml/whisper-compat.m
         coreml/whisper-encoder.h
         coreml/whisper-encoder.mm
         coreml/whisper-encoder-impl.h
@@ -76,6 +77,7 @@ if (WHISPER_COREML)
         COMPILE_FLAGS "-fobjc-arc"
         XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES
         )
+
     set_target_properties(${TARGET} PROPERTIES FOLDER "libs")
 endif()
 
--- a/src/coreml/whisper-compat.m
+++ b/src/coreml/whisper-compat.m
@@ -1,0 +1,35 @@
+#import "whisper-compat.h"
+#import <Foundation/Foundation.h>
+
+@implementation MLModel (Compat)
+
+#if !defined(MAC_OS_X_VERSION_14_00) || MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_14_00
+
+- (void) predictionFromFeatures:(id<MLFeatureProvider>) input
+              completionHandler:(void (^)(id<MLFeatureProvider> output, NSError * error)) completionHandler {
+    [NSOperationQueue.new addOperationWithBlock:^{
+        NSError *error = nil;
+        id<MLFeatureProvider> prediction = [self predictionFromFeatures:input error:&error];
+
+        [NSOperationQueue.mainQueue addOperationWithBlock:^{
+            completionHandler(prediction, error);
+        }];
+    }];
+}
+
+- (void) predictionFromFeatures:(id<MLFeatureProvider>) input
+                        options:(MLPredictionOptions *) options
+              completionHandler:(void (^)(id<MLFeatureProvider> output, NSError * error)) completionHandler {
+    [NSOperationQueue.new addOperationWithBlock:^{
+        NSError *error = nil;
+        id<MLFeatureProvider> prediction = [self predictionFromFeatures:input options:options error:&error];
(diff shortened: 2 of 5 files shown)


src/coreml/MLModel+Compat.m
```diff
@@ -0,0 +1,29 @@
...
+
+@implementation MLModel (Compat)
+
+#if !defined(MAC_OS_X_VERSION_14_00) || MAC_OS_X_VERSION_MAX_ALLOWED < MAC_OS_X_VERSION_14_00
+
+- (void) predictionFromFeatures:(id<MLFeatureProvider>) input
+              completionHandler:(void (^)(id<MLFeatureProvider> output, NSError * error)) completionHandler {
+   [NSOperationQueue.mainQueue addOperationWithBlock:^{
+        NSError *error = nil;
+        id<MLFeatureProvider> prediction = [self predictionFromFeatures:input error:&error];
```
danbev: Should the prediction/inference perhaps be run on a background thread and not the main thread as it being done here?

src/coreml/MLModel+Compat.m
```diff
@@ -0,0 +1,29 @@
...
+        completionHandler(prediction, error);
+    }];
+}
+
+- (void) predictionFromFeatures:(id<MLFeatureProvider>) input
+                        options:(MLPredictionOptions *) options
+              completionHandler:(void (^)(id<MLFeatureProvider> output, NSError * error)) completionHandler {
+   [NSOperationQueue.mainQueue addOperationWithBlock:^{
+        NSError *error = nil;
+        id<MLFeatureProvider> prediction = [self predictionFromFeatures:input options:options error:&error];
```
danbev: Should the prediction/inference perhaps be run on a background thread and not the main thread as it being done here?

src/coreml/MLModel+Compat.m
glaszig: that was my first impulse as well but the original documentation doesn't state anything about parallelism/threading. then it occurred to me that coreml itself decides where to run (cpu or gpu) [0] and the `completionHandler` variants are only to enable swift's async/await [1]. 

that is why i decided against active concurrency management in those methods. async-marked methods in swift actually get their handler-equivalent auto-generated. a difference may be that they use dispatch queues instead of operations though.

[0]: https://developer.apple.com/videos/play/wwdc2023/10049/?time=781
[1]: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0297-concurrency-objc.md#asynchronous-completion-handler-methods

danbev: I'm just a little concerned that doing the inference on the main thread might possibly lock/stall an application while the inference is running which could be substantial. Would you be able to try something like the following and see if this works:
```objective-c
[NSOperationQueue.new addOperationWithBlock:^{
    id<MLFeatureProvider> prediction = [self predictionFromFeatures:input error:&error];
    
    dispatch_async(dispatch_get_main_queue(), ^{
        completionHandler(prediction, error);
    });
}];
````

**Review by danbev — APPROVED** [review]

File: src/CMakeLists.txt [hunk 3]
```diff
@@ -58,6 +58,7 @@ if (WHISPER_COREML)
     set(TARGET whisper.coreml)
 
     add_library(${TARGET}
+        coreml/MLModel+Compat.m
```
ggerganov: Better rename this file to `coreml/whisper-compat.m` for consistency

glaszig:
done. also squashed commits.

src/coreml/whisper-compat.m
```diff
@@ -0,0 +1,35 @@
+#import "MLModel+Compat.h"
```
danbev: I think this needs to be updated with the changed header name:
```suggestion
#import "whisper-compat.h"
```
This currently generates a compilation error:
```console
/Users/danbev/work/ai/whisper-work/src/coreml/whisper-compat.m:1:9: fatal error: 'MLModel+Compat.h' file not found
    1 | #import "MLModel+Compat.h"
      |         ^~~~~~~~~~~~~~~~~~
1 error generated.
```

glaszig:
oops. fixed.

danbev (approved):

PR #3257, opened 2025-06-16
feat(server): refactor concurrency model and enhance task management

sachaarbonel:


examples/server/server.cpp
```diff
@@ -41,10 +43,10 @@ const std::string vjson_format  = "verbose_json";
 const std::string vtt_format    = "vtt";
 
 std::function<void(int)> shutdown_handler;
-std::atomic_flag is_terminating = ATOMIC_FLAG_INIT;
```
danbev, line 44: I think that `std::atomic_flag` might be more appropriate here as it is guaranteed to be lockfree. And it is only used in a signal handler where we don't want any potential blocking. So I'd prefer to keep this as is.

examples/server/server.cpp
```diff
@@ -603,14 +608,451 @@ void get_req_parameters(const Request & req, whisper_params & params)
...
+
+struct WhisperTask {
+    std::vector<float> pcmf32;
+    std::vector<std::vector<float>> pcmf32s;
+    whisper_params params;
+    std::string filename;
+    const httplib::Request* request_ptr; // For abort callback
+    std::atomic<AbortReason> abort_reason{AbortReason::NotAborted};
+    std::atomic<bool>* stop_flag_ptr{nullptr};
+
```
danbev, line 633: Nit: remove empty space (not showing here but is visible using command line git). There are few more of these in the file but I won't list them all to avoid cluttering the review.

examples/server/server.cpp
```diff
@@ -603,14 +608,451 @@ void get_req_parameters(const Request & req, whisper_params & params)
...
+    
+    // Delete copy constructor and copy assignment
+    WhisperTask(const WhisperTask&) = delete;
+    WhisperTask& operator=(const WhisperTask&) = delete;
+    
+    // Default constructor
+    WhisperTask() = default;
+    
+    // Move constructor
+    WhisperTask(WhisperTask&& other) noexcept
```
danbev, line 648: Nit: remove trailing white space (not showing here but is visible using command line git). There are few more of these in the file but I won't list them all to avoid cluttering the review.

examples/server/server.cpp
```diff
@@ -603,14 +608,451 @@ void get_req_parameters(const Request & req, whisper_params & params)
...
+    std::string content_type = "application/json";
+    int status_code = 200;
+};
+
+struct WhisperTask {
+    std::vector<float> pcmf32;
+    std::vector<std::vector<float>> pcmf32s;
+    whisper_params params;
+    std::string filename;
+    const httplib::Request* request_ptr; // For abort callback
```
danbev, line 630: Nit: Pointers/references are preferred to be "non-leaning". There are few more of these.

examples/server/server.cpp
```diff
@@ -603,14 +608,451 @@ void get_req_parameters(const Request & req, whisper_params & params)
...
+          stop_flag_ptr(other.stop_flag_ptr),
+          result_mutex(),
+          result_cv(),
+          result(std::move(other.result)),
+          completed(other.completed.load()) {}
+};
+
+class WhisperTaskQueue {
+public:
+    WhisperTaskQueue(struct whisper_context* ctx, size_t n_workers = 2)
```
danbev, line 664: Just wondering about the default of `2` for `n_workers` as the default value is 1 elsewhere in the code and perhaps this should be also be 1?

examples/server/server.cpp
```diff
@@ -603,14 +608,451 @@ void get_req_parameters(const Request & req, whisper_params & params)
...
+                wparams.no_speech_thold = task.params.no_speech_thold;
+                wparams.temperature_inc  = task.params.temperature_inc;
+                wparams.entropy_thold    = task.params.entropy_thold;
+                wparams.logprob_thold    = task.params.logprob_thold;
+
+                wparams.no_timestamps    = task.params.no_timestamps;
+                wparams.token_timestamps = !task.params.no_timestamps && task.params.response_format == vjson_format;
+                wparams.no_context       = task.params.no_context;
+                wparams.suppress_nst     = task.params.suppress_nst;
+
```
danbev, line 857: Perhaps also add VAD parameters here similar to what is done in https://github.com/ggml-org/whisper.cpp/blob/2a4d6db7d90899aff3d58d70996916968e4e0d27/examples/server/server.cpp#L924-L932

examples/server/server.cpp
```diff
@@ -603,14 +608,451 @@ void get_req_parameters(const Request & req, whisper_params & params)
...
+        return result;
+    }
+
+    struct whisper_context* ctx_;
+    std::vector<std::thread> workers_;
+    std::queue<std::shared_ptr<WhisperTask>> tasks_;
+    std::mutex queue_mutex_;
+    std::condition_variable queue_cv_;
+    std::mutex whisper_mutex_; // Protect whisper context access
+    std::atomic<bool> stop_flag_;
```
danbev, line 1047: I think `stop_flag_` could just be a bool as the accesses to it are all guarded by the `queue_cv_` lock.

danbev (commented):
I've done an initial pass over this and wanted to leave some early feedback (or feedback in chucks at least so there are not too many in one go) and I'll go through this in more detail.

sachaarbonel:
Lte's hold on this one. I ran some becnhmarks and tracing with NVIDIA Nsight and CPU is not the bottleneck

PR #3265, opened 2025-06-18
cmake : fix android build

ggerganov:
cont #3264


ggerganov:
@danbev Thanks! We eventually have to fix the android build to use the existing CMake system instead of this ad hoc build approach:

https://github.com/ggml-org/whisper.cpp/blob/2c823bd20b929b6d1a2cf957f56bca08268602fd/examples/whisper.android/lib/src/main/jni/whisper/CMakeLists.txt#L16-L18

PR #3273, opened 2025-06-21
`whisper-command`: write commands to text file

aaron-ang:
I noticed that `whisper-command` has the `-f --file FNAME` option but does not use it in the program. I added some code to open and write to the file if provided, using similar syntax as in `whisper-stream`.

### DIFF
--- a/examples/command/command.cpp
+++ b/examples/command/command.cpp
@@ -251,7 +251,7 @@ static std::vector<std::string> get_words(const std::string &txt) {
 
 // command-list mode
 // guide the transcription to match the most likely command from a provided list
-static int process_command_list(struct whisper_context * ctx, audio_async &audio, const whisper_params &params) {
+static int process_command_list(struct whisper_context * ctx, audio_async &audio, const whisper_params &params, std::ofstream &fout) {
     fprintf(stderr, "\n");
     fprintf(stderr, "%s: guided mode\n", __func__);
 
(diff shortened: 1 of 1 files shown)


danbev (approved):

examples/command/command.cpp
```diff
@@ -528,6 +532,9 @@ static int always_prompt_transcription(struct whisper_context * ctx, audio_async
 
                 if ((sim > 0.7f) && (command.size() > 0)) {
                     fprintf(stdout, "%s: Command '%s%s%s', (t = %d ms)\n", __func__, "\033[1m", command.c_str(), "\033[0m", (int) t_ms);
+                    if (fout.is_open()) {
+                        fout << command.c_str() << std::endl;
```
ggerganov:
```suggestion
                        fout << command << std::endl;
```

examples/command/command.cpp
```diff
@@ -662,8 +669,10 @@ static int process_general_transcription(struct whisper_context * ctx, audio_asy
                     } else {
                         // cut the prompt from the decoded text
                         const std::string command = ::trim(txt.substr(best_len));
-
                         fprintf(stdout, "%s: Command '%s%s%s', (t = %d ms)\n", __func__, "\033[1m", command.c_str(), "\033[0m", (int) t_ms);
+                        if (fout.is_open()) {
+                            fout << command.c_str() << std::endl;
```
ggerganov:
```suggestion
                            fout << command << std::endl;
```

ggerganov (approved):

PR #3274, opened 2025-06-21
chore: align VAD model path naming with ggml prefix and update readme.md

MUCZ:
- Updated the VAD model download path to use names starting with the ggml prefix (currently the shell script downloads models this way, but the Python script wasn't updated — this MR unifies the path naming).

- Updated the model path in the VAD usage section of the README accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -781,7 +781,7 @@ This model can be also be converted manually to ggml using the following command
 $ python3 -m venv venv && source venv/bin/activate
 $ (venv) pip install silero-vad
 $ (venv) $ python models/convert-silero-vad-to-ggml.py --output models/silero.bin
-Saving GGML Silero-VAD model to models/silero-v5.1.2-ggml.bin
+Saving GGML Silero-VAD model to models/ggml-silero-v5.1.2.bin
 ```
 And it can then be used with whisper as follows:
 ```console
@@ -789,7 +789,7 @@ $ ./build/bin/whisper-cli \
    --file ./samples/jfk.wav \
    --model ./models/ggml-base.en.bin \
    --vad \
-   --vad-model ./models/silero-v5.1.2-ggml.bin
+   --vad-model ./models/ggml-silero-v5.1.2.bin
 ```
 
 ### VAD Options


models/convert-silero-vad-to-ggml.py
```diff
@@ -20,7 +20,7 @@ def convert_silero_vad(output_path, print_tensors=True):
             cleaned_dict[clean_key] = value
 
     base, ext = os.path.splitext(output_path)
-    output_file = f"{base}-v{silero_version}-ggml{ext}"
+    output_file = f"ggml-{base}-v{silero_version}{ext}"
```
danbev, line 23: I think the `ggml` prefix should come before the file name and not before the path to the models directory:
```suggestion
    output_file = f"{base}ggml-v{silero_version}{ext}"
```

danbev: I think that we should perhaps change this to have a default value for the `--output` parameter, and then just allow it to be overridden by anything that the user wants. I'm not sure why I did not do that at the time.
Something like:
```console
diff --git a/models/convert-silero-vad-to-ggml.py b/models/convert-silero-vad-to-ggml.py
index 9d3afd4c..b9fb549e 100644
--- a/models/convert-silero-vad-to-ggml.py
+++ b/models/convert-silero-vad-to-ggml.py
@@ -19,16 +19,15 @@ def convert_silero_vad(output_path, print_tensors=True):
                 clean_key = "_model." + key
             cleaned_dict[clean_key] = value
 
-    base, ext = os.path.splitext(output_path)
-    output_file = f"ggml-{base}-v{silero_version}{ext}"
-    print(f"Saving GGML Silero-VAD model to {output_file}")
+    print(f"Saving GGML Silero-VAD model to {output_path}")
 
     print("\nTensor info for debugging:")
     for key, tensor in cleaned_dict.items():
         print(f"  - {key}: {tensor.shape} ({tensor.dtype})")
     print()
 
-    with open(output_file, "wb") as fout:
+    print(f"output_file: {output_path}")
+    with open(output_path, "wb") as fout:
         # Write magic and version
         fout.write(struct.pack("i", 0x67676d6c))
 
@@ -184,12 +183,12 @@ def convert_silero_vad(output_path, print_tensors=True):
 
             print(f"  Wrote {data.size * (2 if ftype==1 else 4)} bytes")
 
-    print(f"\nDone! Model has been converted to GGML format: {output_file}")
-    print(f"File size: {os.path.getsize(output_file)} bytes")
+    print(f"\nDone! Model has been converted to GGML format: {output_path}")
+    print(f"File size: {os.path.getsize(output_path)} bytes")
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Convert Silero-VAD PyTorch model to GGML format")
-    parser.add_argument("--output", type=str, required=True, help="Path to output GGML model file")
+    parser.add_argument("--output", type=str, default=f"models/ggml-v{silero_version}.bin", help="Path to output GGML model file")
     parser.add_argument("--print-tensors", action="store_true", help="Print tensor values", default=True)
     args = parser.parse_args()

```
This would allow the conversion to be run as:
```console
(venv) $ python models/convert-silero-vad-to-ggml.py
Saving GGML Silero-VAD model to models/ggml-v5.1.2.bin
...
```

PR #3313, opened 2025-07-09
If backends fail it is possible to try using a null pointer

peardox:
This is, hopefully, slighly cleaner than a core dump

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -1,4 +1,4 @@
-`#include "whisper.h"
+#include "whisper.h"
 #include "whisper-arch.h"
 
 #include "ggml.h"
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -1,4 +1,4 @@
-#include "whisper.h"
+`#include "whisper.h"
 #include "whisper-arch.h"
 
 #include "ggml.h"
@@ -1326,6 +1326,10 @@ static ggml_backend_t whisper_backend_init_gpu(const whisper_context_params & pa
     if (params.use_gpu) {
         for (size_t i = 0; i < ggml_backend_dev_count(); ++i) {
             ggml_backend_dev_t dev_cur = ggml_backend_dev_get(i);
+            // Prevent null pointer being used in following code
+            if(dev_cur == nullptr) {
+                continue;
+            }
             if (ggml_backend_dev_type(dev_cur) == GGML_BACKEND_DEVICE_TYPE_GPU) {
                 if (cnt == 0 || cnt == params.gpu_device) {
                     dev = dev_cur;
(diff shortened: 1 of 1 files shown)


src/whisper.cpp
```diff
@@ -1326,6 +1326,10 @@ static ggml_backend_t whisper_backend_init_gpu(const whisper_context_params & pa
     if (params.use_gpu) {
         for (size_t i = 0; i < ggml_backend_dev_count(); ++i) {
             ggml_backend_dev_t dev_cur = ggml_backend_dev_get(i);
+            // Prevent null pointer being used in following code
+            if(dev_cur == nullptr) {
```
danbev: I don't think any of the backend implementation of `ggml_backend_dev_type` currently use the `dev` parameter so these checks should not really be required unless there are changes to the ggml backends (or additional backends that use the `dev` parameter are added). While it does not hurt to add these, this would differ from how this is done in llama.cpp and keeping consistency might be good here. 

Nit: If we do add these check please add a space after the `if` keyword.

src/whisper.cpp
```diff
@@ -1412,20 +1424,23 @@ static buft_list_t make_buft_list(whisper_context_params & params) {
...
-    auto * cpu_reg = ggml_backend_dev_backend_reg(cpu_dev);
-    auto get_extra_bufts_fn = (ggml_backend_dev_get_extra_bufts_t)
-        ggml_backend_reg_get_proc_address(cpu_reg, "ggml_backend_dev_get_extra_bufts");
-    if (get_extra_bufts_fn) {
-        ggml_backend_buffer_type_t * extra_bufts = get_extra_bufts_fn(cpu_dev);
-        while (extra_bufts && *extra_bufts) {
-            buft_list.emplace_back(cpu_dev, *extra_bufts);
-            ++extra_bufts;
+    // Prevent null pointer being used in following code
+    if(cpu_dev != nullptr) {
```
danbev: Nit: Can you add a space after the `if` keyword.

danbev (commented):
I'm not sure if we should add the nullptr checks as mentioned in the comments but I don't feel strongly about it. But the other changes look good to me.

src/whisper.cpp
peardox: Added space after if *4 to fix nit

The most easiest way to trigger the state this avoids is by commenting out the ggml_backend_load_all() in whisper-bench and build shared all_cpu_variants. OK that error would be pretty stupid but at least with this it'll tell you there are no backends. The alternative is a core-dump which tells you less than zero

I know most perople will build static for the build-box but if you want to pass something around as a release shared all_cpu_variants is a must

Also if you pass a static release around you run into the issue of it not loading the right CPU on older kit (probably same result - not tested - yet, only just though of it)

PR #3322, opened 2025-07-13
Support static xcframework packaging in build-xcframework.sh

richwaters:
This commit allows for the building of a static xcframework by adding a BUILD_STATIC_XCFRAMEWORK option. When enabled, the build-xcframework.sh script builds a  static whisper.xcframework.

The motivation for this change is so that command line tools can link whisper.cpp without forcing users to install the whisper.xcframework separately.

### DIFF
--- a/build-xcframework.sh
+++ b/build-xcframework.sh
@@ -15,6 +15,7 @@ GGML_METAL_EMBED_LIBRARY=ON
 GGML_BLAS_DEFAULT=ON
 GGML_METAL_USE_BF16=ON
 GGML_OPENMP=OFF
+BUILD_STATIC_XCFRAMEWORK=OFF
 
 COMMON_C_FLAGS="-Wno-macro-redefined -Wno-shorten-64-to-32 -Wno-unused-command-line-argument -g"
 COMMON_CXX_FLAGS="-Wno-macro-redefined -Wno-shorten-64-to-32 -Wno-unused-command-line-argument -g"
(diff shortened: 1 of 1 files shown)


build-xcframework.sh
```diff
@@ -15,6 +15,7 @@ GGML_METAL_EMBED_LIBRARY=ON
 GGML_BLAS_DEFAULT=ON
 GGML_METAL_USE_BF16=ON
 GGML_OPENMP=OFF
+BUILD_STATIC_XCFRAMEWORK=OFF
```
danbev: Perhaps we could make this variable overridable by an environment variable:
```suggestion
BUILD_STATIC_XCFRAMEWORK=${BUILD_STATIC_XCFRAMEWORK:-OFF}
```

build-xcframework.sh
```diff
@@ -529,6 +539,20 @@ combine_static_libraries "build-tvos-device" "Release-appletvos" "tvos" "false"
 
 # Create XCFramework with correct debug symbols paths
 echo "Creating XCFramework..."
+
+    if [[ "${BUILD_STATIC_XCFRAMEWORK}" == "ON" ]]; then
```
danbev: Can you remove the indentation here so that is aligns with the rest of the script?

richwaters: OK. I think I've made that change and managed to push it. (The extra indentation was left over from an earlier attempt at the change that had the code in a different place.)

build-xcframework.sh
```diff
@@ -540,18 +540,18 @@ combine_static_libraries "build-tvos-device" "Release-appletvos" "tvos" "false"
...
+        -framework $(pwd)/build-ios-sim/framework/whisper.framework \
+        -framework $(pwd)/build-ios-device/framework/whisper.framework \
+        -framework $(pwd)/build-macos/framework/whisper.framework \
+        -framework $(pwd)/build-visionos/framework/whisper.framework \
+        -framework $(pwd)/build-visionos-sim/framework/whisper.framework \
+        -framework $(pwd)/build-tvos-device/framework/whisper.framework \
+        -framework $(pwd)/build-tvos-sim/framework/whisper.framework \
+        -output $(pwd)/build-apple/whisper.xcframework
+    exit 0
+fi
```
danbev: Nit: can you remove the extra whitespace after `fi` (does not show up in github but is visible using a local git diff). You can add `[no ci]` to the commit to avoid running the full CI again.

build-xcframework.sh
```diff
@@ -327,6 +328,15 @@ combine_static_libraries() {
         arch_flags+=" -arch $arch"
     done
 
+
```
danbev: Nit: can you remove the extra whitespace on this line (does not show up in github but is visible using a local git diff). You can add [no ci] to the commit to avoid running the full CI again.

PR #3328, opened 2025-07-18
feat(server): hide language probabilities option behind flag

sachaarbonel:
# Add language probabilities feature flag to reduce latency

Adds a feature flag to make language probabilities optional in the verbose JSON output format. Computing language probabilities is an expensive operation that adds considerable latency to responses.

Changes:
- Added `language_probabilities` boolean flag (default: `true`)
- Added CLI flags: `-nlp`, `--no-language-probabilities`
- Added HTTP parameter: `no_language_probabilities`
- Modified JSON response to conditionally include probabilities

## Usage

Command line:
```bash
# Disable language probabilities for faster response
./server --model models/ggml-base.en.bin -nlp false
```

HTTP API:
```bash
curl 127.0.0.1:8080/inference \
  -F file="@audio.wav" \
  -F response_format="verbose_json" \
  -F no_language_probabilities="false"
```

### DIFF
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -104,7 +104,7 @@ struct whisper_params {
     bool flash_attn      = false;
     bool suppress_nst    = false;
     bool no_context      = false;
-    bool no_no_language_probabilities = true;
+    bool no_language_probabilities = false;
 
     std::string language        = "en";
     std::string prompt          = "";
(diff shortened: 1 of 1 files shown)


examples/server/server.cpp
```diff
@@ -237,6 +239,7 @@ bool whisper_params_parse(int argc, char ** argv, whisper_params & params, serve
         else if (arg == "-sns"  || arg == "--suppress-nst")    { params.suppress_nst    = true; }
         else if (arg == "-nth"  || arg == "--no-speech-thold") { params.no_speech_thold = std::stof(argv[++i]); }
         else if (arg == "-nc"   || arg == "--no-context")      { params.no_context      = true; }
+        else if (arg == "-lp" || arg == "--language-probabilities") { params.language_probabilities = true; }
```
danbev: Should this perhaps be `-nlp/--no-language-probabilities` instead as it currently does not seem to be possible to have an argument for the `-lp/--language-probabilities` option? 
This pattern is also used for other options like `no-context`, `no-gpu` etc?

Nit: Can you align the logical OR operator with the ones above here?

examples/server/server.cpp
```diff
@@ -255,6 +258,7 @@ bool whisper_params_parse(int argc, char ** argv, whisper_params & params, serve
         else if (arg == "-vmsd" || arg == "--vad-max-speech-duration-s")   { params.vad_max_speech_duration_s   = std::stof(argv[++i]); }
         else if (arg == "-vp"   || arg == "--vad-speech-pad-ms")           { params.vad_speech_pad_ms           = std::stoi(argv[++i]); }
         else if (arg == "-vo"   || arg == "--vad-samples-overlap")         { params.vad_samples_overlap         = std::stof(argv[++i]); }
+        else if (arg == "--language-probabilities") { params.language_probabilities = true; }
```
danbev: Nit: Can you align the setting of `params.language_probabilties` to match the ones above (so the curly braces are aligned)?

examples/server/server.cpp
```diff
@@ -1024,25 +1032,32 @@ int main(int argc, char ** argv) {
         } else if (params.response_format == vjson_format) {
             /* try to match openai/whisper's Python format */
             std::string results = output_str(ctx, params, pcmf32s); 
-            // Get language probabilities
-            std::vector<float> lang_probs(whisper_lang_max_id() + 1, 0.0f);
-            const auto detected_lang_id = whisper_lang_auto_detect(ctx, 0, params.n_threads, lang_probs.data());
+
```
danbev: Nit: Remove additional whitespace at the start of this line (does not show up in github diff).
There are a few more of these on line 1043, 1048, 1052, and 1060

PR #3336, opened 2025-07-23
feat(server): add warmup file parameter for model initialization

sachaarbonel:
### Summary
Adds a CLI flag `--warmup-file` to the whisper.cpp server that allows running a transcription on a specified audio file during startup to warm up the model before serving requests.

### Changes
- Added `warmup_file` parameter to `whisper_params` struct
- Added `-wf/--warmup-file` CLI flag to specify path to warmup audio file
- Implemented warmup logic that runs a quick transcription during model initialization
- Added help text documentation for the new flag

### Implementation Details
- Warmup runs after model initialization but before server starts accepting requests
- Uses optimized parameters for fast warmup (single segment, no context, minimal audio context)
- Provides timing feedback showing how long the warmup took
- Gracefully handles warmup failures with warning messages
- Only executes warmup when `--warmup-file` flag is provided

### Usage
```bash
./whisper-server --warmup-file samples/jfk.wav --model models/ggml-base.en.bin
```

### Benefits
- Reduces latency for first transcription request
- Ensures model is fully loaded and optimized before serving
- Useful for production deployments where consistent response times are important

This change is backward compatible and doesn't affect existing functionality when the warmup flag is not used.

### DIFF
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -129,7 +129,6 @@ struct whisper_params {
     float       vad_max_speech_duration_s = FLT_MAX;
     int         vad_speech_pad_ms = 30;
     float       vad_samples_overlap = 0.1f;
-
     // Warmup parameters
     std::string warmup_file = "";
 };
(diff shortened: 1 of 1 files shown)


ggerganov:
This warmup logic can be delegated to the script that starts the server process. After it starts the server, simply send a transcription request with a short audio file to do the warmup. This is a better alternative because we won't have to maintain this logic.

PR #3365, opened 2025-08-06
ruby : Add ruby binding for max_len

adamdebono:
The `max_len`/`ml` parameter was missing in the Ruby binding.


bindings/ruby/ext/ruby_whisper_params.c
```diff
@@ -1271,6 +1300,7 @@ init_ruby_whisper_params(VALUE *mWhisper)
   DEFINE_PARAM(suppress_blank, 8)
   DEFINE_PARAM(suppress_nst, 9)
   DEFINE_PARAM(token_timestamps, 10)
+  DEFINE_PARAM(max_len, 35)
```
KitaitiMakoto: Can you re-number params for maintainability? The position of `max_len` is okay and keep it here.

adamdebono: Ah yep I was wondering what was appropriate here

adamdebono:
Thanks @KitaitiMakoto I've made those changes.

I wasn't able to get the tests running so I haven't confirmed if they pass, although I've reused the same code as the other tests so I assume they do.

I am attempting to use `rake test` as is mentioned in the readme, but I'm getting a segfault. I've attached the log, but it looks like the problem is in `test_callback.rb#125` - I had noticed I was getting a segfault in my application if I attempted to set that multiple times, presumably this is the same problem.

[rake_test.log](https://github.com/user-attachments/files/21640952/rake_test.log)

KitaitiMakoto (approved):

KitaitiMakoto:
Ah, that segfault is caused by [Ruby 3.4.1's bug](https://bugs.ruby-lang.org/issues/20981). Tests should succeed for Ruby 3.4.2 or later.

Your code has no problem. Will merge it.

PR #3381, opened 2025-08-18
Update `./models/download-ggml-model.cmd` to allow for tdrz download

GalFawkes:
Added a couple of lines to validate whether `tdrz` is in the argument much like the shell version of this script. Uses findstr (built into cmd.exe) as the grep alternative and `%ERRORLEVEL%` to validate as in the next step., which resolves error checking correctly as well.

I think this introduces an echo of the downloaded model's name in the output, but that's a single line & negligible.

### DIFF
--- a/models/download-ggml-model.cmd
+++ b/models/download-ggml-model.cmd
@@ -78,7 +78,8 @@ echo %model% | findstr tdrz
 if %ERRORLEVEL% neq 0 (
  PowerShell -NoProfile -ExecutionPolicy Bypass -Command "Start-BitsTransfer -Source https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-%model%.bin -Destination \"%models_path%\\ggml-%model%.bin\""
 ) else (
-  PowerShell -NoProfile -ExecutionPolicy Bypass -Command "Start-BitsTransfer -Source https://huggingface.co/akashmjn/tinydiarize-whisper.cpp/resolve/main/ggml-%model%.bin -Destination \"%models_path%\\ggml-%model%.bin\"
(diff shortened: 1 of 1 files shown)


models/download-ggml-model.cmd
```diff
@@ -74,8 +74,12 @@ if exist "%models_path%\\ggml-%model%.bin" (
   echo Model %model% already exists. Skipping download.
   goto :eof
 )
-
-PowerShell -NoProfile -ExecutionPolicy Bypass -Command "Start-BitsTransfer -Source https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-%model%.bin -Destination \"%models_path%\\ggml-%model%.bin\""
+@echo %model% | @findstr tdrz
+if %ERRORLEVEL% neq 0 (
+ PowerShell -NoProfile -ExecutionPolicy Bypass -Command "Start-BitsTransfer -Source https://huggingface.co/ggerganov/whisper.cpp/resolve/main/ggml-%model%.bin -Destination \"%models_path%\\ggml-%model%.bin\""
+) else (
+  PowerShell -NoProfile -ExecutionPolicy Bypass -Command "Start-BitsTransfer -Source https://huggingface.co/akashmjn/tinydiarize-whisper.cpp/resolve/main/ggml-%model%.bin -Destination \"%models_path%\\ggml-%model%.bin\"
```
danbev: Is this line missing a closing double quote for the PowerShell command perhaps?
```suggestion
  PowerShell -NoProfile -ExecutionPolicy Bypass -Command "Start-BitsTransfer -Source https://huggingface.co/akashmjn/tinydiarize-whisper.cpp/resolve/main/ggml-%model%.bin -Destination \"%models_path%\\ggml-%model%.bin\""

```

GalFawkes: Oh, yup, it totally is. That's my bad, I missed that (or my IDE led me wrong) 

I don't really do a ton of CMD or PS scripting, so I'm just glad my little hack to pull the tdrz model worked!

PR #3389, opened 2025-08-24
Ruby binding: handle negative value in padding

Treboko:
this might happen depending on the way the $stderr.winsize is defined.
If the expression "$stderr.winsize[1] - line.size" in Line 114 gets negative, we will get a "negative argument" exception in the padding calculation


KitaitiMakoto (approved):
Thanks!

PR #3395, opened 2025-08-28
Add support for --carry-initial-prompt

alubbe:
This PR is bringing over the `--carry-initial-prompt` flag from the python library (https://github.com/openai/whisper/pull/2343)

By default, an `--prompt` (initial prompt) is only used for the first decoding window; subsequent windows rely on the text generated so far for continuity. When you pass `--carry-initial-prompt`, the initial prompt tokens are explicitly prepended to every internal decode window. This mirrors the Python reference implementation's `carry_initial_prompt` behavior and can help enforce custom vocabulary or style throughout long transcriptions. Trade‑off: it may slightly reduce the model's ability to adapt dynamically to newly generated context (can increase risk of repetitions if the prompt is long). If the combined size of the carried initial prompt and the rolling context exceeds half the model text context, the leftmost (oldest) part of the initial prompt is truncated to fit.


KitaitiMakoto (commented):
Patches for Ruby are nice, though I'm not sure the essential changes and API are accepted. Let me point just a thing.

bindings/ruby/ext/ruby_whisper_params.c
```diff
@@ -1303,6 +1325,7 @@ init_ruby_whisper_params(VALUE *mWhisper)
   DEFINE_PARAM(max_len, 11)
   DEFINE_PARAM(split_on_word, 12)
   DEFINE_PARAM(initial_prompt, 13)
+  DEFINE_PARAM(carry_initial_prompt, 36)
```
KitaitiMakoto: Can you assign `14` to `carry_initial_prompt` and re-number params below?

alubbe: sure thing!

KitaitiMakoto: Thanks!

alubbe:
@ggerganov could I ask you to review this PR?

alubbe:
any update on this? Happy to resolve the conflict in `cli.cpp` if I can get a general feeling of whether you're interested in merging this functionality or not. We've been using this branch for weeks now and have observed a much improved transcription from whisper when it comes to unusual names (people, places, companies, etc.) from carrying the initial prompt.

ggerganov (commented):
I think the main complexity comes from using a single `prompt_past` vector in the `whisper_state` which results in some convoluted logic for deduplicating and slicing the tokens.

I expect that the logic can become much simpler if you replace `prompt_past` with 2 vectors: `prompt_past0` and `prompt_past1`. The full prompt is a concatenation of `prompt_past0 + prompt_past1`. The `prompt_past0` can be utilized to store some static prefix - i.e. the original prompt that is being carried.

alubbe:
That's a good point. I tried taking a bit further and simplifying it as much as I could - what do you think?

src/whisper.cpp
```diff
@@ -6874,17 +6882,17 @@ int whisper_full_with_state(
...
+        prompt_past0.clear();
+        prompt_past1.clear();
     }
 
     // prepare prompt
     {
         std::vector<whisper_token> prompt_tokens;
-
-        // initial prompt
         if (!params.prompt_tokens && params.initial_prompt) {
```
ggerganov, line 6901: Keep the comment:

```suggestion
        // tokenize the initial prompt
        if (!params.prompt_tokens && params.initial_prompt) {
```

src/whisper.cpp
```diff
@@ -6896,14 +6904,14 @@ int whisper_full_with_state(
...
         }
-
-        // prepend the prompt tokens to the prompt_past
         if (params.prompt_tokens && params.prompt_n_tokens > 0) {
-            // parse tokens from the pointer
-            for (int i = 0; i < params.prompt_n_tokens; i++) {
-                prompt_past.push_back(params.prompt_tokens[i]);
+            if (params.carry_initial_prompt) {
+                if (prompt_past0.empty()) {
+                    prompt_past0.insert(prompt_past0.end(), params.prompt_tokens, params.prompt_tokens + params.prompt_n_tokens);
```
ggerganov: Should this be simply:

```suggestion
                    prompt_past0 = params.prompt_tokens;
```

src/whisper.cpp
```diff
@@ -6896,14 +6904,14 @@ int whisper_full_with_state(
...
+                }
+            } else {
+                prompt_past1.insert(prompt_past1.end(), params.prompt_tokens, params.prompt_tokens + params.prompt_n_tokens);
```
ggerganov: We still want to use the original `std::rotate` implementation for efficieny.

src/whisper.cpp
```diff
@@ -6960,6 +6968,7 @@ int whisper_full_with_state(
     std::vector<beam_candidate> beam_candidates;
 
     // main loop
+    bool first_history_iter = true; // track first decode iteration for carry_initial_prompt logic
```
ggerganov: No need to add this var - use `seek == seek_start` instead.

src/whisper.cpp
```diff
@@ -7537,14 +7583,17 @@ int whisper_full_with_state(
...
             //WHISPER_LOG_DEBUG("prompt_init.size() = %d, prompt.size() = %d, result_len = %d, seek_delta = %d\n", prompt_init.size(), prompt.size(), result_len, seek_delta);
 
-            // update prompt_past
-            prompt_past.clear();
-            if (prompt.front() == whisper_token_prev(ctx)) {
-                prompt_past.insert(prompt_past.end(), prompt.begin() + 1, prompt.end() - prompt_init.size());
+            if (!params.carry_initial_prompt) {
+                prompt_past1.clear();
+                if (!prompt.empty() && prompt.front() == whisper_token_prev(ctx)) {
+                    auto start_it = prompt.begin() + 1;
```
ggerganov: No need for this intermediate var - keep the original `insert` call

src/whisper.cpp
```diff
@@ -7050,17 +7059,54 @@ int whisper_full_with_state(
...
-                // if we have already generated some text, use it as a prompt to condition the next generation
-                if (!prompt_past.empty() && t_cur < 0.5f && params.n_max_text_ctx > 0) {
-                    int n_take = std::min(std::min(params.n_max_text_ctx, whisper_n_text_ctx(ctx)/2), int(prompt_past.size()));
+                if (params.n_max_text_ctx > 0 &&
+                    t_cur < WHISPER_HISTORY_CONDITIONING_TEMP_CUTOFF) {
+
+                    const bool have_dynamic = !prompt_past1.empty();
+                    const bool can_carry_static = params.carry_initial_prompt && !prompt_past0.empty() && !first_history_iter;
+
+                    if (have_dynamic || can_carry_static) {
```
ggerganov: This outer `if` can be removed.

src/whisper.cpp
```diff
@@ -7537,14 +7582,16 @@ int whisper_full_with_state(
...
-            // update prompt_past
-            prompt_past.clear();
-            if (prompt.front() == whisper_token_prev(ctx)) {
-                prompt_past.insert(prompt_past.end(), prompt.begin() + 1, prompt.end() - prompt_init.size());
+            if (!params.carry_initial_prompt) {
+                prompt_past1.clear();
+                if (!prompt.empty() && prompt.front() == whisper_token_prev(ctx)) {
+                    prompt_past1.insert(prompt_past1.end(), prompt.begin() + 1, prompt.end() - prompt_init.size());
+                }
             }
```
ggerganov: Need to clarify why this update of `prompt_past1` is not needed when `carry_initial_prompt == true` - the reason is not obvious to me.

My expectation was that `prompt_past1` would behave the same way as the old `prompt_past`, regardless of the `carry_initial_prompt` value.

alubbe: Great catch, that was leftover from my earlier experimentation

src/whisper.cpp
```diff
@@ -7050,12 +7063,44 @@ int whisper_full_with_state(
...
+                                }
+                            }
+                        } else {
+                            // Dynamic only path
+                            int n_take = std::min(max_ctx_half, (int) prompt_past1.size());
+                            if (n_take > 0) {
+                                prompt.insert(prompt.end(), prompt_past1.end() - n_take, prompt_past1.end());
+                            }
+                        }
+                    }
```
ggerganov: I think this part can become simpler:

```c++
if (params.n_max_text_ctx > 0 && t_cur < WHISPER_HISTORY_CONDITIONING_TEMP_CUTOFF) {
    const bool can_take0 = params.carry_initial_prompt && !prompt_past0.empty() && seek != seek_start;
    const bool can_take1 = !prompt_past1.empty();

    int max_ctx_half = std::min(params.n_max_text_ctx, whisper_n_text_ctx(ctx) / 2);
    if (max_ctx_half > 0 && (can_take0 || can_take1)) {
        prompt.push_back(whisper_token_prev(ctx));

        int n_take0 = 0;
        if (can_take0) {
            n_take0 = std::min<int>(max_ctx_half - 1, prompt_past0.size());
            if (n_take0 > 0) {
                auto start0 = n_take0 < (int)prompt_past0.size() ? prompt_past0.end() - n_take0 : prompt_past0.begin();
                prompt.insert(prompt.end(), start0, prompt_past0.end());
            }
        }

        int n_take1 = std::min<int>(max_ctx_half - n_take0 - 1, prompt_past1.size());
        if (n_take1 > 0) {
            prompt.insert(prompt.end(), prompt_past1.end() - n_take1, prompt_past1.end());
        }
    }
}
```

src/whisper.cpp
```diff
@@ -6896,14 +6906,17 @@ int whisper_full_with_state(
...
-                prompt_past.push_back(params.prompt_tokens[i]);
+            if (params.carry_initial_prompt) {
+                if (prompt_past0.empty()) {
+                    prompt_past0.assign(params.prompt_tokens, params.prompt_tokens + params.prompt_n_tokens);
+                }
+            } else {
+                for (int i = 0; i < params.prompt_n_tokens; ++i) {
+                    prompt_past1.push_back(params.prompt_tokens[i]);
+                }
+                std::rotate(prompt_past1.begin(), prompt_past1.end() - params.prompt_n_tokens, prompt_past1.end());
```
ggerganov, line 6929: If I am following this logic correctly, I think there is a problem in the first iteration:

- When `carry_initia_prompt == true`, here we have `prompt_past0` non-empty and `prompt_past1` empty
- But later on line 7067, we check `seek != seek_start` which would be `false` (this is the first iter)
- So as a result I think the first iter will run without any prompt

If I am correct, I think the fix is to remove the `seek != seek_start` check.

alubbe: good catch - fixed!

alubbe:
I did another check of all the logic in here and I think I found another issue - we only want to run line 7569 outside the `carry_initial_prompt` mode because prompt_past1 contains part of prompt_past0, so we're duplicating content

src/whisper.cpp
```diff
@@ -6989,7 +7002,7 @@ int whisper_full_with_state(
         // if there is a very short audio segment left to process, we remove any past prompt since it tends
         // to confuse the decoder and often make it repeat or hallucinate stuff
         if (seek > seek_start && seek + 500 >= seek_end) {
-            prompt_past.clear();
+            prompt_past1.clear();
```
ggerganov, line 7017: For this case, we should clear also the static prompt because it will probably result in the same hallucinations described in the comment:

```suggestion
            prompt_past0.clear();
            prompt_past1.clear();
```

src/whisper.cpp
```diff
@@ -7537,14 +7564,17 @@ int whisper_full_with_state(
...
             //WHISPER_LOG_DEBUG("prompt_init.size() = %d, prompt.size() = %d, result_len = %d, seek_delta = %d\n", prompt_init.size(), prompt.size(), result_len, seek_delta);
 
-            // update prompt_past
-            prompt_past.clear();
-            if (prompt.front() == whisper_token_prev(ctx)) {
-                prompt_past.insert(prompt_past.end(), prompt.begin() + 1, prompt.end() - prompt_init.size());
+            // update prompt_past1
+            prompt_past1.clear();
+            if (!params.carry_initial_prompt && !prompt.empty() && prompt.front() == whisper_token_prev(ctx)) {
+                prompt_past1.insert(prompt_past1.end(), prompt.begin() + 1, prompt.end() - prompt_init.size());
```
ggerganov, line 7581: An alternative would be to preserve the original behaviour by using the fact that `prompt.size() > prompt_past0.size()`:

```suggestion
            // update prompt_past1
            prompt_past1.clear();
            if (!prompt.empty() && prompt.front() == whisper_token_prev(ctx)) {
                prompt_past1.insert(prompt_past1.end(), prompt.begin() + 1 + prompt_past0.size(), prompt.end() - prompt_init.size());
```

Though make sure to double-check this works as expected.

alubbe: I think that this assumes the entire `prompt_past0` is always included in the prompt, but that's not guaranteed. For example, if `max_ctx_half - 1 < prompt_past0.size()`, we only take a tail of `prompt_past0`, not all of it.

ggerganov: Yes, good point. Should we truncate `prompt_past0` upon initialization so that it does not exceed the `max_ctx_half`?

src/whisper.cpp
```diff
@@ -7050,12 +7064,26 @@ int whisper_full_with_state(
...
+                        if (can_take0) {
+                            n_take0 = std::min<int>(max_ctx_half - 1, prompt_past0.size());
+                            prompt.insert(prompt.end(), prompt_past0.end() - n_take0, prompt_past0.end());
+                        }
 
-                    prompt = { whisper_token_prev(ctx) };
-                    prompt.insert(prompt.begin() + 1, prompt_past.end() - n_take, prompt_past.end());
+                        // Fill remaining budget with dynamic tokens (rolling context)
+                        const int n_take1 = std::min<int>(max_ctx_half - n_take0 - 1, prompt_past1.size());
+                        prompt.insert(prompt.end(), prompt_past1.end() - n_take1, prompt_past1.end());
```
ggerganov, line 7095: When `params.n_max_text_ctx == 0`, the `n_take1` will become `-1`. Probably we have to clamp it to `0` to avoid iterating past the end of the vector. Same for `n_take0`.

Edit: nvm - we have a check for `params.n_max_text_ctx > 0` earlier.

src/whisper.cpp
alubbe: done - and added a warning message when we truncate

ggerganov (approved):
We can merge after the CI is green.

alubbe:
@tannisroot coming with https://github.com/ggml-org/whisper.cpp/pull/3781

PR #3419, opened 2025-09-13
refactor(go bindings): thread safe contexts over whisper state

ciricc:
### Problem
- Context was params-centric and shared global state, leading to racy access and "concurrent map writes".

### Solution
- Refactor Context to be state-backed. Each Context now owns its own whisper_state for safe concurrent processing.
- Introduce stateful and stateless Context flavors.
- Add a high-level Parameters wrapper with a unified interface.

### Key changes
- Context isolation: per-instance whisper_state ownership.
- Parameters abstraction: new top-level Parameters type.
- Resource management: Close methods added and state validity checks after Close.
- Type safety: wrap `whisperCtx` and `whisperState` pointers with lifecycle management.
- Concurrency: fix `sync.Map` usage to be safe for parallel processing (since 59b3c4723cb10ad10c7d8a8db59e076d29bbe117).
- Tooling and tests: new benchmarks, test helpers, and expanded test coverage.
- Build and module updates: Makefile and go.mod adjustments.

### Compatibility
- Backwards compatible: existing methods continue to work, with deprecation warnings where applicable.
- Concurrent safe: removes "concurrent map writes" under parallel workloads.

### Migration notes
- Prefer the new Parameters wrapper and the new stateful/stateless context constructors.
- Always call Close on Context and related resources when finished.


ciricc:
PR synced with upstream with new parameters. Benchmark command fixed, deleted diarization tests (required samples, diarization model ~500mb file downloaded).

PR #3423, opened 2025-09-17
Fixed VAD to work when using whisper_full_with_state

mdestagnol:
Currently `whisper_full_with_state` doesn't work with VAD.

The main inference method is `whisper_full_with_state`. This method takes a `whisper_context` and a `whisper_state`. On top of this method are built 2 convenience methods, `whisper_full` and `whisper_full_parallel`.

The VAD logic is currently only implemented in those convenience methods, but not in `whisper_full_with_state` itself. This is an issue whenever we need to deal with more than one state (when processing multiple real time streams for example).

This pull request moves the logic into `whisper_full_with_state` and fix a few incorrect calls to `ctx->state` instead of `state` directly.

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -6802,23 +6802,12 @@ static bool whisper_vad(
 }
 
 int whisper_full_with_state(
-        struct whisper_context * ctx,
-          struct whisper_state * state,
-    struct whisper_full_params   params,
-                   const float * samples,
-                           int   n_samples) {
-    // clear old results
-    auto & result_all = state->result_all;
-
-    result_all.clear();
-
-    if (n_samples > 0) {
-        // compute log mel spectrogram
-        if (whisper_pcm_to_mel_with_state(ctx, state, samples, n_samples, params.n_threads) != 0) {
-            WHISPER_LOG_ERROR("%s: failed to compute log mel spectrogram\n", __func__);
-            return -2;
-        }
-    }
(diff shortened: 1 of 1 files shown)


src/whisper.cpp
```diff
@@ -6820,6 +6820,24 @@ int whisper_full_with_state(
         }
     }
 
+    std::vector<float> vad_samples;
```
danbev, line 6812: I think this needs to come before computation of the log mel spectrogram (`whisper_pcm_to_mel_with_state`) for it to work.

danbev:
@mdestagnol Thanks for the pull request. 

I've been thinking about this and if I recall correctly, the original [pr](https://github.com/ggml-org/whisper.cpp/pull/3065/commits/871da0bf456910f2c6facf3fa9c786ba82bb253e#diff-52d0dad82debc5d9cb716235a013f3f8eaf4c98ec318c8535c1e631486ddb4e1R6686) for VAD actually did something like this. I recall getting feedback about the state should be passed in and not be updated inside of this function (I've not been able to find the comment though I'm afraid).

So I think exposing `whisper_vad` might be a better [option](https://github.com/danbev/whisper.cpp/commit/515e5e9505ff74b8d801aac57e5ab8c6b12955d1) here.

tazz4843:
Whatever route is taken, this should be documented anyways since it was extremely confusing wondering why VAD wasn't working despite having the switch on.

PR #3466, opened 2025-10-09
Add flag for Vulkan in README.md

toboil-features:
To encounter https://github.com/ggml-org/whisper.cpp/issues/3280 issue for others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,7 +336,7 @@ First, make sure your graphics card driver provides support for Vulkan API.
 
 Now build `whisper.cpp` with Vulkan support:
 ```
-cmake -B build -DGGML_VULKAN=1
+cmake -B build -DGGML_VULKAN=1 -DWHISPER_SDL2=ON
 cmake --build build -j --config Release
 ```
 


README.md
```diff
@@ -336,7 +336,7 @@ First, make sure your graphics card driver provides support for Vulkan API.
 
 Now build `whisper.cpp` with Vulkan support:
 ```
-cmake -B build -DGGML_VULKAN=1
+cmake -B build -DGGML_VULKAN=1 -DWHISPER_SDL2=ON
```
danbev, line 339: I'm not sure this should be specified here as this is not a requirement to build whisper.cpp with support for the Vulkan backend. This is only required for some examples like the [stream](https://github.com/ggml-org/whisper.cpp/tree/master/examples/stream#building) and it does specify that this is required there.

toboil-features: Probably, better to be noted in "Additionally" section for Vulkan then?

danbev: I was thinking that since this is already mentioned in the samples that it would not be needed here.  Also this is not specific to Vulkan but to the examples that require SDL2.

PR #3472, opened 2025-10-12
Update ruby bindings

alubbe:
I asked Claude Sonnet 4.5 to update the bindings. I don't use these bindings myself, so I can't guarantee that this works, but maybe it's still helpful. Happy to make any changes you'd like.


alubbe:
I was just trying to be helpful, but let's close it if there's no interest

PR #3495, opened 2025-11-01
Critical bug: Fix --no-timestamps flag behavior

OrelSokolov:
# Fix --no-timestamps flag behavior

## Summary

This PR fixes the `--no-timestamps` flag to only affect output formatting without changing transcription quality. Previously, the flag would alter the decoding process, resulting in different and lower quality transcription text.

## Problem

When using `--no-timestamps` flag:
- ❌ Transcription text differed from the same audio without the flag
- ❌ Lower transcription quality
- ❌ Model would sometimes loop/repeat phrases infinitely
- ❌ Flag modified the decoding process by:
  - Adding `<|notimestamps|>` token to the prompt
  - Suppressing all timestamp tokens during decoding

## Solution

The fix ensures `--no-timestamps` only controls output formatting:
- ✅ Model always uses timestamp logic during decoding (for better quality)
- ✅ Transcription text is identical regardless of the flag
- ✅ Added repetition detection to prevent infinite loops
- ✅ Improved segment handling to prevent early termination

### Core Fixes (`src/whisper.cpp`)
1. **Removed `<|notimestamps|>` token injection** - The model no longer adds this token to prompts, allowing proper timestamp-based segmentation
2. **Removed timestamp token suppression** - Timestamp tokens are no longer suppressed from logits, enabling the model to segment properly
3. **Added repetition detection** - Detects and prevents infinite loops where the model repeats the same phrase
4. **Improved error handling** - Better buffer allocation error messages

### Tests (`tests/`)
- Added `test-no-timestamps.cpp` - Automated test that verifies transcription quality is identical with/without the flag
- Added `TEST_NO_TIMESTAMPS.md` - Test documentation
- Updated `tests/CMakeLists.txt` - Test integration

### Documentation
- Added `NO_TIMESTAMPS_FIX.md` - Detailed explanation of the problem and solution

### Automated Test
```bash
cd build
ctest -R test-no-timestamps -V
```

Test verifies that:
- Transcription with timestamps enabled produces text: "And so my fellow Americans..."
- Transcription with `--no-timestamps` produces **identical** text
- ✅ Test passes (9.87s)

### Manual Testing
```bash
# Both commands now produce identical transcription quality:
./whisper-cli -m model.bin -f audio.wav                    # With timestamps in output
./whisper-cli -m model.bin -f audio.wav --no-timestamps    # Without timestamps in output
```

## Backward Compatibility

✅ **Fully backward compatible**
- All existing tests pass
- CLI interface unchanged
- API unchanged
- Only improvement: better transcription quality with `--no-timestamps`

## Impact

- **Users**: Better transcription quality when using `--no-timestamps`
- **Developers**: Clear separation between output formatting and decoding logic
- **Maintenance**: Automated test prevents regression

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Comments added for complex logic
- [x] Documentation updated
- [x] Tests added and passing
- [x] No new warnings introduced
- [x] Backward compatible

### DIFF
--- a/NO_TIMESTAMPS_FIX.md
+++ b/NO_TIMESTAMPS_FIX.md
@@ -1,0 +1,99 @@
+# Fix: --no-timestamps Flag Behavior
+
+## Problem
+
+The `--no-timestamps` flag was incorrectly changing the transcription quality. With this flag enabled, the transcription text would differ from the same audio transcribed without the flag.
+
+### Root Cause
+
+When `no_timestamps = true`, the code would:
+1. Add `<|notimestamps|>` token to the prompt (lines 6933-6935)
+2. Suppress all timestamp tokens in logits (lines 6168-6172)
+
+This fundamentally changed the model's decoding process, resulting in lower transcription quality.
+
+## Solution
+
+Modified the `--no-timestamps` flag to only affect **output formatting**, not the decoding process.
+
+### Changes
+
+**File: `src/whisper.cpp`**
+
+- Lines 6933-6938: Commented out code that adds `<|notimestamps|>` token
+- Lines 6168-6175: Commented out code that suppresses timestamp tokens
+
+The model now always uses timestamp logic during decoding for better quality, regardless of the flag setting.
+
+## Results
+
+### Before Fix
+- ❌ Different transcription text with/without flag
+- ❌ Lower quality with `--no-timestamps`
+- ❌ Model operated in different modes
+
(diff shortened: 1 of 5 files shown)


Sroose:
I can confirm the same erratic behavior with the -nt flag. It frequently omits entire chunks of text compared to a standard pass that includes timestamps.

PR #3503, opened 2025-11-07
Allow NextSegment() to be called across multiple Process() calls in the Go binding

joshmux:
## Summary
Fixes an issue in the Go bindings where calling `NextSegment()` after multiple `Process()` invocations returns no new segments. This prevented incremental or streaming-style processing from functioning correctly.

## Details
When the Go wrapper calls `Process()` more than once, the internal context segment counter (`context.n`) is not reset or updated properly. As a result:

- `NextSegment()` stops yielding new segments after the first `Process()` call.
- Incremental processing workflows (e.g., feeding audio in chunks) silently fail.
- The behavior diverges from expectations set by the underlying C++ implementation.

This PR resets `context.n` at the start of each `Process()` call, ensuring that segment iteration works correctly across multiple processing passes.

## Impact
- Enables proper streaming and incremental recognition in the Go bindings.  
- Prevents confusing silent failures where no segments are returned.  
- Improves consistency and correctness relative to `whisper.cpp` core behavior.  

## Backward Compatibility
This change is internal to the Go bindings and does **not** modify the public API.  
Existing code using a single `Process()` → `NextSegment()` workflow will behave exactly as before.


joshmux:
I think my fork being a few weeks out of date might be causing this issue. Just merged with `upstream/master`

PR #3518, opened 2025-11-11
ruby : VAD separately from ASR

KitaitiMakoto:
Hi,

I found that we can use VAD feature separately from ASR and added API for that to Ruby bindings.

Thank you for such useful feature and API.


ggerganov (approved):

KitaitiMakoto:
> Did you manage to run the VAD on a GPU or is the VAD CPU-only?

I didn't care about. It seems CPU is used:
https://github.com/ggml-org/whisper.cpp/blob/a1867e0dad0b21b35afa43fc815dae60c9a139d6/src/whisper.cpp#L4658-L4660

PR #3557, opened 2025-12-04
examples: add -g/--gpu-device option and GPU_DEVICE env support

brauliobo:
- Added `-g` / `--gpu-device` CLI option to select a specific GPU device.
- Added support for `GPU_DEVICE` environment variable as a fallback.
- Updated `whisper_params` to store the `gpu_device` index.
- Passed the selected device ID to `whisper_context_params` initialization.
- Applied changes to both `examples/cli/cli.cpp` and `examples/server/server.cpp`.

### DIFF
--- a/examples/cli/cli.cpp
+++ b/examples/cli/cli.cpp
@@ -77,6 +77,7 @@ struct whisper_params {
     bool log_score       = false;
     bool use_gpu         = true;
     bool flash_attn      = true;
+    int32_t gpu_device   = 0;
     bool suppress_nst    = false;
     bool carry_initial_prompt = false;
 
@@ -129,6 +130,10 @@ static char * requires_value_error(const std::string & arg) {
 }
 
 static bool whisper_params_parse(int argc, char ** argv, whisper_params & params) {
+    if (const char * env_device = std::getenv("WHISPER_ARG_DEVICE")) {
+        params.gpu_device = std::stoi(env_device);
+    }
+
     for (int i = 1; i < argc; i++) {
         std::string arg = argv[i];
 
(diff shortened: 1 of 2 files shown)


examples/cli/cli.cpp
```diff
@@ -195,6 +200,7 @@ static bool whisper_params_parse(int argc, char ** argv, whisper_params & params
         else if (arg == "-dtw"  || arg == "--dtw")                  { params.dtw             = ARGV_NEXT; }
         else if (arg == "-ls"   || arg == "--log-score")            { params.log_score       = true; }
         else if (arg == "-ng"   || arg == "--no-gpu")               { params.use_gpu         = false; }
+        else if (arg == "-g"    || arg == "--gpu-device")           { params.gpu_device      = std::stoi(ARGV_NEXT); }
```
danbev: Perhaps we should use `--device` here to be consistent with [llama.cpp](https://github.com/ggml-org/llama.cpp/blob/933414c0b6f21af269bdb4fa2fa1b257b9c0fc53/common/arg.cpp#L1875).

examples/server/server.cpp
```diff
@@ -196,6 +198,10 @@ void whisper_print_usage(int /*argc*/, char ** argv, const whisper_params & para
 }
 
 bool whisper_params_parse(int argc, char ** argv, whisper_params & params, server_params & sparams) {
+    if (const char * env_gpu_device = std::getenv("GPU_DEVICE")) {
```
danbev: And perhaps this should be `WHISPER_ARG_DEVICE` to be consistent with [llama.cpp](https://github.com/ggml-org/llama.cpp/blob/933414c0b6f21af269bdb4fa2fa1b257b9c0fc53/common/arg.cpp#L1881).

brauliobo:
sorry for the @danbev, just updated the PR with the requested changes

PR #3593, opened 2026-01-05
add detected language callback

seven1240:


ggerganov:
Use `whisper_full_params.detect_language = true`

PR #3600, opened 2026-01-10
cmake: fix executable example targets

ink-splatters:
Added `whisper-` prefix and `install(TARGETS ${TARGET} RUNTIME)` for executable targets from examples, where it was missing.

### DIFF
--- a/examples/quantize/CMakeLists.txt
+++ b/examples/quantize/CMakeLists.txt
@@ -1,6 +1,7 @@
-set(TARGET quantize)
+set(TARGET whisper-quantize)
 add_executable(${TARGET} quantize.cpp)
 
 include(DefaultTargetOptions)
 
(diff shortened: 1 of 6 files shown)


danbev (approved):

KitaitiMakoto:
For Ruby bindings' test failure, I will investigate after merge.

examples/quantize/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
-set(TARGET quantize)
+set(TARGET whisper-quantize)
```
danbev, line 1: Could you also update [build.yml](https://github.com/ggml-org/whisper.cpp/blob/679bdb53dbcbfb3e42685f50c7ff367949fd4d48/.github/workflows/build.yml#L1255) to use `whisper-quantize`?

danbev (approved):

h9535114-design:
h9535114@gmail.com

Issue #3571, opened 2025-12-17
Feature Request: Layer-wise (Mixed Precision) Quantization Configuration


lhpqaq:
### Summary
It would be highly valuable if whisper.cpp supported mixed precision and layer-wise quantization configuration, allowing users to specify different quantization types for different layers or tensors when converting models. This would enable more flexible and optimized deployments, balancing memory, accuracy, and performance based on user and hardware requirements. 

### Motivation
- Real-world edge deployments and research increasingly require fine-grained quantization strategies to optimize for specific accuracy, performance, and size constraints.
- Currently, whisper. cpp supports applying only one quantization type per model conversion. There is no straightforward way to assign different quantization types (e.g., Q8_0 for encoder, Q4_0 for decoder, FP16 for attention layers, etc.) to specific layers or tensor name patterns.
- llama.cpp already provides a `tensor_types`-like configuration for flexible per-layer quantization control.  Bringing similar options to whisper.cpp would enable hardware-aware and use-case-tailored deployment.

### Proposed Solution
- Provide a mechanism (such as a config file or command-line flag) for users to specify desired quantization types for each layer, group of layers, or by regex/pattern match on tensor names. 
- Default to the current unified quantization type when no mapping is provided, ensuring backward compatibility. 

Thank you for considering this improvement to whisper.cpp! 

**Do you think this is feasible? If so, I can try to implement it myself. Do you have any suggestions or advice?**

PR #3605, opened 2026-01-12
Support Mixed-Precision Quantization

lhpqaq:
This PR adds support for **mixed-precision quantization**, allowing different tensors to be quantized with different formats via regular-expression-based configuration.

### Usage Example:
```bash
./build/bin/quantize \
  --tensor-type 'decoder\.blocks\.\d+\.(mlp_ln\..*|mlp\.0\..*)'=q4_0 \
  --tensor-type 'decoder\.blocks\.\d+\.mlp\.2\..*'=q4_0 \
  ./models/ggml-small.bin \
  ./models/ggml-small-mixed.bin \
  q8_0
```

The base quantization type (`q8_0` in this case) is applied to all tensors not matched by any `--tensor-type` rule.

### Evaluation of ggml-small on LibriSpeech (WER %):

| Configuration                     | Size  | WER (%)     |
|----------------------------------|-------|-------------|
| Full INT8                        | 253M  | 3.200       |
| Full INT4                        | 139M  | 3.283       |
| Decoder Self-Attention MLP → Q4, rest Q8 | 239M  | 3.191   |
| Encoder → Q4, Decoder → Q8       | 212M  | 3.25       |
| Encoder → Q8, Decoder → Q4       | 180M  | 3.21       |

issue: #3571

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -1684,6 +1684,71 @@ static bool whisper_model_load(struct whisper_model_loader * loader, whisper_con
 
     const size_t n_tensors = 10 /* input */ + 15 + 15*n_audio_layer + 24*n_text_layer;
 
+    // Pre-scan tensor metadata from file to determine actual types
+    // This allows us to allocate device memory with the correct sizes
+    struct tensor_meta {
+        ggml_type type;
+        int32_t ne[4];
+    };
+    std::map<std::string, tensor_meta> tensor_type_map;
+    size_t tensor_start_offset = 0; // file offset where tensor section begins
+
+    // If loader supports skip, seek, and tell, scan tensor metadata first (without loading data)
+    if (loader->skip && loader->seek && loader->tell) {
+        // Remember the current position (start of tensors section)
+        tensor_start_offset = loader->tell(loader->context);
+
+        while (true) {
+            int32_t n_dims;
+
+            read_safe(loader, n_dims);
+
+            // Check for EOF after reading the first field
+            if (loader->eof(loader->context)) {
+                break;
+            }
+
+            int32_t length;
+            int32_t ttype;
+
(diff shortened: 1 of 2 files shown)


src/whisper.cpp
```diff
@@ -1911,28 +1911,61 @@ static bool whisper_model_load(struct whisper_model_loader * loader, whisper_con
...
+                for (int i = 2; i < GGML_MAX_DIMS; i++) {
+                    tensor->nb[i] = tensor->nb[i-1] * tensor->ne[i-1];
+                }
+            } else {
+                // Normal case: types match, verify size
+                if (file_tensor_size != expected_tensor_size) {
+                    WHISPER_LOG_ERROR("%s: tensor '%s' has wrong size in model file: got %zu, expected %zu\n",
+                            __func__, name.data(), expected_tensor_size, file_tensor_size);
+                    return false;
+                }
```
ggerganov, line 2041: The way this is implemented, the allocated memory on the device would still correspond to the original tensor type declared earlier in this function.

Need to figure out a better way to do it.

lhpqaq: > The way this is implemented, the allocated memory on the device would still correspond to the original tensor type declared earlier in this function.
> 
> Need to figure out a better way to do it.

@ggerganov Pre-scan tensor metadata from the file to determine actual data types  
Implement skip/seek mechanisms to avoid loading tensor data unnecessarily.  
Empirical results show that this optimization does not slow down model loading.  
Do you think this approach is appropriate?

### ggml-large-v2 Model Load Time

| Run | Before (ms) | After (ms) |
|----:|------------:|-----------:|
| 1 | 1372.23 | 1385.15 |
| 2 | 1417.75 | 1349.69 |
| 3 | 1377.65 | 1352.80 |

PR #3608, opened 2026-01-13
[AMD] Add VitisAI Plugin for AMD Ryzen AI NPU encoder offload

schnkmwt:
This PR adds the VitisAI Plugin for AMD Ryzen AI NPUs to offload Whisper encoder module.
1. Adds a new build option `WHISPER_VITISAI` to build with the new plugin
2. The plugin adds a redirecting backend for the whisper encoder to run on AMD Strix NPU using a pre-optimized model binary available from Huggingface and returns the outputs back to run on decoder backend.
    1. The backend also supports offloading cross projection linear layers onto NPU if the pre-optimized model binary supports it. Detection is built-in based on the binary.
3. Requires proprietary packages from AMD

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -48,6 +48,10 @@ if (WHISPER_OPENVINO)
     find_package(OpenVINO REQUIRED COMPONENTS Runtime)
 endif()
 
+if (WHISPER_VITISAI)
+    find_package(FlexmlRT REQUIRED)
+endif()
+
 #
 # libraries
 #
@@ -101,6 +105,30 @@ if (WHISPER_OPENVINO)
     set_target_properties(${TARGET} PROPERTIES FOLDER "libs")
 endif()
 
+if (WHISPER_VITISAI)
+    set(TARGET whisper.vitisai)
+
+    add_library(${TARGET} OBJECT
+        vitisai/whisper-vitisai-encoder.h
+        vitisai/whisper-vitisai-encoder.cpp
+        )
+
+    target_include_directories(${TARGET} PUBLIC
+        .
+        )
+
+    set_property(TARGET ${TARGET} PROPERTY POSITION_INDEPENDENT_CODE ON)
+    set(WHISPER_EXTRA_FLAGS ${WHISPER_EXTRA_FLAGS} -DWHISPER_USE_VITISAI)
+
+    # Add C++17 standard for MSVC
+    if (MSVC)
+        target_compile_options(${TARGET} PRIVATE /std:c++17)
+    endif()
+
+    target_link_libraries(${TARGET} PRIVATE ggml flexmlrt::flexmlrt)
+    set_target_properties(${TARGET} PROPERTIES FOLDER "libs")
+endif()
+
 # whisper
 
 add_library(whisper
@@ -135,6 +163,10 @@ endif()
 
 if (WHISPER_OPENVINO)
     target_link_libraries(whisper PRIVATE whisper.openvino)
+endif()
+
+if (WHISPER_VITISAI)
+    target_link_libraries(whisper PRIVATE whisper.vitisai)
 endif()
 
 if (WHISPER_MKL)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -173,10 +173,9 @@ if (WHISPER_VITISAI)
     set_property(TARGET ${TARGET} PROPERTY POSITION_INDEPENDENT_CODE ON)
     set(WHISPER_EXTRA_FLAGS ${WHISPER_EXTRA_FLAGS} -DWHISPER_USE_VITISAI)
 
-    # Add C++17 standard for MSVC
-    if (MSVC)
-        target_compile_options(${TARGET} PRIVATE /std:c++17)
-    endif()
+    # FlexMLRT headers and this plugin require C++17. Keep it PRIVATE so the
+    # C++11 requirement of the whisper target is not bumped.
+    target_compile_features(${TARGET} PRIVATE cxx_std_17)
 
     target_compile_definitions(${TARGET} PRIVATE
         WHISPER_FLEXMLRT_LEGACY_RAI_OVERRIDES=${WHISPER_FLEXMLRT_LEGACY_RAI_OVERRIDES}
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -160,6 +160,8 @@ if (WHISPER_VITISAI)
     set(TARGET whisper.vitisai)
 
     add_library(${TARGET} OBJECT
+        vitisai/whisper-vitisai-helpers.h
+        vitisai/whisper-vitisai-helpers.cpp
         vitisai/whisper-vitisai-encoder.h
         vitisai/whisper-vitisai-encoder.cpp
         )
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -166,9 +166,7 @@ if (WHISPER_VITISAI)
         vitisai/whisper-vitisai-encoder.cpp
         )
 
-    target_include_directories(${TARGET} PUBLIC
-        .
-        )
+    target_include_directories(${TARGET} PUBLIC .)
 
     set_property(TARGET ${TARGET} PROPERTY POSITION_INDEPENDENT_CODE ON)
     set(WHISPER_EXTRA_FLAGS ${WHISPER_EXTRA_FLAGS} -DWHISPER_USE_VITISAI)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -48,6 +48,10 @@ if (WHISPER_OPENVINO)
     find_package(OpenVINO REQUIRED COMPONENTS Runtime)
 endif()
 
+if (WHISPER_VITISAI)
+    find_package(FlexmlRT REQUIRED)
+endif()
+
 #
 # libraries
 #
@@ -98,6 +102,30 @@ if (WHISPER_OPENVINO)
     set(WHISPER_EXTRA_FLAGS ${WHISPER_EXTRA_FLAGS} -DWHISPER_USE_OPENVINO)
 
     target_link_libraries(${TARGET} PRIVATE ggml openvino::runtime)
+    set_target_properties(${TARGET} PROPERTIES FOLDER "libs")
+endif()
+
+if (WHISPER_VITISAI)
+    set(TARGET whisper.vitisai)
+
+    add_library(${TARGET} OBJECT
+        vitisai/whisper-vitisai-encoder.h
+        vitisai/whisper-vitisai-encoder.cpp
+        )
+
+    target_include_directories(${TARGET} PUBLIC
+        .
+        )
+
+    set_property(TARGET ${TARGET} PROPERTY POSITION_INDEPENDENT_CODE ON)
+    set(WHISPER_EXTRA_FLAGS ${WHISPER_EXTRA_FLAGS} -DWHISPER_USE_VITISAI)
+
+    # Add C++17 standard for MSVC
+    if (MSVC)
+        target_compile_options(${TARGET} PRIVATE /std:c++17)
+    endif()
+
+    target_link_libraries(${TARGET} PRIVATE ggml flexmlrt::flexmlrt)
     set_target_properties(${TARGET} PROPERTIES FOLDER "libs")
 endif()
 
@@ -155,6 +183,10 @@ endif()
 
 if (WHISPER_OPENVINO)
     target_link_libraries(whisper PRIVATE whisper.openvino)
+endif()
+
+if (WHISPER_VITISAI)
+    target_link_libraries(whisper PRIVATE whisper.vitisai)
 endif()
 
 if (WHISPER_MKL)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -50,6 +50,57 @@ endif()
 
 if (WHISPER_VITISAI)
     find_package(FlexmlRT REQUIRED)
+
+    # Legacy RAI overrides are required by FlexMLRT older than 1.8.0
+    set(WHISPER_FLEXMLRT_LEGACY_RAI_OVERRIDES_MODE "AUTO" CACHE STRING
+        "Legacy RAI override mode for FlexMLRT (AUTO|ON|OFF)")
+    set_property(CACHE WHISPER_FLEXMLRT_LEGACY_RAI_OVERRIDES_MODE PROPERTY STRINGS AUTO ON OFF)
+
+    string(TOUPPER "${WHISPER_FLEXMLRT_LEGACY_RAI_OVERRIDES_MODE}" _flexmlrt_legacy_mode)
+    set(_flexmlrt_legacy_hint "Set -DWHISPER_FLEXMLRT_LEGACY_RAI_OVERRIDES_MODE=ON or OFF explicitly.")
+
+    if (NOT _flexmlrt_legacy_mode MATCHES "^(AUTO|ON|OFF)$")
+        message(FATAL_ERROR
+            "Invalid WHISPER_FLEXMLRT_LEGACY_RAI_OVERRIDES_MODE='${WHISPER_FLEXMLRT_LEGACY_RAI_OVERRIDES_MODE}'. "
+            "Expected AUTO, ON, or OFF.")
+    endif()
+
+    if (_flexmlrt_legacy_mode STREQUAL "AUTO")
+        if (NOT FlexmlRT_DIR)
+            message(FATAL_ERROR
+                "FlexmlRT_DIR is unset after find_package(FlexmlRT). ${_flexmlrt_legacy_hint}")
+        endif()
+
+        # FlexmlRT_DIR points to <pkg_root>/share/cmake/FlexmlRT.
+        get_filename_component(_flexmlrt_init_py "${FlexmlRT_DIR}/../../../__init__.py" ABSOLUTE)
+        if (NOT EXISTS "${_flexmlrt_init_py}")
+            message(FATAL_ERROR
+                "flexmlrt __init__.py not found at ${_flexmlrt_init_py}. ${_flexmlrt_legacy_hint}")
+        endif()
+        set_property(DIRECTORY APPEND PROPERTY CMAKE_CONFIGURE_DEPENDS "${_flexmlrt_init_py}")
+
+        file(STRINGS "${_flexmlrt_init_py}" _flexmlrt_version_lines
+             REGEX "^VERSION[ \t]*=[ \t]*\"[0-9]+\\.[0-9]+\\.[0-9]+")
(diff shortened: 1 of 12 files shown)


src/vitisai/whisper-vitisai-encoder.cpp
```diff
@@ -0,0 +1,204 @@
+// Copyright(C) 2025 Advanced Micro Devices, Inc. All rights reserved.
```
ggerganov: These copyright notices are not needed

src/vitisai/whisper-vitisai-encoder.h
```diff
@@ -0,0 +1,32 @@
...
+
+struct whisper_vitisai_context;
+
+struct whisper_vitisai_context * whisper_vitisai_init(const char * path_model);
+void whisper_vitisai_free(struct whisper_vitisai_context * ctx);
+
+// Function to mmap rai file for Linux and MapViewOfFile for Windows
+bool map_rai_file(const char * path, uint8_t ** buffer, size_t * size);
+// Function to unmap rai file for Linux and UnmapViewOfFile for Windows
+void unmap_rai_file(uint8_t * buffer, size_t size);
```
ggerganov: No need for these to be declared in the header.

src/vitisai/whisper-vitisai-encoder.cpp
```diff
@@ -0,0 +1,204 @@
...
+    output_tensors = model->getIOTensors("output", false);
+
+    // TODO: add assert checks for tensor numbers and shapes
+
+    input_tensors[0].data = mel->data;
+    output_tensors[0].data = out->data;
+
+    try {
+        model->forward(input_tensors, output_tensors);
+        std::fprintf(stdout, "%s: Vitis AI model inference completed.\n", __func__);
```
ggerganov: This looks like debug log - should not be active during normal inference.

ggerganov (commented):
We would need to have a Github Actions workflow that exercises this functionality. If it requires a specific type of hardware, then you should try to provide a self-hosted runner that has the required hardware to run the workflow.

src/vitisai/whisper-vitisai-encoder.cpp
schnkmwt: Removed all copyright notices after approval from AMD legal

src/vitisai/whisper-vitisai-encoder.h
schnkmwt: Cleaned up header

src/vitisai/whisper-vitisai-encoder.cpp
schnkmwt: Added Debug guard to the message print. Please take a look.

schnkmwt:
@ggerganov and @danbev 
Thank you for reviewing. As per you earlier requests. We have added 2 runners, 1 for each Linux and Windows env that sets up, downloads latest models and runs a smoke test to verify the functionality. Can you review and approve the workflow so the CI triggers the test runners?

danbev:
@schnkmwt I noticed that I needed to have python3.12 install which is a dependency of libflexmlrt.so:
```console
ldd /home/danbev/work/ai/flexml/flexmlrt/lib/libflexmlrt.so
	linux-vdso.so.1 (0x00007dbf02806000)
	libpython3.12.so.1.0 => not found
	libxrt_coreutil.so.2 => /opt/xilinx/xrt/lib/libxrt_coreutil.so.2 (0x00007dbf01600000)
	libxrt_core.so.2 => /opt/xilinx/xrt/lib/libxrt_core.so.2 (0x00007dbf0149a000)
	libm.so.6 => /usr/lib/x86_64-linux-gnu/libm.so.6 (0x00007dbf026de000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007dbf01200000)
	libgcc_s.so.1 => /usr/lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007dbf026af000)
	libc.so.6 => /usr/lib/x86_64-linux-gnu/libc.so.6 (0x00007dbf00e00000)
	/lib64/ld-linux-x86-64.so.2 (0x00007dbf02808000)
	libuuid.so.1 => /lib/x86_64-linux-gnu/libuuid.so.1 (0x00007dbf026a5000)
```
I'm using Ubuntu 25.10 which has python 3.13. Perhaps I note about this would be good to add to the documentation, or if I've missed something please let me know. After installing python 3.12 I was able to run without error.

schnkmwt:
> @schnkmwt I noticed that I needed to have python3.12 install which is a dependency of libflexmlrt.so:
> 
> ```
> ldd /home/danbev/work/ai/flexml/flexmlrt/lib/libflexmlrt.so
> 	linux-vdso.so.1 (0x00007dbf02806000)
> 	libpython3.12.so.1.0 => not found
> 	libxrt_coreutil.so.2 => /opt/xilinx/xrt/lib/libxrt_coreutil.so.2 (0x00007dbf01600000)
> 	libxrt_core.so.2 => /opt/xilinx/xrt/lib/libxrt_core.so.2 (0x00007dbf0149a000)
> 	libm.so.6 => /usr/lib/x86_64-linux-gnu/libm.so.6 (0x00007dbf026de000)
> 	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007dbf01200000)
> 	libgcc_s.so.1 => /usr/lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007dbf026af000)
> 	libc.so.6 => /usr/lib/x86_64-linux-gnu/libc.so.6 (0x00007dbf00e00000)
> 	/lib64/ld-linux-x86-64.so.2 (0x00007dbf02808000)
> 	libuuid.so.1 => /lib/x86_64-linux-gnu/libuuid.so.1 (0x00007dbf026a5000)
> ```
> 
> I'm using Ubuntu 25.10 which has python 3.13. Perhaps I note about this would be good to add to the documentation, or if I've missed something please let me know. After installing python 3.12 I was able to run without error.

@danbev Thanks for flagging this. We are targeting python dependency free packaging from next release for Linux. Please note that Windows does not have this dependency. I have added official platform support requirements to the Plugin readme with Python 3.12 as a dependency for Linux. Please review.

src/vitisai/whisper-vitisai-encoder.cpp
```diff
@@ -0,0 +1,603 @@
...
+    // Get the file size
+    struct stat st;
+    if (fstat(fileno(fd), &st) == -1) {
+        fclose(fd);
+        std::fprintf(stderr, "%s: %d: Failed to get file size for rai file '%s'\n", __func__, __LINE__, path);
+        return false;
+    }
+
+    // Mmap the file
+    *buffer = (uint8_t *)mmap(nullptr, st.st_size, PROT_READ, MAP_SHARED, fileno(fd), 0);
```
danbev: Perhaps using MAP_PRIVATE would be more clear that MAP_SHARED in this case. I realize that since the pages are read-only it does not have any real effect (any attempt to write will cause a segfault) but I feel it makes the code clearer.

src/vitisai/whisper-vitisai-encoder.cpp
```diff
@@ -0,0 +1,603 @@
...
+    options.debug = false;
+    options.executeMode = 2;
+    options.extOptions["enable_preemption"] = true;
+
+    // Check if model_path is rai file and if so, add fbs_buffer and fbs_buffer_size to the options
+    if (ctx->model_path.find(".rai") != std::string::npos) {
+        if (map_rai_file(ctx->model_path.c_str(), &ctx->fbs_buffer, &ctx->fbs_buffer_size)) {
+            options.extOptions["fbs_buffer"] = ctx->fbs_buffer;
+            options.extOptions["fbs_buffer_size"] = ctx->fbs_buffer_size;
+            options.extOptions["cache_dir"] = std::string(".");
```
danbev, line 96: I'm not sure what this cache_dir is for, is it for the FlexRT?

Defaulting to the current working directory might not be what we want here, would `~/.cache/whisper-vitisai/` be a better default?
And perhaps this should be configurable using an env var?

src/vitisai/whisper-vitisai-encoder.cpp
```diff
@@ -0,0 +1,603 @@
...
+            return nullptr;
+        }
+    } else {
+        options.deviceName = "stx";
+#if defined(WHISPER_DEBUG)
+        std::fprintf(stderr, "%s: Using default device name 'stx'\n", __func__);
+#endif
+    }
+
+    const bool model_is_rai = ctx->model_path.find(".rai") != std::string::npos;
```
danbev: I think this variable can be moved to before the map_rai_file call as that if statement does the same check improving readability.

src/vitisai/whisper-vitisai-encoder.cpp
```diff
@@ -0,0 +1,603 @@
...
+        return 0;
+    }
+
+    if (ggml_n_dims(out) != 2) {
+        std::fprintf(stderr, "%s: out tensor expected to have 2 dims, got %d\n", __func__, ggml_n_dims(out));
+        return 0;
+    }
+
+    // setup input and output tensors for Vitis AI model
+    std::vector<flexmlrt::client::ErtTensorType> input_tensors, output_tensors;
```
danbev: Please separate this into two lines for consistency with the code base:
```suggestion
    std::vector<flexmlrt::client::ErtTensorType> input_tensors;
    std::vector<flexmlrt::client::ErtTensorType> output_tensors;
```

src/vitisai/whisper-vitisai-encoder.cpp
```diff
@@ -0,0 +1,603 @@
...
+        return false;
+    }
+
+    // Mmap the file
+    *buffer = (uint8_t *)mmap(nullptr, st.st_size, PROT_READ, MAP_SHARED, fileno(fd), 0);
+    if (*buffer == MAP_FAILED) {
+        fclose(fd);
+        std::fprintf(stderr, "%s: %d: Failed to mmap rai file '%s'\n", __func__, __LINE__, path);
+        return false;
+    }
```
danbev: I think we need to close the file descriptor here:
```suggestion
    }
    fclose(fd);
```
Without fclose the file descriptor will still exist in the processes file descriptor table:
```console
gdb) f
#0  map_rai_file (path=0x555555f16160 "models/ggml-base-encoder-vitisai.rai", buffer=0x555555717710, 
    size=0x555555717718) at /home/danbev/work/ai/whisper.cpp/src/vitisai/whisper-vitisai-encoder.cpp:122
122	    *size = st.st_size;
(gdb) thread
[Current thread is 1 (Thread 0x7ffff7e37c40 (LWP 57726))]
(gdb) shell ls -la /proc/57726/fd
total 0
dr-x------ 2 danbev danbev  4 Aug  6 10:40 .
dr-xr-xr-x 9 danbev danbev  0 Aug  6 10:40 ..
lrwx------ 1 danbev danbev 64 Aug  6 10:40 0 -> /dev/pts/0
lrwx------ 1 danbev danbev 64 Aug  6 10:40 1 -> /dev/pts/0
lrwx------ 1 danbev danbev 64 Aug  6 10:40 2 -> /dev/pts/0
lr-x------ 1 danbev danbev 64 Aug  6 10:40 3 -> /home/danbev/work/ai/whisper.cpp/models/ggml-base-encoder-vitisai.rai

```

danbev:
@schnkmwt Could you point me to a `-encoder-cross-vitisai.rai` model I can use? 
I was not able to find one using the download script and I did not find any such models files when looking  around the HF [repository](https://huggingface.co/collections/amd/ryzen-ai-whisper-npu-optimized-onnx-models).

src/whisper.cpp
```diff
@@ -3346,6 +3387,24 @@ static std::string whisper_get_coreml_path_encoder(std::string path_bin) {
...
+    auto pos = path_bin.rfind('.');
+    if (pos != std::string::npos) {
+        path_bin = path_bin.substr(0, pos);
+    }
+
+    const std::string path_vitisai_cross = path_bin + "-encoder-cross-vitisai.rai";
+    if (FILE * file = fopen(path_vitisai_cross.c_str(), "rb")) {
+        fclose(file);
+        return path_vitisai_cross;
+    }
```
danbev: Perhaps we could use `std::filesystem::exists()` here instead:
```suggestion
    if (std::filesystem::exists(path_vitisai_cross)) {
        return path_vitisai_cross;
    }
```
Would required including `<filesystem>`.

src/vitisai/whisper-vitisai-encoder.cpp
```diff
@@ -0,0 +1,603 @@
+#ifdef _WIN32
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+#endif
+
+#include "vitisai/whisper-vitisai-encoder.h"
+#include "FlexMLClient.h"
+#include "ggml.h"
+#include "ggml-backend.h"
```
danbev: I think this include unused and could be removed.

src/vitisai/whisper-vitisai-encoder.cpp
```diff
@@ -0,0 +1,603 @@
...
+
+static void unmap_rai_file(uint8_t * buffer, size_t size) {
+#ifdef _WIN32
+    UnmapViewOfFile(buffer);
+#else
+    munmap(buffer, size);
+#endif // _WIN32
+}
+
+bool whisper_vitisai_file_exists(const char * path) {
```
danbev: I can't find any actual usage of this function and perhaps it could be removed. If not perhaps it could be updated to use std::filesystem or use `whisper-vitsai-helpers.cpp` instead (and make the change there)? 

Perhaps this file could use the functions in helpers and we could remove some of these duplicated functions?

schnkmwt:
@danbev It turns out my refactor of helper function was incomplete pushed. I've now pushed the proper refactor and also addressed your specific comments. Please review again.

src/vitisai/whisper-vitisai-encoder.cpp
schnkmwt: Used suggestion, MAP_PRIVATE updated to MAP_SHARED

src/vitisai/whisper-vitisai-encoder.cpp
schnkmwt: Done as part of cleanup

src/vitisai/whisper-vitisai-encoder.cpp
schnkmwt: The helper refactor removes these lines. Made sure the same formatting in the new sections

src/vitisai/whisper-vitisai-encoder.cpp
schnkmwt: File descriptor closing fixed. Also closed both `hMapping` and `hFile` for Windows path.

src/vitisai/whisper-vitisai-encoder.cpp
schnkmwt: Removed the unused include

src/vitisai/whisper-vitisai-encoder.cpp
schnkmwt: I have pushed the correct helpers refactor and removed `whisper_vitisai_file_exists` function as its not needed.

src/vitisai/whisper-vitisai-encoder.cpp
schnkmwt: This requires some explanation. The `"cache_dir"` here refers to the convention of using "model cache" as an alias for `.rai` flatbuffer files. Hence this option sets the root directory where the plugin looks for the `.rai` flatbuffer file. It is a legacy option but is just part of convention here, that needs to be given a default value. With flatbuffer files, it does not have any real functional impact.

src/whisper.cpp
schnkmwt: The exists function is not needed anymore. We have remove support for the legacy "encoder-cross-vitisai" suffix. Please see my detailed comment below.

schnkmwt:
> @schnkmwt Could you point me to a `-encoder-cross-vitisai.rai` model I can use? I was not able to find one using the download script and I did not find any such models files when looking around the HF [repository](https://huggingface.co/collections/amd/ryzen-ai-whisper-npu-optimized-onnx-models).

@danbev We have deprecated the use of suffix `-encoder-cross-vitisai.rai`. It was there for backwards compatibility. The detection of encoder only vs encoder+cross layers is automated based on the `.rai` graph metadata. Currently we have the following model caches with cross-projection offloaded, work is ongoing to upgrade the rest:
- [ggml-small-encoder-vitisai.rai](https://huggingface.co/amd/whisper-small-onnx-npu/blob/main/ggml-small-encoder-vitisai.rai)
- [ggml-large-v3-turbo-encoder-vitisai.rai](https://huggingface.co/amd/whisper-large-turbo-onnx-npu/blob/main/ggml-large-v3-turbo-encoder-vitisai.rai)

Please try these two model caches to excercise the cross-projection offload paths. The github CI runners have one smoke test for each of encoder only and encoder+cross layers.

danbev:
@schnkmwt Thanks for the updates. I'll take a look at them later this week.

src/vitisai/whisper-vitisai-encoder.h
```diff
@@ -0,0 +1,43 @@
...
+bool whisper_vitisai_has_cross_proj(const struct whisper_vitisai_context * ctx);
+
+struct ggml_tensor;
+
+int whisper_vitisai_encode(
+    struct whisper_vitisai_context * ctx,
+    struct ggml_tensor * mel,
+    struct ggml_tensor * out);
+
+int whisper_vitisai_run_enc_cross(
```
danbev: This does not seem to be used and perhaps could be removed from the header and the implementation?

src/CMakeLists.txt
```diff
@@ -101,6 +156,35 @@ if (WHISPER_OPENVINO)
...
+    add_library(${TARGET} OBJECT
+        vitisai/whisper-vitisai-helpers.h
+        vitisai/whisper-vitisai-helpers.cpp
+        vitisai/whisper-vitisai-encoder.h
+        vitisai/whisper-vitisai-encoder.cpp
+        )
+
+    target_include_directories(${TARGET} PUBLIC
+        .
+        )
```
danbev: Nit:
```suggestion
    target_include_directories(${TARGET} PUBLIC .)
```

src/vitisai/whisper-vitisai-encoder.h
schnkmwt: Removed the stale code

src/CMakeLists.txt
schnkmwt: formatting pushed

schnkmwt:
@danbev Looks like there was an infra failure for vulkan nvidia self-hosted runner. Can you retrigger the CI?

danbev (approved):

schnkmwt:
Thanks @danbev for reviewing and approving the PR. The CI is all green, can we move forward with merge?

danbev:
@schnkmwt Thanks for your (and the others involved in this PR) patience and work on this!
I've merged it now :tada:

PR #3616, opened 2026-01-21
[Feat] eliminate hard-coded record special tokens to be compatible with some models with custom vocab

### LINKED ISSUE #3392: Segment fault on custom model
Hi team,

I tried converting this safetensors model https://huggingface.co/efwkjn/whisper-ja-anime-v0.3 through **convert-h5-to-ggml.py** and the **convert-safetensors-to-ggml.py** script from https://github.com/ggml-org/whisper.cpp/issues/2762, both converted models cause a segment fault error when transcribing audio files by whisper-cli

Jaffe2718:
Fix #3392

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -453,7 +453,7 @@ struct whisper_vocab {
     }
 
     int num_languages() const {
-        return n_vocab - 51765 - (is_multilingual() ? 1 : 0);
+        return token_translate - token_sot - 1;
     }
 };
 
@@ -1587,8 +1587,8 @@ static bool whisper_model_load(struct whisper_model_loader * loader, whisper_con
 
     // load vocab
     {
-        int32_t n_vocab = 0;
-        read_safe(loader, n_vocab);
+        int32_t n_common_vocab = 0;
+        read_safe(loader, n_common_vocab);
 
         //if (n_vocab != model.hparams.n_vocab) {
         //    WHISPER_LOG_ERROR("%s: invalid model file '%s' (bad vocab size %d != %d)\n",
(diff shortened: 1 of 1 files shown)


src/whisper.cpp
```diff
@@ -1587,8 +1587,8 @@ static bool whisper_model_load(struct whisper_model_loader * loader, whisper_con
 
     // load vocab
     {
-        int32_t n_vocab = 0;
-        read_safe(loader, n_vocab);
+        int32_t n_common_vocab = 0;
```
Jaffe2718, line 1590: Some models, such as `ggml-tiny.en.bin` include special tokens such as `<|endoftext|>` in vocab, which is obviously incorrect, so these tokens should be excluded

src/whisper.cpp
```diff
@@ -1615,32 +1617,35 @@ static bool whisper_model_load(struct whisper_model_loader * loader, whisper_con
...
+        vocab.token_transcribe = vocab.n_vocab - 1506;     // <|transcribe|>
+        vocab.token_solm       = vocab.n_vocab - 1505;     // <|startoflm|>
+        vocab.token_prev       = vocab.n_vocab - 1504;     // <|startofprev|>
+        vocab.token_nosp       = vocab.n_vocab - 1503;     // <|nospeech|>
+        vocab.token_not        = vocab.n_vocab - 1502;     // <|notimestamps|>
+        vocab.token_beg        = vocab.n_vocab - 1501;     // timestamps from <|0.00|> to <|30.00|>, 1501 tokens
+
+        if (n_common_vocab < model.hparams.n_vocab) {
+            WHISPER_LOG_INFO("%s: adding %d extra tokens\n", __func__, model.hparams.n_vocab - n_common_vocab);
+            for (int i = n_common_vocab; i < model.hparams.n_vocab; i++) {
```
Jaffe2718, line 1648: I know this modifies the algorithm for native whisper to define vocab, but native whisper is also hardcoded. For some Whisper models like [whisper-ja-anime-v0.3](https://huggingface.co/efwkjn/whisper-ja-anime-v0.3) on HuggingFace, if the vocab size is modified, it is not compatible with whisper.cpp because the index is off-bounds when loading the vocab of these models.

PR #3617, opened 2026-01-21
whisper: add whisper_vad_set_context to set vad_context in the state

xlmod:
Add the ability to save a VAD context inside a state to prevent loading the VAD model multiple times when VAD is already used outside `whisper_full`, or when you want to load it with something other than the `model_path`.

### DIFF
--- a/include/whisper.h
+++ b/include/whisper.h
@@ -718,6 +718,12 @@ extern "C" {
     WHISPER_API void whisper_vad_free_segments(struct whisper_vad_segments * segments);
     WHISPER_API void whisper_vad_free         (struct whisper_vad_context  * ctx);
 
+    // Inject external VAD context for use with params.vad = true.
+    // The caller retains ownership - whisper will not free this context.
+    // Frees any previously set internal VAD context.
+    WHISPER_API void whisper_state_set_vad(struct whisper_state * state, struct whisper_vad_context * vctx);
+    WHISPER_API void whisper_set_vad      (struct whisper_context * ctx, struct whisper_vad_context * vctx);
+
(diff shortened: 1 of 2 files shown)


include/whisper.h
```diff
@@ -718,6 +718,12 @@ extern "C" {
     WHISPER_API void whisper_vad_free_segments(struct whisper_vad_segments * segments);
     WHISPER_API void whisper_vad_free         (struct whisper_vad_context  * ctx);
 
+    // Inject external VAD context for use with params.vad = true.
+    // The caller retains ownership - whisper will not free this context.
+    // Frees any previously set internal VAD context.
+    WHISPER_API void whisper_vad_set_context_to_state(struct whisper_state * state, struct whisper_vad_context * vctx);
+    WHISPER_API void whisper_vad_set_context         (struct whisper_context * ctx, struct whisper_vad_context * vctx);
```
ggerganov: Rename to:


```suggestion
    WHISPER_API void whisper_state_set_vad(struct whisper_state * state, struct whisper_vad_context * vctx);
    WHISPER_API void whisper_set_vad      (struct whisper_context * ctx, struct whisper_vad_context * vctx);
```

src/whisper.cpp
```diff
@@ -921,6 +921,7 @@ struct whisper_state {
     int32_t exp_n_audio_ctx = 0; // 0 - use default
 
     whisper_vad_context * vad_context = nullptr;
+    bool external_vad = false;
```
ggerganov:
```suggestion
    bool vad_external = false;
```

PR #3647, opened 2026-02-04
ruby : add `Whisper::Context::Params`, fix token memory management

KitaitiMakoto:
Hi,

This pull request

* introduces `Whisper::Context::Params` so that users can use GPU, disable flash attention, etc.
* fixes memory management of `Whisper::Token`

Thank you.

### DIFF
--- a/bindings/ruby/README.md
+++ b/bindings/ruby/README.md
@@ -394,6 +394,20 @@ whisper
   .full(Whisper::Params.new, samples)
 ```
 
+Custom context params
+---------------------
+
+You can use customize `Whisper::Context`'s behavior using `Whisper::Context::Params`.
+
+```ruby
+context_params = Whisper::Context::Params.new(
+  use_gpu: false,
+  flash_attn: false,
+  # etc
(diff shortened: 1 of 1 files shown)


bindings/ruby/README.md
```diff
@@ -247,6 +247,58 @@ whisper.transcribe("path/to/audio.wav", params)
 
 ```
 
+### Tokens ###
+
+Each segment has tokens.
+
+To enable token timestamps, you need set `Whisper::Params#token_timestamps = true`. Then, retrieve tokens from segments using `Whisper::Segment#each_token`.
```
danbev:
```suggestion
To enable token timestamps, you need to set `Whisper::Params#token_timestamps = true`. Then, retrieve tokens from segments using `Whisper::Segment#each_token`.
```

KitaitiMakoto: Thanks!

KitaitiMakoto:
Canceled Ci other than Ruby.

PR #3677, opened 2026-02-23
vad : add streaming detect + explicit state reset

danielbodart:
## Summary

- Add `whisper_vad_detect_speech_no_reset()` — identical to `whisper_vad_detect_speech` but does not reset LSTM hidden/cell state, enabling temporal continuity when calling per-chunk in a streaming loop
- Add `whisper_vad_reset_state()` — explicit state reset for use between utterances
- Refactor `whisper_vad_detect_speech` as a thin wrapper (reset + no_reset) — zero behavior change for existing callers

## Motivation

`whisper_vad_detect_speech` calls `ggml_backend_buffer_clear(vctx->buffer, 0)` on every invocation, which resets the Silero LSTM hidden/cell states. This is correct for batch processing (the current use case), but prevents temporal continuity when calling per-chunk in a streaming loop — the LSTM effectively degrades to a feedforward classifier with no memory between chunks.

For streaming applications that call VAD once per chunk (e.g. 512 samples at 16kHz = 32ms), the model needs to carry state across calls to make use of its recurrent architecture.

## Changes

Two new public API functions following existing naming conventions:

```c
// Like whisper_vad_detect_speech, but does not reset LSTM state.
// Use for streaming: call whisper_vad_reset_state() between utterances.
WHISPER_API bool whisper_vad_detect_speech_no_reset(
        struct whisper_vad_context * vctx,
        const float * samples,
        int   n_samples);

// Reset LSTM hidden/cell states to zero.
WHISPER_API void whisper_vad_reset_state(struct whisper_vad_context * vctx);
```

`whisper_vad_detect_speech` is now `reset + no_reset` — existing callers (including `whisper_vad_segments_from_samples`, `test-vad.cpp`, `examples/speech.cpp`) are completely unaffected.

### DIFF
--- a/include/whisper.h
+++ b/include/whisper.h
@@ -695,6 +695,16 @@ extern "C" {
                            const float * samples,
                                    int   n_samples);
 
+    // Like whisper_vad_detect_speech, but does not reset LSTM state.
+    // Use for streaming: call whisper_vad_reset_state() between utterances.
+    WHISPER_API bool whisper_vad_detect_speech_no_reset(
+            struct whisper_vad_context * vctx,
+                           const float * samples,
+                                   int   n_samples);
+
+    // Reset LSTM hidden/cell states to zero.
+    WHISPER_API void whisper_vad_reset_state(struct whisper_vad_context * vctx);
+
     WHISPER_API int     whisper_vad_n_probs(struct whisper_vad_context * vctx);
     WHISPER_API float * whisper_vad_probs  (struct whisper_vad_context * vctx);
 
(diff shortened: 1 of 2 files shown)


marek-hradil:
This seems like exactly what I need! Any updates on this?

marek-hradil:
@ggerganov Heyy, sorry, don't want to be annoying, but could you have a quick look? Seems like a small change opened since February, which would make the VAD in whisper more useful.

KitaitiMakoto:
danbev, who implemented VAD feature reacted this pull request with eyes, so, he must be recognizing this. I guess he's just busy.

PR #3693, opened 2026-03-06
Do not assume /usr/lib library installation.

waptaff:
Current `pkgconfig` configuration file installation path and its contents assume libraries are installed under `/usr/lib` and this is not always the case, for instance `/usr/lib64` is quite possible under Gentoo Linux.

Thus use the `CMAKE_INSTALL_LIBDIR` variable instead of a hardcoded `lib`.


danbev (approved):
Verified this locally 👍

danbev:
@waptaff Thanks!

PR #3704, opened 2026-03-11
fix: allow flash attention in encoder when DTW is enabled

### LINKED ISSUE #3662: Only disable flash attention in the decoder when dtw flag is enabled
Currently, the only way to use dtw is by disabling flash attention for the encoder and decoder.

However, we only really need to disable it in the decoder for dtw to work.

Acelogic:
## Summary

- DTW token timestamps no longer silently disabled when flash attention is enabled
- Flash attention remains active for encoder self-attention and decoder self-attention
- Only cross-attention falls back to the non-flash path (needed to extract KQ_soft_max weights for DTW)

## Details

Previously, `whisper_init_with_params_no_state` disabled DTW entirely when `flash_attn` was set:
```cpp
if (params.flash_attn && params.dtw_token_timestamps) {
    params.dtw_token_timestamps = false; // DTW silently disabled
}
```

DTW only needs the explicit cross-attention weights (`KQ_soft_max`) from the decoder, which the flash attention path doesn't produce (it fuses Q*K*V into one operation). The encoder self-attention and decoder self-attention don't interact with DTW at all.

The fix introduces a `flash_cross` flag (`flash_attn && !dtw_token_timestamps`) used at two matching locations:
1. **Encoder** (cross-attention KV storage layout) — determines how K/V are stored in `kv_cross`
2. **Decoder** (cross-attention computation) — determines whether to use the fused flash path or the explicit K*Q → softmax → KQV path

Both locations must use the same condition to keep the KV cache layout consistent.

## Test plan

- [x] Builds cleanly on macOS with Metal
- Functional testing requires a whisper model with DTW-compatible alignment heads

Fixes #3662

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -2317,7 +2317,11 @@ static struct ggml_cgraph * whisper_build_graph_cross(
         struct ggml_tensor * k;
         struct ggml_tensor * v;
 
-        if (wctx.params.flash_attn) {
+        // Use non-flash layout for cross-attention KV when DTW is enabled,
+        // since DTW needs the explicit cross-attention weights (KQ_soft_max)
+        const bool flash_cross = wctx.params.flash_attn && !wctx.params.dtw_token_timestamps;
+
+        if (flash_cross) {
             k = ggml_view_1d(ctx0, wstate.kv_cross.k, n_state*n_ctx,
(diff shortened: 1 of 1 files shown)


hlevring:
Is it possible this one can be merged, it would be helpful to get dtw working.

jettoblack:
It's working well for me. I hope this gets merged.

Issue #1798, opened 2024-01-22
CJK character (3 Byte) is split into two tokens in json output. 


HaujetZhao:
## Detail

Noticed that whisper.cpp now can output full json, and it's default output encoding is utf-8. 

When the output characters are all made of ascii characters, it works perfect. 

But when the output  CJK characters, a small issue arises.  

A utf-8 basic: UTF-8 is a variable-length character encoding where common ASCII characters are represented using one byte, while a broader range of characters, including most CJK characters, are encoded using 2 or 3 bytes (mainly 3). 

The encoding rules for UTF-8 are as follows:

- Single-byte encoding ranges from 0x00 to 0x7F (corresponding to ASCII characters).
- Double-byte encoding ranges from 0xC2 80 to 0xDF BF.
- Triple-byte encoding ranges from 0xE0 A0 80 to 0xEF BF BF. 

In the full json output, many CJK characters are frequently separated into two tokens: the first token has two bytes, and the second has one byte. These two tokens are not valid utf-8 characters, causing the json file can't be read using utf-8 encoding. 

![image](https://github.com/ggerganov/whisper.cpp/assets/19181833/583308f7-6948-49b8-a59b-cb7ca62e5ccc)

## Reproduce 

I used the [v1.5.4 Windows binary Release](https://github.com/ggerganov/whisper.cpp/releases/download/v1.5.4/whisper-bin-x64.zip).  

Here is the zipped wav sound file: 

[test-zh.wav.zip](https://github.com/ggerganov/whisper.cpp/files/14013002/test-zh.wav.zip)

Using the command: 

```
main.exe --model ../model/medium.bin --language zh -otxt -ojf  test-zh.wav
```

A txt and a json-full result is produced:

- [test-zh.wav.txt](https://github.com/ggerganov/whisper.cpp/files/14013044/test-zh.wav.txt)
- [test-zh.wav.json](https://github.com/ggerganov/whisper.cpp/files/14013042/test-zh.wav.json)

## Possible solution 

A possible solution is to check if the token is a valid utf-8 character and concat the broken tokens before output to json file.

bobqianic:
This is a known issue where it outputs tokens directly instead of words. I mean, even when using English, sometimes it will give you broken words.

PR #3751, opened 2026-04-11
cli : merge tokens split across UTF-8 boundaries in JSON output

texasich:
When a multi-byte UTF-8 codepoint — most commonly a CJK character, which takes 3 bytes — happens to land across two or more adjacent whisper tokens, the `-ojf`/`--output-json-full` writer was emitting each partial-byte token as its own JSON string. That produces invalid UTF-8, and most downstream parsers rightly choke on it (reported in #1798).

### Root cause

`output_json` in `examples/cli/cli.cpp` iterates at the token level and writes one JSON object per token. The segment-level text helpers (`whisper_full_get_segment_text`) concatenate tokens first, so they're fine — only the per-token loop was broken.

### Fix

Two small additions to `examples/cli/cli.cpp`:

1. **`utf8_trailing_bytes_needed(const std::string &)`** — a static helper that walks back past UTF-8 continuation bytes, finds the last lead byte, and returns how many more continuation bytes are still expected (0 if the tail is already a complete codepoint or looks malformed).

2. **Merge loop in `output_json`** — before writing tokens, adjacent tokens are accumulated into a `std::vector<merged_token>` while the accumulated text still ends on an incomplete UTF-8 sequence. Each merged entry keeps the first token's `id`, `p`, and `t_dtw` and extends `t1` to the last absorbed token. The emission loop then runs over the merged list instead of raw tokens.

No public API change. No format change for codepoints that already fit in one token. Only `output_json` is touched; `output_score` writes tab-separated plain text and never had a JSON-validity issue, so it's left alone.

### Not a duplicate of #3619

PR #3619 (`feat(cli): add word-level LRC output with UTF-8 fix`) adds a new LRC output mode and includes a UTF-8 fix scoped to that new writer. It does not touch `output_json` or address the issue described in #1798. The two changes are independent.

### Testing

Standalone unit tests against the helper (all pass):
- empty string, pure ASCII → 0
- complete and incomplete 2-byte sequences (é)
- complete and incomplete 3-byte sequences (私, the CJK char from the bug report)
- all three partial states of a 4-byte emoji
- incremental merge simulation: 私 in three tokens → merges into one entry

Built cleanly against current master: `cmake --build build --target whisper-cli -j`, no new warnings in `cli.cpp`.

### DIFF
--- a/examples/cli/cli.cpp
+++ b/examples/cli/cli.cpp
@@ -31,6 +31,33 @@ static void replace_all(std::string & s, const std::string & search, const std::
     }
 }
 
+// Returns the number of trailing continuation bytes still needed for `s` to end
+// on a complete UTF-8 codepoint. Returns 0 if the tail of `s` is already a
+// complete codepoint (or if the tail looks malformed and we should stop merging).
+// Used to merge whisper tokens whose bytes split a multi-byte UTF-8 character
+// (e.g. CJK), so the JSON output stays valid UTF-8. See issue #1798.
+static int utf8_trailing_bytes_needed(const std::string & s) {
+    const int n = (int) s.size();
+    int i = n - 1;
+    // walk back past continuation bytes (10xxxxxx)
+    while (i >= 0 && ((unsigned char) s[i] & 0xC0) == 0x80) {
+        --i;
+    }
+    if (i < 0) {
+        // all continuation bytes, or empty — nothing we can do
+        return 0;
+    }
+    const unsigned char c = (unsigned char) s[i];
+    int expected;
+    if      ((c & 0x80) == 0x00) expected = 1; // ASCII
+    else if ((c & 0xE0) == 0xC0) expected = 2;
+    else if ((c & 0xF0) == 0xE0) expected = 3;
+    else if ((c & 0xF8) == 0xF0) expected = 4;
+    else                         return 0;     // malformed lead, give up
+    const int have = n - i;
+    return have >= expected ? 0 : (expected - have);
+}
+
 // command-line parameters
 struct whisper_params {
     int32_t n_threads     = std::min(4, (int32_t) std::thread::hardware_concurrency());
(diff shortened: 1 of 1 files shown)


texasich:
gentle ping — this fixes the CJK character split in JSON output (#1798). verified it doesn't overlap with #3619 which only touches the new LRC writer. happy to iterate if anything needs adjusting.

texasich:
Gentle nudge on this one — been open a while. Happy to rebase if there are conflicts, or make any changes needed. Let me know if anything's blocking review. 🙏

examples/cli/cli.cpp
```diff
@@ -31,6 +31,33 @@ static void replace_all(std::string & s, const std::string & search, const std::
...
+        // all continuation bytes, or empty — nothing we can do
+        return 0;
+    }
+    const unsigned char c = (unsigned char) s[i];
+    int expected;
+    if      ((c & 0x80) == 0x00) expected = 1; // ASCII
+    else if ((c & 0xE0) == 0xC0) expected = 2;
+    else if ((c & 0xF0) == 0xE0) expected = 3;
+    else if ((c & 0xF8) == 0xF0) expected = 4;
+    else                         return 0;     // malformed lead, give up
```
danbev: Please use braces for the body of the if/else statements to be consistent with the rest of the code base:
```suggestion
    if ((c & 0x80) == 0x00) {
        expected = 1;
    } else if ((c & 0xE0) == 0xC0) {
        expected = 2;
    } else if ((c & 0xF0) == 0xE0) {
        expected = 3;
    } else if ((c & 0xF8) == 0xF0) {
        expected = 4;
    } else {
        return 0;
    }
```

examples/cli/cli.cpp
```diff
@@ -31,6 +31,33 @@ static void replace_all(std::string & s, const std::string & search, const std::
     }
 }
 
+// Returns the number of trailing continuation bytes still needed for `s` to end
+// on a complete UTF-8 codepoint. Returns 0 if the tail of `s` is already a
+// complete codepoint (or if the tail looks malformed and we should stop merging).
+// Used to merge whisper tokens whose bytes split a multi-byte UTF-8 character
+// (e.g. CJK), so the JSON output stays valid UTF-8. See issue #1798.
```
danbev: Nit: Please use the full url to the issue (https://github.com/ggml-org/whisper.cpp/issues/1798).

texasich:
Addressed review from @danbev:
- Braces added to if/else chain for consistency
- Full URL for issue #1798 reference

Ready for re-review.

texasich:
Dual-agent Codex review complete — both pass.

**Codex A (diff):** Braces semantic equivalence verified (256-byte simulation), clean compile.

**Codex B (whole-codebase):** No leaks, no unsafe buffers, edge cases handled.

⚠️ Also found: `examples/server/server.cpp` has the same UTF-8 token-split bug — opened #3821.

texasich:
👋 Friendly ping — CI is green and this fixes CJK token boundary handling in JSON output. Let me know if any changes are needed.

danbev (approved):

PR #3766, opened 2026-04-19
whisper : correct per-token timestamps in parallel processing merge (#3726)

### LINKED ISSUE #3726: `--processors` /` n_processors` greater than 1 _still_ produces incorrect token timestamps
hi, i would like to flag #2036 as not being fixed, see here the following reproduction

when running

```sh
build/bin/whisper-cli -m models/ggml-large-v3-turbo.bin --output-file domecium --output-json-full --language en -f domecium.mp3 --processors 2
```

on this file from this [short poetry collection](https://librivox.org/short-poetry-collection-265-by-various/)

[domecium.mp3](https://github.com/user-attachments/files/26315445/domecium.mp3)

it mentions that it has been split around 1:35

```
whisper_full_parallel: the audio has been split into 2 chunks at the following times:
whisper_full_parallel: split 1 - 00:01:35.460
whisper_full_parallel: the transcription quality may be degraded near these boundaries
```

When we look at [domecium.json](https://github.com/user-attachments/files/26315429/domecium.json), we see the token level timestamps reset around that boundary

```json
			{
					"text": "[_TT_1266]",
					"timestamps": {
						"from": "00:01:35,460",
						"to": "00:01:35,460"
					},
					"offsets": {
						"from": 95460,
						"to": 95460
					},
					"id": 51631,
					"p": 0.95917,
					"t_dtw": -1
				}
			]
		},
		{
			"timestamps": {
				"from": "00:01:35,460",
				"to": "00:01:47,640"
			},
			"offsets": {
				"from": 95460,
				"to": 107640
			},
			"text": " in days bygone long gone my father's mother who is now blest with the blest would take me out to walk",
			"tokens": [
				{
					"text": "[_BEG_]",
					"timestamps": {
						"from": "00:00:00,000",
						"to": "00:00:00,000"
					},
					"offsets": {
						"from": 0,
						"to": 0
					},
					"id": 50365,
					"p": 0.997823,
					"t_dtw": -1
				},
				{
					"text": " in",
					"timestamps": {
						"from": "00:00:00,000",
						"to": "00:00:00,290"
					},
					"offsets": {
						"from": 0,
						"to": 290
					},
					"id": 294,
					"p": 0.75974,
					"t_dtw": -1
				},
```

So the token level timestamps still reset. The segment level timestamps are still correct however.

achyutbenz19:
## Summary

Fixes #3726.

`whisper_full_parallel` splits the input audio into N chunks and spawns one worker per chunk to call `whisper_full_with_state`. When merging worker results back into the main state, it offset-corrects each segment's `t0` and `t1` by the chunk's starting sample, but leaves the per-token timestamps (`result.tokens[j].t0` / `.t1`) in chunk-relative time. Token timestamps therefore reset to `00:00:00` at every chunk boundary while segment timestamps show correct absolute times, so the two disagree in every worker beyond the first.

## Scope of the change

`src/whisper.cpp`, +11/-7 on the merge loop of `whisper_full_parallel`.

1. Hoist the repeated offset expression `100 * ((i + 1) * n_samples_per_processor) / WHISPER_SAMPLE_RATE + offset_t` into a local `chunk_offset`.
2. Apply the same offset to every `token.t0` / `token.t1` inside `result.tokens` before the segment is pushed into `ctx->state->result_all`.

Non-parallel (`n_processors == 1`) is unchanged: that branch returns early via `whisper_full` and never enters this merge loop.

## Reproduction

On current master (`166c20b`), with `ggml-base.bin` and `long-en-70s.wav`:

```
whisper-cli -m ggml-base.bin --processors 2 -ojf -of out long-en-70s.wav
```

Sample from `out.json` (before this patch):

```
segment 00:00:38.600 -> 00:00:41.400 | token[0].t0 = 00:00:03,750
segment 00:00:41.400 -> 00:00:43.960 | token[0].t0 = 00:00:06,550
segment 00:01:08.440 -> 00:01:10.440 | token[0].t0 = 00:00:33,440
```

All three tokens show chunk-relative times (relative to the start of worker 1's slice) while the segments are correctly absolute.

After this patch:

```
segment 00:00:38.600 -> 00:00:41.400 | token[0].t0 = 00:00:38,750
segment 00:00:41.400 -> 00:00:43.960 | token[0].t0 = 00:00:41,550
segment 00:01:08.440 -> 00:01:10.440 | token[0].t0 = 00:01:08,440
```

Token and segment timelines agree.

## Differential matrix

`model=base`, `fixture ∈ {long-en-70s, long-en-55s, speech-en}`, `procs ∈ {1, 2, 3}`. 9 cells per build.

| 9 | 6 | 6 | 0 | 3 | 0 |

Target cells (`procs ∈ {2, 3}`) improve: every `-ojf` output now has token timestamps agreeing with segment timestamps. Non-target cells (`procs=1`) are byte-identical across master and this patch, confirming single-worker transcriptions are untouched.

## What this does not do

- Does not change segment-level timestamps, which were already correct.
- Does not change decoder behavior, beam-search, or any audio processing. The only change is a post-decode offset applied to token metadata in one merge loop.
- Does not interact with the VAD path (VAD runs inside `whisper_full`, not `whisper_full_parallel`; #3754 covered that path).

## Tools used

`git`, `cmake`, `whisper-cli`, and [`audiokit`](https://github.com/YouLearn-AI/audiokit) for the differential matrix.

## Disclosure

I am an AI assistant (Anthropic's Claude) helping a user contribute this fix. Numbers above come from actual runs against commit `166c20b` on an Apple Silicon Mac. The regress config and raw per-cell outputs are available.

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -6812,7 +6812,21 @@ int whisper_full_with_state(
     if (params.language == nullptr || strlen(params.language) == 0 || strcmp(params.language, "auto") == 0 || params.detect_language) {
         std::vector<float> probs(whisper_lang_max_id() + 1, 0.0f);
 
-        const auto lang_id = whisper_lang_auto_detect_with_state(ctx, state, 0, params.n_threads, probs.data());
+        // Respect the user's offset_ms when choosing the window for language
+        // detection. Previously this was hardcoded to 0, which meant that on
+        // audio like "1 minute of French then 30 minutes of German" with
+        // offset_ms=60000, transcription correctly started at the 1-minute
+        // mark but language detection still picked the French prefix.
+        // If the offset falls past the end of the audio (including short
+        // files where offset_ms exceeds total duration), fall back to 0 so
+        // auto-detect still returns a valid language. The too-short-audio
+        // guard later in this function will then handle the empty-decode
+        // case cleanly.
+        // ref: https://github.com/ggml-org/whisper.cpp/issues/1831
+        const int detect_offset_ms = (params.offset_ms > 0 && (params.offset_ms / 10) < state->mel.n_len_org)
+            ? params.offset_ms
+            : 0;
+        const auto lang_id = whisper_lang_auto_detect_with_state(ctx, state, detect_offset_ms, params.n_threads, probs.data());
         if (lang_id < 0) {
             WHISPER_LOG_ERROR("%s: failed to auto-detect language\n", __func__);
             return -3;
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -6812,21 +6812,7 @@ int whisper_full_with_state(
     if (params.language == nullptr || strlen(params.language) == 0 || strcmp(params.language, "auto") == 0 || params.detect_language) {
         std::vector<float> probs(whisper_lang_max_id() + 1, 0.0f);
 
-        // Respect the user's offset_ms when choosing the window for language
-        // detection. Previously this was hardcoded to 0, which meant that on
-        // audio like "1 minute of French then 30 minutes of German" with
-        // offset_ms=60000, transcription correctly started at the 1-minute
-        // mark but language detection still picked the French prefix.
-        // If the offset falls past the end of the audio (including short
-        // files where offset_ms exceeds total duration), fall back to 0 so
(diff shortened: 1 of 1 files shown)


danbev:

PR #3782, opened 2026-04-29
server: Return speaker information in JSON

alubbe:
This PR includes the speaker information in the JSON response of the server, and also makes the wav audio stereo in diarization mode.


alubbe:

For this PR specifically, my issue is that I'm migrating an application from using the CLI to using one central server and there a) was no way to get speaker information from the JSON and b) all audio was forced into mono.

alubbe:
For comparison, openAI also uses the `speaker` field for diarization: https://developers.openai.com/api/docs/guides/speech-to-text#speaker-diarization

danbev (approved):

PR #3783, opened 2026-04-30
server: Fix no_speech_thold not being read

alubbe:
Somehow we forgot to include it in `get_req_parameters`


alubbe:

For this PR specifically, my issue is that I'm migrating an application from using the CLI to using one central server and this param was missing in the server.

danbev (approved):

PR #3784, opened 2026-04-30
server: Fix params leak between requests

alubbe:
Example:

1. Request A sends bad audio, but also includes:
   ```bash
   -F carry_initial_prompt="true" \
   -F prompt="Always spell Kubernetes as K8s"
   ```

2. The server parses those fields into the shared `params`.

3. Then Request A fails while reading the audio and returns early.

4. Before the fix, the server never reached:
   ```cpp
   params = default_params;
   ```

5. Request B sends a normal transcription request without any prompt.

6. The server could still use Request A’s stale settings, so Request B may unexpectedly behave as if it had:
   ```bash
   -F carry_initial_prompt="true" \
   -F prompt="Always spell Kubernetes as K8s"
   ```

That is confusing because Request B did nothing wrong, but its output can be influenced by a previous failed request. The fix makes every request start from a fresh local copy of the default params, so failed requests cannot contaminate later ones.

### DIFF
--- a/examples/server/server.cpp
+++ b/examples/server/server.cpp
@@ -817,7 +817,7 @@ int main(int argc, char ** argv) {
         }
         auto audio_file = req.get_file_value("file");
 
-        // check non-required fields
+        whisper_params params = default_params;
         get_req_parameters(req, params);
 
         std::string filename{audio_file.filename};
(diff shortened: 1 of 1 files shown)


alubbe:

For this PR specifically, my issue is that I'm migrating an application from using the CLI to using one central server and this bug has hit us unexpectledy.

danbev (approved):

PR #3785, opened 2026-04-30
server: Add support for controlling token_timestamps directly

alubbe:


alubbe:

For this PR specifically, my issue is that I'm migrating an application from using the CLI to using one central server and using `no_timestamps` wasn't granular enough for us, we need to able to disable `token_timestamps` without losing timestamps as a whole.

ggerganov (approved):

danbev (approved):
Verified this manually against local whisper-server 👍

PR #3790, opened 2026-05-05
metal: fix flash attention nsg8 dv64 path

adavyas:
- Fix Metal FlashAttention single-output-tile handling. Adds an explicit NO == 1 path so the kernel loads one value tile and accumulates into lo[0] instead of skipping the multiply-accumulate because NO/2 == 0.
- Preserve the existing paired-tile fast path. Keeps the current two-tile loop for NO > 1, while adding a safe remainder/small-shape path for tensor shapes that only produce one output tile.


ggerganov:
In which situation do you encounter this problem?

adavyas:
this was the shape that i had changed: Q=8,  C=64,  NSG=8, DK=64, DV=64

all these values are the defaults except NSG=8. baseline is NSG-4

PR #3798, opened 2026-05-09
Fix #984: About max_tokens skipping remaining audio issue

### LINKED ISSUE #984: SetMaxTokensPerSegment in go binding causes the segements to end prematurely
When I use `context.SetMaxTokensPerSegment` and `context.NextSegment`, it generates incomplete segments. e.g. when tested using the jfk.wav sample, it gives the following output:

```
wspr_test.go:33: {Text:And so my fellow Americans ask Start:0s End:30s}
```

when this value is not set, it gives the correct output:

```
wspr_test.go:33: {Text:And so my fellow Americans ask not what your country can do for you Start:0s End:8s}
wspr_test.go:33: {Text:ask what you can do for your country. Start:8s End:11s}
```

test case: https://github.com/JackKCWong/go-whisper-test/blob/main/pkg/whisper/wspr_test.go#L29

annaeina:
When a user sets a limit via `max_tokens` (e.g., through Go bindings' `SetMaxTokensPerSegment()`), the transcription often terminates prematurely. This happens because if the token limit is hit before a natural timestamp token is generated, the decoder finishes the current segment but leaves `seek_delta` pointing to the end of the current 30s audio chunk. Consequently, the `seek` pointer skips the remainder of the audio, returning only the first truncated segment.

This patch changes the decoding behavior for normal timestamped transcription:

- when `max_tokens` is reached, suppress text tokens so the next sampled token is a timestamp
- use that timestamp as the segment boundary
- avoid treating this forced timestamp ending as the special `single_timestamp_ending` case that skips the full chunk

The existing behavior for `single_segment` and `no_timestamps` is left unchanged.

Added a Go regression test that sets `SetMaxTokensPerSegment(5)` on the JFK sample and checks that more than one segment is returned and later text is still present.

Tested locally with `ggml-small.en.bin`:

Before:

```text
[0s->30s] And so, my fellow
```

After:

```text
[0s->1.04s]  And so, my
[1.04s->3.84s]  fellow Americans, ask
[3.84s->6.34s]  not what your country
[6.34s->8.1s]  can do for you
[8.1s->9.38s]  ask what you can
[9.38s->12.38s]  do for your country
```

Regression test:

```text
--- PASS: TestProcessMaxTokensPerSegment
PASS
```

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -6216,6 +6216,12 @@ static void whisper_process_logits(
             }
         }
 
+        if (!params.no_timestamps && !params.single_segment && params.max_tokens > 0 && (int) tokens_cur.size() >= params.max_tokens) {
+            for (int i = 0; i < vocab.token_eot; ++i) {
+                logits[i] = -INFINITY;
+            }
+        }
+
         // suppress sot and nosp tokens
         logits[vocab.token_sot]  = -INFINITY;
         logits[vocab.token_nosp] = -INFINITY;
@@ -7722,7 +7728,12 @@ int whisper_full_with_state(
             }
 
             // ref: https://github.com/ggml-org/whisper.cpp/pull/2629
+            const bool max_tokens_timestamp_ending = params.max_tokens > 0 &&
+                !params.single_segment &&
+                tokens_cur.size() > (size_t) params.max_tokens;
+
             const bool single_timestamp_ending = tokens_cur.size() > 1 &&
+                !max_tokens_timestamp_ending &&
                 tokens_cur[tokens_cur.size() - 2].id < whisper_token_beg(ctx) &&
                 tokens_cur[tokens_cur.size() - 1].id > whisper_token_beg(ctx);
             if (single_timestamp_ending) {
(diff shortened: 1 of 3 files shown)


src/whisper.cpp
```diff
@@ -6216,6 +6216,12 @@ static void whisper_process_logits(
             }
         }
 
+        if (!params.no_timestamps && !params.single_segment && params.max_tokens > 0 && (int) tokens_cur.size() >= params.max_tokens) {
```
ggerganov, line 6220:
```suggestion
        // ref: https://github.com/ggml-org/whisper.cpp/pull/3798
        if (!params.no_timestamps && !params.single_segment && params.max_tokens > 0 && (int) tokens_cur.size() >= params.max_tokens) {
```

ggerganov (approved):

danbev (approved):

annaeina:
Hi @danbev, 
I added a CI fix (overwrite: true) to resolve the artifact collision. Waiting for workflow approval. Thanks!

PR #3818, opened 2026-05-21
fix: server /inference fails to decode in-memory audio (regression)

ServeurpersoCom:
### fix: server /inference fails to decode in-memory audio (regression)

whisper-server /inference returns HTTP 400 on any valid WAV (including samples/jfk.wav) for builds without WHISPER_FFMPEG and without --convert: "error: failed to read audio data from (RIFF...)".

Cause: the handler passes the uploaded bytes audio_file.content to read_audio_data, which since the miniaudio migration only handles a filename via ma_decoder_init_file. It tries to open a path starting with RIFF and fails.

Regression introduced by 7d3da68f (#2759, "examples : use miniaudio for direct decoding"), which dropped the is_wav_buffer + drwav_init_memory in-memory path that read_wav previously used. #3707 later added the 400 status that surfaced it as Invalid request.

Fix: factor the PCM extraction into a shared helper and add a read_audio_data(const char* buffer, size_t size, ...) overload that decodes from memory via ma_decoder_init_memory (already used by the function for the stdin path). server.cpp calls it with the upload content. The filename overload is unchanged.

Verified: samples/jfk.wav now returns 200 on the /inference endpoint instead of 400, and transcription quality checked against my production setup.

AI Usage: Opus 4.7 + MCP computer use (rootless container with GPU)

### DIFF
--- a/examples/common-whisper.cpp
+++ b/examples/common-whisper.cpp
@@ -39,6 +39,42 @@
 extern bool ffmpeg_decode_audio(const std::string & ifname, std::vector<uint8_t> & wav_data);
 #endif
 
+// extract f32 PCM frames from an initialized decoder, downmix to mono and keep the stereo split
+static bool read_audio_from_decoder(ma_decoder & decoder, std::vector<float>& pcmf32, std::vector<std::vector<float>>& pcmf32s, bool stereo) {
+    ma_result result;
+    ma_uint64 frame_count;
+    ma_uint64 frames_read;
+
+    if ((result = ma_decoder_get_length_in_pcm_frames(&decoder, &frame_count)) != MA_SUCCESS) {
+        fprintf(stderr, "error: failed to retrieve the length of the audio data (%s)\n", ma_result_description(result));
+        return false;
+    }
+
+    pcmf32.resize(stereo ? frame_count*2 : frame_count);
+
+    if ((result = ma_decoder_read_pcm_frames(&decoder, pcmf32.data(), frame_count, &frames_read)) != MA_SUCCESS) {
+        fprintf(stderr, "error: failed to read the frames of the audio data (%s)\n", ma_result_description(result));
+        return false;
+    }
+
+    if (stereo) {
+        std::vector<float> stereo_data = pcmf32;
+        pcmf32.resize(frame_count);
+        for (uint64_t i = 0; i < frame_count; i++) {
+            pcmf32[i] = (stereo_data[2*i] + stereo_data[2*i + 1]);
+        }
+        pcmf32s.resize(2);
+        pcmf32s[0].resize(frame_count);
+        pcmf32s[1].resize(frame_count);
+        for (uint64_t i = 0; i < frame_count; i++) {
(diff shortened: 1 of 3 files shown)


examples/common-whisper.cpp
```diff
@@ -39,6 +39,42 @@
 extern bool ffmpeg_decode_audio(const std::string & ifname, std::vector<uint8_t> & wav_data);
 #endif
 
+// extract f32 PCM frames from an initialized decoder, downmix to mono and keep the stereo split
+static bool read_audio_from_decoder(ma_decoder & decoder, std::vector<float>& pcmf32, std::vector<std::vector<float>>& pcmf32s, bool stereo) {
```
danbev: Nit: The codebase tries to use a space before the `&` like the first parameter (`decoder`) does.

examples/common-whisper.cpp
```diff
@@ -109,41 +145,22 @@ bool read_audio_data(const std::string & fname, std::vector<float>& pcmf32, std:
...
-        std::vector<float> stereo_data = pcmf32;
-        pcmf32.resize(frame_count);
+    return read_audio_from_decoder(decoder.decoder, pcmf32, pcmf32s, stereo);
+}
 
-        for (uint64_t i = 0; i < frame_count; i++) {
-            pcmf32[i] = (stereo_data[2*i] + stereo_data[2*i + 1]);
-        }
+// decode audio bytes already held in memory
+bool read_audio_data(const char * buffer, size_t buffer_size, std::vector<float>& pcmf32, std::vector<std::vector<float>>& pcmf32s, bool stereo) {
```
danbev: Nit: The codebase tries to use a space before the `&` similar to the `*`  in the first parameter (`buffer`).

examples/server/server.cpp
```diff
@@ -868,7 +868,7 @@ int main(int argc, char ** argv) {
             // remove temp file
             std::remove(temp_filename.c_str());
         } else {
-            if (!::read_audio_data(audio_file.content, pcmf32, pcmf32s, params.diarize))
+            if (!::read_audio_data(audio_file.content.data(), audio_file.content.size(), pcmf32, pcmf32s, params.diarize))
             {
```
danbev: Nit: Not part of this PR but this opening bracket should be on the line above to be consistent.

danbev (approved):

danbev:
@ServeurpersoCom Thanks! 
Merging without running CI again (verified the changes locally).

ServeurpersoCom:
I use it daily on a Telegram bot to transcribe all audio messages, and as a source for GGML voice cloning pipeline and llama.cpp input.

With the following WAV format:

```
Audio
Format                                   : PCM
Paramètres du format                     : Little / Signed
Identifiant du codec                     : 1
Durée                                    : 44s 485 ms
Type de débit                            : Constant
Débit                                    : 256 kb/s
Canaux                                   : 1 canal
Echantillonnage                          : 16,0 kHz
Profondeur des couleurs                  : 16 bits
Taille du flux                           : 1,36 Mio (100%)
```

PR #3847, opened 2026-05-31
ci : refactor + optimize

ggerganov:
WIP

Started some refactoring of the CI workflows. The goal is to restructure similar to how the workflows are organized now in [llama.cpp](https://github.com/ggml-org/llama.cpp/tree/master/.github/workflows).

- [x] Split `build.yml` into separate workflow. Using [fork](https://github.com/danbev/whisper.cpp/actions).
- [x] Optimize `ccache` usage. Using [fork](https://github.com/danbev/whisper.cpp/actions).
- [x] Remove obsolete/redundant jobs. Using [fork](https://github.com/danbev/whisper.cpp/actions).
- [x] Add linux build artifacts to release. Example [release](https://github.com/danbev/whisper.cpp/releases/tag/untagged-a82262f5cbc2f584ca82)
- [x] Add HF_TOKEN as secret and use in model download script to avoid huggingface rate limiting. Done in [fork](https://github.com/danbev/whisper.cpp/actions/runs/26891371447/job/79318030775#step:3:8) and in this repo.


danbev:
As mentioned in https://github.com/ggml-org/whisper.cpp/pull/3847/commits/70a62ea151fe3673b39811f87370a0b9ebfd49c8 I think we should add a HF_TOKEN to avoid the issue with rate limits for model downloads. This can happen to any of the jobs that download models and just having the retry does not seem to be enough.

danbev:
I've skipped  the freebsd job as this as not enabled previously (it was commented out). Instead of commenting it out I think we can just disable that job once this is merged.

danbev:
There are some naming inconsistencies at the moment regarding the CI jobs but I'll open a follow-up PR to fix that.

danbev:
@ggerganov I don't seem to be able to add  you as a reviewer as you opened this PR. 
Can you take a look when you get a chance and see if this is what you had in mind?

ggerganov:
Some follow-up tasks:

- For all github actions, make sure we are using `ggml-org` forks. This is for security reasons, to avoid 3rd-parties being highjacked.
- Decide which jobs to remove. Mainly keep track of the [largest caches](https://github.com/ggml-org/whisper.cpp/actions/caches?query=sort%3Asize-desc) and see what makes sense to keep. For example, the ~1.7GB msys caches seem excessive and would most likely have to be removed. If after removing the cache, the jobs become very slow (more than 20 mins), then we will disable these jobs. etc.
- Identify slow linux jobs that don't currently utilize `ccache` and update them to start using it

danbev:
I've disabled the freebsd job: https://github.com/ggml-org/whisper.cpp/actions/workflows/build-freebsd.yml

ggerganov:
Also, we have to do something about the Docker images. It does not make sense to build them on every commit. It's probably best to do it like in `llama.cpp` - once per day: https://github.com/ggml-org/llama.cpp/actions/runs/26933499193/workflow

danbev:
- [x] I've pinned actions to specific commit sha's to avoid versions or tags in PRs that touched github actions. The remaining are covered by https://github.com/ggml-org/whisper.cpp/pull/3865.
- [x] The "For example, the ~1.7GB msys caches seem excessive and would most likely have to be removed. " has been [resolved](https://github.com/ggml-org/whisper.cpp/pull/3858).
- [x] Docker images are now [scheduled](https://github.com/ggml-org/whisper.cpp/pull/3854).
- [ ] Most jobs have been updated to now use ccache. For example https://github.com/ggml-org/whisper.cpp/pull/3859. I'll take another look at if there are still jobs that are very slow.
- [x] ~Go through the github actions we use and updated them to use ggml-org forks (creating new forks for the ones that we are using).~ Pinning will en enough for these third-party github actions.

ggerganov:
> Go through the github actions we use and updated them to use ggml-org forks

Pinning to commit hashes of 3rd-party repos should be OK.

PR #3899, opened 2026-06-21
include parakeet in build-xcframework.sh

daaku:
This will enable using parakeet from the pre-built xcframework libraries. Tested this by using the `build-apple/whisper.xcframework` in my application and successfully using the parakeet APIs.

### DIFF
--- a/build-xcframework.sh
+++ b/build-xcframework.sh
@@ -115,6 +115,7 @@ setup_framework_structure() {
 
     # Copy all required headers (common for all platforms)
     cp include/whisper.h           ${header_path}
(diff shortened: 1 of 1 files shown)


danbev (approved):

danbev:
@daaku Thanks!

daaku:
In fact, I should be the one thanking you for adding support for Parakeet. It's a much faster model, and in the last few days I have already used it to save me from typing a whole lot of words, including most of this message ;)

I especially appreciate how trivially I've replaced my existing use of the whisper model, both on my Linux as well as my macOS speech to text tools thanks to the fact that you maintained virtually the same API.

PR #3905, opened 2026-06-23
whisper : optional ANEForge encoder backend (Apple Neural Engine)

sbryngelson:
As offered in #3903, this adds an optional encoder backend that runs the Whisper encoder directly on the Apple Neural Engine through [ANEForge](https://github.com/sbryngelson/ANEForge), as an alternative to the CoreML encoder. On M-series it is about 2x faster than the CoreML encoder and faster than the Metal GPU encoder, at ~5x lower energy, with the same transcripts.

Per-call encoder latency, trained checkpoints, measured with `whisper-bench` (CoreML and Metal are the existing backends, fp16):

| model  | ANEForge (ANE) | CoreML (ANE) | Metal (GPU) | vs CoreML | vs Metal |
| ------ | -------------: | -----------: | ----------: | --------: | -------: |
| tiny   |     5.7 ms     |    11.2 ms   |    7.3 ms   |   2.0x    |   1.3x   |
| base   |    12.2 ms     |    23.0 ms   |   13.5 ms   |   1.9x    |   1.1x   |
| small  |    40.3 ms     |    77.2 ms   |   40.9 ms   |   1.9x    |   1.0x   |
| medium |   117.6 ms     |   236.0 ms   |  120.0 ms   |   2.0x    |   1.0x   |

Cosine 0.999 against the reference encoder; the transcript of `jfk.wav` is unchanged across tiny/small/medium, and a 199 s real-speech clip matches the stock Metal encoder to within 3 words out of 370 (the expected variation between two fp16 encoders).

## How it hooks in

It follows the existing CoreML/OpenVINO seam exactly. The backend fills `embd_enc` through the same `whisper_encode_external` path, gated at runtime by the `ANEFORGE_ENCODER` env var:

- `src/aneforge/whisper-aneforge.{h,cpp}` (the backend), one line in `src/CMakeLists.txt`, and ~30 lines in `src/whisper.cpp` (an include, a state field, the encode branch, init, and free).
- With `ANEFORGE_ENCODER` unset the build behaves exactly like stock whisper.cpp. The backend `dlopen`s the ANEForge dispatch dylib only when enabled, so there is **no new link-time dependency** and nothing changes for existing users.
- Encoder only; the autoregressive decoder is untouched.

```sh
ANEFORGE_DYLIB=... ANEFORGE_ENCODER=./whisper-small-ane \
  ./build/bin/whisper-cli -m models/ggml-small.bin -f samples/jfk.wav
```

## Notes / caveats

- Apple Silicon only. The ANE work lives in the external `aneforge` package (pip), which reaches the ANE through Apple private Espresso `e5rt` symbols, the same stack CoreML uses internally. Undocumented, can change with macOS, so this is a research path, gated behind the env var.
- The exported encoder bundle is built on the target Mac at load (~0.6 s) since it is keyed to the OS build; it is fixed to `n_audio_ctx=1500`, so it does not honor `--audio-ctx`.
- Builds clean on current `master` (verified `whisper-cli`).

Export script, the full benchmark, and reproduction steps are in the companion repo: https://github.com/sbryngelson/whisper-aneforge

Happy to adjust the seam, naming, or gating to fit how you would prefer an optional backend to land.

### DIFF
--- a/src/aneforge/whisper-aneforge.cpp
+++ b/src/aneforge/whisper-aneforge.cpp
@@ -1,0 +1,114 @@
+// ANEForge encoder backend (see whisper-aneforge.h). dlopen's the ANEForge dispatch
+// dylib (libane_e5rt_dispatch.dylib) so no extra link configuration is needed; the
+// path comes from ANEFORGE_DYLIB. Compiles the persisted encoder MIL once at init
+// (compile-with-cache; the on-device program is not cold-loadable across processes),
+// then dispatches it per encode.
+#include "whisper-aneforge.h"
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <cstdint>
+#include <string>
+#include <vector>
+#include <dlfcn.h>
+
+typedef struct ane_e5rt_program ane_e5rt_program_t;
+typedef ane_e5rt_program_t * (*compile_fn)(const char *, const char *, uint64_t,
+    const char * const *, const size_t *, size_t, const char * const *, const size_t *, size_t);
+typedef int  (*set_in_fn)(ane_e5rt_program_t *, const char *, const uint16_t *, size_t);
+typedef int  (*get_out_fn)(ane_e5rt_program_t *, const char *, uint16_t *, size_t);
+typedef int  (*exec_fn)(ane_e5rt_program_t *);
+typedef void (*release_fn)(ane_e5rt_program_t *);
+
+struct whisper_aneforge_context {
+    void * dl = nullptr;
+    ane_e5rt_program_t * prog = nullptr;
+    set_in_fn set_input = nullptr;
+    get_out_fn get_output = nullptr;
+    exec_fn execute = nullptr;
+    release_fn release = nullptr;
+    std::string mel_port, pos_port, out_port;
+    size_t mel_n = 0, pos_n = 0, out_n = 0;
+    std::vector<uint16_t> mel16;
+    std::vector<uint16_t> out16;
+};
+
+// arm64 has a native IEEE binary16 type; these casts compile to NEON fcvt and
+// auto-vectorize, unlike branchy scalar bit-twiddling. ANEForge's fp16 ports are
+// IEEE binary16, which __fp16 matches bit-for-bit.
+static inline void f32_to_f16(const float * src, uint16_t * dst, size_t n) {
+    __fp16 * d = (__fp16 *) dst;
+    for (size_t i = 0; i < n; i++) d[i] = (__fp16) src[i];
+}
+static inline void f16_to_f32(const uint16_t * src, float * dst, size_t n) {
+    const __fp16 * s = (const __fp16 *) src;
+    for (size_t i = 0; i < n; i++) dst[i] = (float) s[i];
+}
+
+struct whisper_aneforge_context * whisper_aneforge_init(const char * bundle_dir) {
+    const char * dylib = getenv("ANEFORGE_DYLIB");
+    if (!dylib) { fprintf(stderr, "aneforge: set ANEFORGE_DYLIB to libane_e5rt_dispatch.dylib\n"); return nullptr; }
+    void * dl = dlopen(dylib, RTLD_NOW | RTLD_LOCAL);
+    if (!dl) { fprintf(stderr, "aneforge: dlopen(%s) failed: %s\n", dylib, dlerror()); return nullptr; }
+
+    auto ctx = new whisper_aneforge_context();
+    ctx->dl = dl;
+    auto compile  = (compile_fn) dlsym(dl, "ane_e5rt_program_compile");
+    ctx->set_input  = (set_in_fn)  dlsym(dl, "ane_e5rt_program_set_input_fp16");
+    ctx->get_output = (get_out_fn) dlsym(dl, "ane_e5rt_program_get_output_fp16");
+    ctx->execute    = (exec_fn)    dlsym(dl, "ane_e5rt_program_execute");
+    ctx->release    = (release_fn) dlsym(dl, "ane_e5rt_program_release");
+    if (!compile || !ctx->set_input || !ctx->get_output || !ctx->execute || !ctx->release) {
+        fprintf(stderr, "aneforge: missing dispatch symbols\n"); whisper_aneforge_free(ctx); return nullptr;
+    }
+
+    // ports.txt: three lines "name nelems" for mel, pos, output (in that order).
+    std::string dir = bundle_dir;
+    FILE * pf = fopen((dir + "/ports.txt").c_str(), "r");
+    if (!pf) { fprintf(stderr, "aneforge: no ports.txt in %s\n", bundle_dir); whisper_aneforge_free(ctx); return nullptr; }
+    char nm[256]; size_t ne;
+    std::string * ports[3] = {&ctx->mel_port, &ctx->pos_port, &ctx->out_port};
+    size_t * nes[3] = {&ctx->mel_n, &ctx->pos_n, &ctx->out_n};
+    for (int i = 0; i < 3; i++) { if (fscanf(pf, "%255s %zu", nm, &ne) != 2) { fclose(pf); whisper_aneforge_free(ctx); return nullptr; } *ports[i] = nm; *nes[i] = ne; }
+    fclose(pf);
+
+    std::string mil = dir + "/model.mil", cache = dir + "/cache";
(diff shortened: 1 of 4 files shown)


ggerganov:
Quite interesting work.

> Query-tiled attention. The score matrix is never fully materialized. The query axis is split into tiles, so each head computes [S, S/3] score tiles rather than the full [S, S] (the flash-attention idea, written as einsum so there is no transpose). This roughly halves attention time, is exact (each tile attends to all keys), and is what puts the ANE ahead of the GPU rather than even with it.

Have you investigated how the number of tiles affects the performance? Is 3 tiles always the best option for all model sizes and hardware?

sbryngelson:
> Quite interesting work.
> 
> > Query-tiled attention. The score matrix is never fully materialized. The query axis is split into tiles, so each head computes [S, S/3] score tiles rather than the full [S, S] (the flash-attention idea, written as einsum so there is no transpose). This roughly halves attention time, is exact (each tile attends to all keys), and is what puts the ANE ahead of the GPU rather than even with it.
> 
> Have you investigated how the number of tiles affects the performance? Is 3 tiles always the best option for all model sizes and hardware?

The count isn't fixed at 3 @ggerganov, sharp question and some learning for both of us here. It's n_tiles = max(1, (S+256)//512), so it scales with the query length, and the encoder's S=1500 gets 3. I swept it on an M5 Pro across the model sizes (S=1500, head_dim=64, n_head 6 to 20), timing the attention core as one fused program:

```
size    H | t=1  t=2  t=3  t=4  t=5  t=6  t=8   (ms)
tiny    6 | 2.50 2.65 2.57 2.54 2.41 2.68 2.51
base    8 | 3.28 3.47 3.33 3.33 3.06 3.51 3.17
small  12 | 5.07 5.45 5.12 5.15 4.90 5.22 5.00
medium 16 |20.31 7.16 6.84 7.12 6.52 6.91 6.59
large  20 |26.83 8.99 8.77 8.60 8.03 8.59 8.20
```

3 is close, but not optimal: 5 is consistently 5-10% faster here. And the first-order effect is just tiling at all. At high head count, the untiled [H,1500,1500] score is about 3x slower (it stops pipelining on the engine), while for tiny and base, it fits in cache and tiling is roughly flat.

The optimum is broad (3 to 8 tiles within 10%), and it's bounded by L2 residency and pipelining, so it moves with the chip, as you noticed.

Rather than a constant, the tile count is now autotuned in ANEForge (I merged a PR for this): it measures the best count once per (chip, S, T, heads, head_dim), caches it, and falls back to the heuristic until a tuned value exists. On the M5 Pro, it picks 5; on the M1/M2, it may pick something else, which is fine. The autotune picks that up with no code change.

sbryngelson:
Following up on the tile-count question. Short version: it is no longer a constant. ANEForge autotunes the count once per (chip, S, T, heads, head_dim) and caches it, so the 3 in the original patch is gone.

Is there anything else you want addressed before this can be considered? I am happy to put it behind a build flag or split it into smaller PRs if that makes review easier.

src/aneforge/whisper-aneforge.cpp
```diff
@@ -0,0 +1,114 @@
...
+    for (size_t i = 0; i < n; i++) d[i] = (__fp16) src[i];
+}
+static inline void f16_to_f32(const uint16_t * src, float * dst, size_t n) {
+    const __fp16 * s = (const __fp16 *) src;
+    for (size_t i = 0; i < n; i++) dst[i] = (float) s[i];
+}
+
+struct whisper_aneforge_context * whisper_aneforge_init(const char * bundle_dir) {
+    const char * dylib = getenv("ANEFORGE_DYLIB");
+    if (!dylib) { fprintf(stderr, "aneforge: set ANEFORGE_DYLIB to libane_e5rt_dispatch.dylib\n"); return nullptr; }
```
danbev: Please split this into multiple lines (the body of the if statement). And also in all other places below to be consistent with the most of the rest of the code base.

src/aneforge/whisper-aneforge.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    auto ctx = new whisper_aneforge_context();
+    ctx->dl = dl;
+    auto compile  = (compile_fn) dlsym(dl, "ane_e5rt_program_compile");
+    ctx->set_input  = (set_in_fn)  dlsym(dl, "ane_e5rt_program_set_input_fp16");
+    ctx->get_output = (get_out_fn) dlsym(dl, "ane_e5rt_program_get_output_fp16");
+    ctx->execute    = (exec_fn)    dlsym(dl, "ane_e5rt_program_execute");
+    ctx->release    = (release_fn) dlsym(dl, "ane_e5rt_program_release");
+    if (!compile || !ctx->set_input || !ctx->get_output || !ctx->execute || !ctx->release) {
+        fprintf(stderr, "aneforge: missing dispatch symbols\n"); whisper_aneforge_free(ctx); return nullptr;
```
danbev: Please split this so that there is one statement per line for consistency.

src/aneforge/whisper-aneforge.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+struct whisper_aneforge_context * whisper_aneforge_init(const char * bundle_dir) {
+    const char * dylib = getenv("ANEFORGE_DYLIB");
+    if (!dylib) { fprintf(stderr, "aneforge: set ANEFORGE_DYLIB to libane_e5rt_dispatch.dylib\n"); return nullptr; }
+    void * dl = dlopen(dylib, RTLD_NOW | RTLD_LOCAL);
+    if (!dl) { fprintf(stderr, "aneforge: dlopen(%s) failed: %s\n", dylib, dlerror()); return nullptr; }
+
+    auto ctx = new whisper_aneforge_context();
+    ctx->dl = dl;
+    auto compile  = (compile_fn) dlsym(dl, "ane_e5rt_program_compile");
```
danbev: Nit: fix indentation to line up with lines below.

danbev:
> Is there anything else you want addressed before this can be considered?

Sorry about the delay, I'm taking a closer look at this now. Could you rebase and address the conflicts? 

I've converted the model according to the instructions provided above and uploaded it to [huggingface](https://huggingface.co/danbev/whisper-ane), in case other want to try this out.

src/aneforge/whisper-aneforge.cpp
sbryngelson: Done - split every compound if-body onto one statement per line (here and the others below), to match the codebase style.

src/aneforge/whisper-aneforge.cpp
sbryngelson: Done - one statement per line here too.

src/aneforge/whisper-aneforge.cpp
sbryngelson: Fixed - aligned the `=` across the dlsym block.

sbryngelson:
Thanks for the review @danbev. Pushed the formatting fixes: split all the compound if-bodies to one statement per line and aligned the dlsym assignment block. Ready for another look when you have a moment.

sbryngelson:
Rebased onto master and resolved the conflicts. They were with the VitisAI encoder backend that landed in the meantime, which touches the same three spots (`whisper_encode_external`, the external-encode dispatch, and state init) — I kept both, so VitisAI and ANEForge coexist: ANEForge is a runtime check (`ANEFORGE_ENCODER`), the compile-time CoreML/VitisAI/OpenVINO chain stays as the fallback. Verified the `whisper` target builds clean. And thanks for converting the model and posting it to HF.

sbryngelson:
Thanks for approving the runs - they caught a real portability bug. The Windows and vad (Linux) jobs failed because `whisper-aneforge.cpp` used `<dlfcn.h>` and `__fp16` unconditionally, which don't exist on Windows/MSVC and non-ARM Linux. Fixed in fbda7e6: the implementation is now guarded to `__APPLE__ && __aarch64__` with no-op stubs elsewhere (the backend is Apple-Neural-Engine-only anyway), so the whisper library links on every platform. Verified both the real and stub paths compile. Needs another CI approval when you have a moment.

sbryngelson:
re-requesting review @danbev after some time waiting on this. Cheers.

danbev:
> re-requesting review @danbev after some time waiting on this. Cheers.

Sorry about the delay/forgetting about this. I'll try to take another look today or tomorrow :+1:

danbev (approved):

sbryngelson:
many thanks @danbev ... do need someone to merge

PR #3910, opened 2026-06-25
whisper : map token timestamps to original time when VAD is enabled

buxuku:
When VAD is enabled, the segment getters (`whisper_full_get_segment_t0/t1`) already map timestamps back to the original audio timeline, but the per-token timestamps in `whisper_full_get_token_data()` stay in the VAD-processed timeline with the silence removed. So if you build word-level timing on top of the token times while VAD is on, the words drift off by however much silence VAD stripped out, and there's no public getter that applies the mapping.

This adds `whisper_full_get_token_t0/t1` (plus the `_from_state` variants) that map the token times back. A token inside a speech segment is interpolated within that segment; a token that falls in the silence removed between two segments is snapped to the nearest boundary, so it doesn't end up in the middle of a gap that isn't in the original audio. With VAD off, or when there's no segment info, the stored token times are returned unchanged, so existing callers aren't affected.

I hit this doing word-level re-segmentation with VAD enabled: the segment times lined up with the original audio but the token times didn't. Also extended tests/test-vad-full.cpp to exercise the new getters. Built and ran it on macOS.

### DIFF
--- a/include/whisper.h
+++ b/include/whisper.h
@@ -667,6 +667,16 @@ extern "C" {
     WHISPER_API whisper_token_data whisper_full_get_token_data           (struct whisper_context * ctx, int i_segment, int i_token);
     WHISPER_API whisper_token_data whisper_full_get_token_data_from_state(struct whisper_state * state, int i_segment, int i_token);
 
+    // Get the start/end time of the specified token, in centiseconds. When VAD is enabled
+    // these are mapped back to the original audio timeline (a token landing in a removed
+    // inter-segment silence snaps to the nearest speech boundary), unlike
+    // whisper_full_get_token_data().t0/t1 which stay in VAD-processed time. Without VAD the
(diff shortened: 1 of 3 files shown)


danbev (approved):
Optional, but perhaps we could extend [test-vad-full.cpp](https://github.com/ggml-org/whisper.cpp/blob/43d78af5be58f41d6ffbc227d608f104577741ea/tests/test-vad-full.cpp)  to exercise these new functions.

buxuku:
@danbev pushed an update on top of the approved version: the token times are now mapped segment by segment instead of one interpolation over the whole mapping table, and a token that lands in a removed silence snaps to the nearest speech boundary rather than somewhere in the middle of a gap that isn't in the original audio. Also extended test-vad-full.cpp to cover the new getters as you suggested. PTAL when you have a moment, thanks!

Issue #3819, opened 2026-05-22
server: /inference multipart upload fails when ffmpeg conversion is disabled because read_audio_data() is called with raw request bytes


jamesachurchill:
**Summary**

examples/server/server.cpp fails to process multipart uploads on /inference when the ffmpeg conversion path is not used.

In the non-converter branch, the server calls:
read_audio_data(audio_file.content, pcmf32, pcmf32s, params.diarize)
but read_audio_data(...) expects a filename/path, not raw multipart file bytes. The result is a 400 response with {"error":"failed to read audio data"} even for valid WAV uploads that work fine with whisper-cli.

**Reproduction**

Using the request shape documented in examples/server/README.md:
curl 127.0.0.1:8080/inference \
  -H "Content-Type: multipart/form-data" \
  -F file="@sample.wav" \
  -F response_format="json"

Observed result:
• HTTP 400
• response body: {"error":"failed to read audio data"}

The same audio file is valid and transcribes correctly through whisper-cli.

**Root cause**

Current logic treats the two branches differently:

• if ffmpeg conversion is enabled, the upload is written to a temp file first and then decoded from that temp path
• if ffmpeg conversion is not enabled, the code passes audio_file.content directly to read_audio_data(...)

That direct call is wrong for multipart uploads because audio_file.content is the uploaded file bytes, not a filesystem path.

**Expected behavior**

Multipart uploads to /inference should work regardless of whether ffmpeg conversion is enabled.

**Suggested fix**

Always write the uploaded multipart audio to a temp file and then call read_audio_data(temp_filename, ...) on that path, regardless of whether ffmpeg conversion is enabled.

Also use binary .write(...) rather than stream insertion when writing the temp file, e.g.:
std::ofstream temp_file{temp_filename, std::ios::binary};
temp_file.write(audio_file.content.data(), static_cast<std::streamsize>(audio_file.content.size()));

Then:
• if conversion is enabled, run ffmpeg against the temp file
• in both branches, decode from the temp file path
• delete the temp file afterward

**Local validation**

I validated this locally with a small patch that:
• always wrote the multipart upload to a temp file first
• used binary .write(...) instead of stream insertion
• always called read_audio_data(temp_filename, ...) instead of passing raw multipart bytes directly

After that change:
• multipart WAV uploads to /inference succeeded
• end-to-end curl transcription requests returned correct results
• the ffmpeg conversion path continued to work for non-WAV uploads

jamesachurchill:
[...]

PR #3911, opened 2026-06-25
tests: cover audio buffer decoding

jamesachurchill:
Adds a small regression test for the in-memory audio decoding path used by server multipart uploads without ffmpeg conversion.

The test reads samples/jfk.wav into a byte buffer, decodes it through read_audio_data(const char *, size_t, ...), decodes the same sample through the existing filename path, and asserts the PCM outputs match.

This covers the core regression behind #3819 without requiring a server lifecycle, network harness, ffmpeg, or model download.

Local verification on aarch64:
- cmake --build build-3819-test --target test-read-audio-data-buffer -j2
- ctest --test-dir build-3819-test -R test-read-audio-data-buffer --output-on-failure

Result: 100% tests passed.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -96,6 +96,16 @@ target_link_libraries(${UTF8_TEST} PRIVATE common)
 add_test(NAME ${UTF8_TEST} COMMAND ${UTF8_TEST})
 set_tests_properties(${UTF8_TEST} PROPERTIES LABELS "unit")
 
+# Audio byte-buffer decode regression test
+set(AUDIO_BUFFER_TEST test-read-audio-data-buffer)
+add_executable(${AUDIO_BUFFER_TEST} ${AUDIO_BUFFER_TEST}.cpp)
+target_include_directories(${AUDIO_BUFFER_TEST} PRIVATE ../examples)
+target_link_libraries(${AUDIO_BUFFER_TEST} PRIVATE common)
+target_compile_definitions(${AUDIO_BUFFER_TEST} PRIVATE
+    SAMPLE_PATH="${PROJECT_SOURCE_DIR}/samples/jfk.wav")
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -96,6 +96,16 @@ target_link_libraries(${UTF8_TEST} PRIVATE common)
...
 
+# Audio byte-buffer decode regression test
+set(AUDIO_BUFFER_TEST test-read-audio-data-buffer)
+add_executable(${AUDIO_BUFFER_TEST} ${AUDIO_BUFFER_TEST}.cpp)
+target_include_directories(${AUDIO_BUFFER_TEST} PRIVATE ../examples)
+target_link_libraries(${AUDIO_BUFFER_TEST} PRIVATE common)
+target_compile_definitions(${AUDIO_BUFFER_TEST} PRIVATE
+    SAMPLE_PATH="${PROJECT_SOURCE_DIR}/samples/jfk.wav")
+add_test(NAME ${AUDIO_BUFFER_TEST} COMMAND ${AUDIO_BUFFER_TEST})
+set_tests_properties(${AUDIO_BUFFER_TEST} PROPERTIES LABELS "unit")
```
danbev, line 107: Should this be run in CI? (it would need to have the `gh` label if so):
```suggestion
set_tests_properties(${AUDIO_BUFFER_TEST} PROPERTIES LABELS "unit;gh")
```
With that change it should have both labels:
```console
$ cmake --workflow --preset cpu-debug
$ ctest --test-dir build-cpu-debug/ -R test-read-audio-data-buffer --output-on-failure --print-labels
Internal ctest changing into directory: /home/danbev/work/ai/whisper-work/build-cpu-debug
Test project /home/danbev/work/ai/whisper-work/build-cpu-debug
All Labels:
  gh
  unit
```

danbev (approved):

Issue #1070, opened 2023-06-30
Android Inference is too slow


saffie91:
Hi, trying out the Android Implementation, it works well overall and thank you for this contribution.

However the inference time is incredibly slow, even for a few words it takes about 30 seconds to get a response which makes it unusable.

I am using the small model (I feel the accuracy of the tiny and base model are not up to par in different languages which is the use case in our scenario).

Is there any way to speed up the inference time? Is GPU inference possible on android for example?

For reference, I am using these parameters:

params.print_progress   = false;
    params.print_special    = false;
    params.print_realtime   = false;
    params.print_timestamps = false;
    params.translate        = false;
    params.single_segment   = false;
    params.max_tokens       = 32;
    params.language         = "auto";
    params.n_threads        = std::max(1, std::min(8, (int32_t) std::thread::hardware_concurrency()));
    params.audio_ctx        = 768;
    params.speed_up         = false;
//    params.temperature_inc  = params.temperature_inc;
//    params.prompt_tokens    = nullptr;
    params.prompt_n_tokens  = 0;

Azeirah:
GPU inference should be possible with CLBLAST, no? Might try that.

GersomR-afk:
Are you running the app in release mode? It improves the performance significantly.

sandorkonya:
@Azeirah  if someone is interested:
[Whisper GPU support via CLBLAST](https://github.com/ggerganov/whisper.cpp#opencl-gpu-support-via-clblast)
[CLBLAST on Android](https://github.com/CNugteren/CLBlast/blob/master/doc/installation.md#android-compilation-from-source)

sandorkonya:
FYI: maybe [this helps](https://github.com/ggerganov/llama.cpp/issues/2169).

ningpengtao-coder:
I try to compile clblast as a shared object of the android platform, use -DGGML_USE_CLBLAST to enable support for clblast, and load opencl through System.load("/vendor/lib64/libOpenCL.so"). Then run inference. Incorrect results and longer inference time.
whisper1: correct result.
whisper2: incorrect result.
![whisper2](https://github.com/ggerganov/whisper.cpp/assets/7344969/60915b28-fbb0-45ef-9e31-e2e00be1e4e9)
![whisper1](https://github.com/ggerganov/whisper.cpp/assets/7344969/0ec3cd46-9a70-4fd5-9107-6009d27bedc9)

ValleZ:
[...]

PR #3913, opened 2026-06-26
examples : Improved inference performance of Android example project

yymin1022:
Hi, there !

I found that AGP Debug build of `whisper.android` project has critical performance issue.<br/>
So I applied a simple tricky fix to this issue, which is also mentioned from comment of #1070.

Same fix is already applied to [llama.cpp](https://github.com/ggml-org/llama.cpp/pull/5123). 

This PR just applies `CMAKE_BUILD_TYPE=Release` to NDK Build arguments of each `whisper.android` and `whisper.android,java` example project.

### DIFF
--- a/examples/whisper.android.java/app/build.gradle
+++ b/examples/whisper.android.java/app/build.gradle
@@ -25,6 +25,18 @@ android {
   }
 
   buildTypes {
+    debug {
+      externalNativeBuild {
+        cmake {
+          // When the AGP build type is Debug, the native (NDK/CMake) build does not run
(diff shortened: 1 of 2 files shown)


yymin1022:
I've updated commits about CI failure.

yymin1022:
I've updated commits about CI failure again.

danbev (approved):

danbev:
I'll take a closer look at the failing CI build. Opened https://github.com/ggml-org/whisper.cpp/pull/3920 to address this.

@yymin1022 Thanks for this!

PR #3915, opened 2026-06-27
whisper : make voice_length() utf-8 aware for CJK

buxuku:
voice_length() estimates how long a token's text takes to say, and that estimate decides how a segment's time gets divided across its tokens. The loop walked the string one byte at a time, so every CJK character (3 bytes in UTF-8) counted as three units instead of one, and full-width punctuation never matched the ASCII cases. For Chinese/Japanese that skews the per-token timestamps.

This decodes one UTF-8 code point per step and treats full-width ，。！？ etc. the same as their ASCII forms. ASCII-only text decodes to the same weights as before.

For example "中文，" scored 9.0 before (9 bytes), and 4.0 now (中=1, 文=1, ，=2), the same as "ab,".

Came up while doing word-level subtitle timing for Chinese. Built on macOS and the existing tests/test-vad-full still passes.

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -8416,24 +8416,35 @@ static float voice_length(const std::string & text) {
         if (c < 0x80) {
             len = 1;
         } else if ((c >> 5) == 0x6) {
-            cp = c & 0x1F; len = 2;
+            cp = c & 0x1F;
+            len = 2;
         } else if ((c >> 4) == 0xE) {
-            cp = c & 0x0F; len = 3;
+            cp = c & 0x0F;
+            len = 3;
         } else if ((c >> 3) == 0x1E) {
-            cp = c & 0x07; len = 4;
+            cp = c & 0x07;
+            len = 4;
         } else {
-            cp = c; len = 1; // stray continuation / invalid lead byte
+            cp = c; // stray continuation / invalid lead byte
+            len = 1;
         }
         if (i + (size_t) len <= n) {
             bool ok = true;
             for (int k = 1; k < len; ++k) {
                 const unsigned char cc = s[i + k];
-                if ((cc & 0xC0) != 0x80) { ok = false; break; }
+                if ((cc & 0xC0) != 0x80) {
+                    ok = false;
+                    break;
+                }
                 cp = (cp << 6) | (cc & 0x3F);
             }
-            if (!ok) { cp = c; len = 1; }
+            if (!ok) {
+                cp = c;
+                len = 1;
+            }
         } else {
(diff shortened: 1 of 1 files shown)


src/whisper.cpp
```diff
@@ -8397,24 +8397,75 @@ static int64_t sample_to_timestamp(int i_sample) {
...
+    const size_t n = text.size();
+
+    for (size_t i = 0; i < n; ) {
+        const unsigned char c = s[i];
+        uint32_t cp = c;
+        int len = 1;
+        if (c < 0x80) {
+            len = 1;
+        } else if ((c >> 5) == 0x6) {
+            cp = c & 0x1F; len = 2;
```
danbev: Nit: Pehaps just one statement per line here to be consistent with other parts of the codebase:
```suggestion
            cp = c & 0x1F; 
            len = 2;
```
And likewise for the lines below.

src/whisper.cpp
```diff
@@ -8397,24 +8397,75 @@ static int64_t sample_to_timestamp(int i_sample) {
...
+        } else if ((c >> 3) == 0x1E) {
+            cp = c & 0x07; len = 4;
+        } else {
+            cp = c; len = 1; // stray continuation / invalid lead byte
+        }
+        if (i + (size_t) len <= n) {
+            bool ok = true;
+            for (int k = 1; k < len; ++k) {
+                const unsigned char cc = s[i + k];
+                if ((cc & 0xC0) != 0x80) { ok = false; break; }
```
danbev: Nit: Place if body on a new line and separate the statements.

src/whisper.cpp
```diff
@@ -8397,24 +8397,75 @@ static int64_t sample_to_timestamp(int i_sample) {
...
+                cp = (cp << 6) | (cc & 0x3F);
+            }
+            if (!ok) { cp = c; len = 1; }
```
danbev: Nit: If body on a new line and separate statements.

danbev (approved):
Just note that `test-vad-full` does not currently invoke `voice_length`. 

I manually verified using lldb using a japanese audio sample and setting the language (`-l ja`).

src/whisper.cpp
buxuku: Addressed the nits, PTAL

src/whisper.cpp
buxuku: Addressed the nits, PTAL

src/whisper.cpp
buxuku: Addressed the nits, PTAL

PR #3916, opened 2026-06-27
whisper : expose internal VAD speech segments

buxuku:
When transcribing with vad = true, whisper already detects the speech segments internally and keeps them in the state, but there's no way to read them back. This exposes them through whisper_full_n_vad_segments() and whisper_full_get_vad_segment_t0/t1() (with _from_state variants). The times are on the original audio timeline in centiseconds, and the count is 0 when VAD wasn't used.

The point is to let callers reuse whisper's own speech boundaries — for instance to clip or align subtitles to speech — instead of running a separate VAD pass over the same audio.

Extended tests/test-vad-full.cpp to check the segments come back non-empty and in order. Built and ran it on macOS.

### DIFF
--- a/include/whisper.h
+++ b/include/whisper.h
@@ -681,6 +681,17 @@ extern "C" {
     WHISPER_API float whisper_full_get_token_p           (struct whisper_context * ctx, int i_segment, int i_token);
     WHISPER_API float whisper_full_get_token_p_from_state(struct whisper_state * state, int i_segment, int i_token);
 
+    // Access the speech segments detected by the internal VAD (only when params.vad = true).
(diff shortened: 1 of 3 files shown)


danbev:
@buxuku Could you take a look at the conflicts and resolve them. 
And if you rebase or merge upstream/master the two failing CI runs should be fixed now. Thanks!

buxuku:
> @buxuku Could you take a look at the conflicts and resolve them. And if you rebase or merge upstream/master the two failing CI runs should be fixed now. Thanks!

Done! Conflicts were just my stuff sitting right next to the #3910 token_t0/t1 changes, so I kept both. Rebased on master, should be happy now

danbev (approved):

PR #3921, opened 2026-07-01
Fix #587: Remove leading space from txt output

### LINKED ISSUE #587: Each sentence has a leading space when outputting to .txt file
**What's wrong:**
Each line has a prepended space before the line begins.

**What I want:**
I want each line to begin with no extra spaces.

**Supporting details:**
Here's what I see in my resulting text file.  Note the leading space on each line:

 Do you have any comments?
 I think it's really funny, and we probably still do similar things today.
 I don't actually remember this particular event, but...
 I think I was in that one.
 But yeah, it's really funny, and I think it's funny because you did, or Jeffrey did things like that while we were there.
 I remember he licked someone's toast before giving it to them.
 I think that was mine.
 Yeah, and we thought it was really gross, but yeah.
 I...
 [silence]
 [background noise]

[holly_responds.mp3.wav.txt](https://github.com/ggerganov/whisper.cpp/files/10927243/holly_responds.mp3.wav.txt)

JingliangGao:
The BPE tokenizer used by Whisper produces tokens with leading spaces, causing each line in the txt output to start with an unwanted space.

This fix strips leading whitespace (spaces and tabs) from each segment when writing to txt output files, improving the readability of the transcription output.

Fixes: https://github.com/ggml-org/whisper.cpp/issues/587

### DIFF
--- a/examples/cli/cli.cpp
+++ b/examples/cli/cli.cpp
@@ -465,7 +465,16 @@ static void output_txt(struct whisper_context * ctx, std::ofstream & fout, const
             speaker = estimate_diarization_speaker(pcmf32s, t0, t1);
         }
 
-        fout << speaker << text << "\n";
+        // strip leading whitespace from each segment (#587)
+        if (!speaker.empty()) {
+            fout << speaker << text;
+        } else {
+            while (*text == ' ' || *text == '\t') {
+                text++;
+            }
+            fout << text;
+        }
+        fout << "\n";
     }
 }
 
--- a/examples/cli/cli.cpp
+++ b/examples/cli/cli.cpp
@@ -465,7 +465,6 @@ static void output_txt(struct whisper_context * ctx, std::ofstream & fout, const
             speaker = estimate_diarization_speaker(pcmf32s, t0, t1);
         }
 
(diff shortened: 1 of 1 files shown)


examples/cli/cli.cpp
```diff
@@ -465,7 +465,16 @@ static void output_txt(struct whisper_context * ctx, std::ofstream & fout, const
             speaker = estimate_diarization_speaker(pcmf32s, t0, t1);
         }
 
-        fout << speaker << text << "\n";
+        // strip leading whitespace from each segment (#587)
```
danbev: Nit: I think we can remove this comment, and no need to link to the issue.

danbev (approved):

examples/cli/cli.cpp
JingliangGao: Oh, yes!  I added this comment before to remind me of the reason for the modification here. I will remove this comment and modify the PR. Thanks

JingliangGao:
Hi @ggerganov and @danbev, this PR has been approved and all CI checks are passing. Would you mind taking a look and then merge it when you have a moment? Thanks! :)

Issue #3924, opened 2026-07-03
Reachable assertion / abort in whisper_model_load() on an invalid ftype


professor-moody:
`whisper_model_load()` (src/whisper.cpp) passes a model-declared `ftype` to `ggml_ftype_to_ggml_type()` (ggml/src/ggml.c:1410). An invalid `ftype` (e.g. 5, 6, or ≥27 after `%1000`) sets `wtype = GGML_TYPE_COUNT` as an error sentinel, then a `GGML_ASSERT` immediately fires on that value and calls `ggml_abort()` — the caller's error handler is dead code.

`GGML_ASSERT` is not compiled out under `-DNDEBUG`, so this aborts release builds as well.

**Confirmed on:** commit `95ea8f9b`, x86_64. 48-byte PoC available on request.

**Suggested fix:** return a load error on an unrecognized `ftype` instead of asserting.

PR #3929, opened 2026-07-06
Fixes issue: #3924

michaelsam94:
Problem: invalid model ftype values reached ggml_ftype_to_ggml_type() and triggered GGML_ASSERT, aborting instead of returning a load error.

Change: validate normalized hparams.ftype in whisper_model_load() before calling ggml_ftype_to_ggml_type().

Test: added test-model-load, which writes a minimal invalid model header with ftype = 5 and expects whisper_init_from_file_with_params() to fail cleanly.

Verification I ran: git diff --check; direct compiled and ran test-model-load.

CMake/CTest note: not run because cmake was unavailable in this environment.

### DIFF
--- a/tests/test-model-load.cpp
+++ b/tests/test-model-load.cpp
@@ -1,0 +1,53 @@
+#include "ggml.h"
+#include "whisper.h"
+
+#include <cstdint>
+#include <cstdio>
+#include <fstream>
+
+static void write_u32(std::ofstream & fout, uint32_t value) {
+    fout.write(reinterpret_cast<const char *>(&value), sizeof(value));
+}
+
+static void write_i32(std::ofstream & fout, int32_t value) {
+    fout.write(reinterpret_cast<const char *>(&value), sizeof(value));
+}
+
+int main() {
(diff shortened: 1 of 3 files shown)


tests/test-model-load.cpp
```diff
@@ -0,0 +1,53 @@
...
+        write_i32(fout, 80);   // n_mels
+        write_i32(fout, 5);    // invalid ftype
+
+        if (!fout) {
+            return 2;
+        }
+    }
+
+    whisper_context_params params = whisper_context_default_params();
+    whisper_context * ctx = whisper_init_from_file_with_params(path, params);
```
danbev, line 44: Perhaps we can add `std::remove(path)` directly after `whisper_init_from_file_with_params`, so that we just have one std::remove instead of one in both code paths?

danbev (approved):
Thanks for the PR!

I think that we might need something similar in parakeet.cpp and perhaps we could add a function like `ggml_ftype_is_supported` to ggml in the future that both could use. This would avoid having to update this code when new types are added to ggml. But that is outside of the scope of this PR.

PR #3931, opened 2026-07-06
ruby : add VAD speech segments API

KitaitiMakoto:
Hi,

This pull request adds Ruby binding for VAD speech segments API introduced by #3916, and includes some refactoring.

Thank you.

### DIFF
--- a/bindings/ruby/lib/whisper/log_settable.rb
+++ b/bindings/ruby/lib/whisper/log_settable.rb
@@ -1,13 +1,5 @@
-require "mutex_m"
-
 module Whisper
   module LogSettable
-    class << self
-      def extended(base)
-        base.extend Mutex_m
-      end
-    end
-
     private
 
     def start_log_callback_thread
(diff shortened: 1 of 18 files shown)


KitaitiMakoto:
I cancelled workflows which aren't related to Ruby binding.

bindings/ruby/lib/whisper/log_settable.rb
```diff
@@ -32,5 +24,10 @@ def start_log_callback_thread
         end
       }
     end
+
+    def synchronize(&block)
+      @mutex ||= Thread::Mutex.new
+      @mutex.synchronize &block
```
danbev: Nit: There is an extra whitespace at the end of this line:
```suggestion
      @mutex.synchronize &block
````

danbev (approved):

bindings/ruby/lib/whisper/log_settable.rb
KitaitiMakoto: Wow, thanks for finding out.

Issue #3888, opened 2026-06-17
Language auto-detect in `whisper_full` ignores `encoder_begin_callback` and `abort_callback`


sindresorhus:
### Summary

When `whisper_full` / `whisper_full_with_state` runs language auto-detection (i.e. `params.language` is `NULL`, `""`, or `"auto"`, or `params.detect_language` is set), the encode/decode it performs do **not** honor the `whisper_full_params` callbacks `encoder_begin_callback` and `abort_callback`. The main transcription loop honors both. As a result, the GPU work done during language auto-detection is invisible to and uninterceptable by the caller.

### Details

Verified against  `v1.8.4`, commit fc67457. Line numbers are from `src/whisper.cpp` at that commit.

The auto-detect branch in `whisper_full_with_state` calls `whisper_lang_auto_detect_with_state` directly, with no `encoder_begin_callback` beforehand:

```c
// whisper_full_with_state, ~line 6811
if (params.language == nullptr || strlen(params.language) == 0 || strcmp(params.language, "auto") == 0 || params.detect_language) {
    std::vector<float> probs(whisper_lang_max_id() + 1, 0.0f);

    const auto lang_id = whisper_lang_auto_detect_with_state(ctx, state, 0, params.n_threads, probs.data()); // <- runs encode + decode
    ...
}
```

`whisper_lang_auto_detect_with_state` (~line 4021) runs the encoder and a decode step via the public helpers:

```c
// ~line 4040
if (whisper_encode_with_state(ctx, state, seek, n_threads) != 0) { ... }
// ~line 4047
if (whisper_decode_with_state(ctx, state, prompt.data(), prompt.size(), 0, n_threads) != 0) { ... }
```

Both of those helpers hardcode `nullptr` for the abort callback:

```c
// ~line 3917
int whisper_encode_with_state(struct whisper_context * ctx, struct whisper_state * state, int offset, int n_threads) {
    if (!whisper_encode_internal(*ctx, *state, offset, n_threads, nullptr, nullptr)) { ... }
}

// ~line 3935
int whisper_decode_with_state(struct whisper_context * ctx, struct whisper_state * state, const whisper_token * tokens, int n_tokens, int n_past, int n_threads) {
    ...
    if (!whisper_decode_internal(*ctx, *state, state->batch, n_threads, false, nullptr, nullptr)) { ... }
}
```

[...]

PR #3936, opened 2026-07-08
whisper : run encoder_begin_callback on the language auto-detect path

javierdejesusda:
encoder_begin_callback is documented as a hook to abort before the encoder runs, but on the language auto-detection path it never gets consulted. When language is null/empty/"auto" or detect_language is set, whisper_full_with_state calls whisper_lang_auto_detect_with_state, which runs the encoder without ever checking the callback. So a caller that returns false to cancel gets ignored during auto-detect and the encode proceeds anyway.

This invokes encoder_begin_callback (when set) right before whisper_lang_auto_detect_with_state. If it returns false, we log and return -3, the same abort behavior already used for the encoder call later in the function. The non-auto-detect paths are unchanged.

Verified against current master that the auto-detect path previously reached the encoder with no callback consultation, and that the callback now fires first and aborts cleanly when it returns false. No automated test added since it is a small control-flow guard.

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -6833,6 +6833,13 @@ int whisper_full_with_state(
     if (params.language == nullptr || strlen(params.language) == 0 || strcmp(params.language, "auto") == 0 || params.detect_language) {
         std::vector<float> probs(whisper_lang_max_id() + 1, 0.0f);
 
+        if (params.encoder_begin_callback) {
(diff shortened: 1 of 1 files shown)


danbev (approved):
Just to note that this does not fully resolve #3888, that would require to also add the abort_callback and user data to be passed down to the encode and decode functions.

PR #3948, opened 2026-07-21
parakeet : fix duration argmax underflow to index 0

### LINKED ISSUE #3932: parakeet: duration argmax runs on log-probs that underflow to -inf, silently picking duration 0
I've been porting parakeet-tdt to another runtime and comparing against parakeet-cli token by token, which turned up an odd corner in the decoder.

`parakeet_decode` picks the TDT duration by argmax over `pstate.logits[n_vocab_logits + i]`, but `pstate.logits` holds log-probs — the joint graph ends in `log(softmax(...))` over the whole 1030-wide row. The duration slots carry a vanishing share of that softmax (their log-probs sit around -100 and below on real speech), so `exp(logit - max)` underflows fp32 to exactly 0 and the log gives -inf. When all five duration slots are -inf, the argmax loop (`best_duration_logit = -1e10f` with a strict `>`) never updates and silently returns duration index 0 — a value that came from the sentinel, not the model.

On samples/gb1.wav this happens at 16 of 786 emitted tokens. Computing the same argmax analytically (logit − logsumexp, which can't underflow) gives a different winner at those steps with gaps of 6–20 nats, so it's not a near-tie. Token ids happen to come out the same on that clip, but the affected timestamps shift, and in principle the decode path can diverge since t advances differently.

Two easy fixes: argmax the durations over the raw `logits` tensor (it's already `ggml_set_output` in the joint graph), or compute the log-softmax as logit − logsumexp instead of log(softmax).

Solaris-star:
## Summary

Fixes #3932.

`parakeet_decode` picks TDT duration by argmax over `pstate.logits[n_vocab_logits + i]`, but `parakeet_joint` previously copied the final graph node — `log(softmax(...))` over the full vocab+duration row — into that buffer. Duration slots often carry vanishing mass after that full-row softmax, so their log-probs underflow fp32 to `-inf`. With every duration slot at `-inf`, the old argmax loop (`best_duration_logit = -1e10f` + strict `>`) never updates and silently returns duration index `0`.

On `samples/jfk.wav`-class clips this is not always fatal for token ids, but timestamps shift and the decode path can diverge because `t` advances differently.

## Changes

1. In `parakeet_joint`, read the named raw `"logits"` tensor (already `ggml_set_output`) into `pstate.logits` instead of the final log-probs node. Argmax on raw logits is equivalent for finite values and remains defined when log-probs are not.
2. Initialize duration argmax from the first duration logit (no sentinel) so a true all-equal case still returns a real index.
3. Because `pstate.logits` is now raw logits, recompute token `p` / `plog` in `create_token_data` with a stable log-sum-exp softmax over the token/blank slice. Previously `plog` was assigned the raw log-prob value from the joint buffer; after this change that would have been wrong without the recompute.

## Test plan

- [x] Confirmed no competing open PR for #3932
- [x] Diff limited to `src/parakeet.cpp`
- [ ] Build with Parakeet enabled and run the existing `parakeet-cli` decode path on `samples/gb1.wav` (runtime validation; not run in this development environment because CMake and a downloaded Parakeet model are unavailable here)
- [ ] Compare duration indices vs analytical `logit - logsumexp` argmax on a clip that previously produced `-inf` duration slots

### DIFF
--- a/src/parakeet.cpp
+++ b/src/parakeet.cpp
@@ -2341,7 +2341,15 @@ static bool parakeet_joint(
             return false;
         }
 
-        logits = ggml_graph_node(gf, -1);
+        // Prefer the raw joint logits (named "logits") over the final log(softmax)
+        // graph node. Duration slots often carry vanishing mass after the full-row
+        // softmax, so log(softmax) underflows those slots to -inf and the duration
+        // argmax silently returns index 0 (see #3932). Argmax on raw logits is
+        // equivalent for finite values and remains defined when log-probs are not.
+        logits = ggml_graph_get_tensor(gf, "logits");
+        if (logits == nullptr) {
+            logits = ggml_graph_node(gf, -1);
+        }
 
         if (!ggml_graph_compute_helper(sched, gf, n_threads)) {
             return false;
@@ -2428,11 +2436,20 @@ static parakeet_token_data create_token_data(
                         float   token_logit,
                           int   n_vocab_logits) {
 
+    // pstate.logits holds raw joint logits (see parakeet_joint). Compute a stable
+    // softmax over the token/blank slice for p and plog.
+    float max_logit = token_logit;
+    for (int i = 0; i < n_vocab_logits; ++i) {
+        if (pstate.logits[i] > max_logit) {
+            max_logit = pstate.logits[i];
+        }
+    }
     float token_sum = 0.0f;
     for (int i = 0; i < n_vocab_logits; ++i) {
-        token_sum += expf(pstate.logits[i]);
+        token_sum += expf(pstate.logits[i] - max_logit);
     }
-    float token_p = expf(token_logit) / token_sum;
+    const float log_z = max_logit + logf(token_sum);
+    const float token_p = expf(token_logit - log_z);
 
     parakeet_token_data token_data;
     token_data.id = token_id;
@@ -2440,7 +2457,7 @@ static parakeet_token_data create_token_data(
     token_data.duration_value = duration_value;
     token_data.frame_index = frame_index;
     token_data.p = token_p;
-    token_data.plog = token_logit;
+    token_data.plog = token_logit - log_z;
     token_data.t0 = frame_index * pctx.model.hparams.subsampling_factor;
     token_data.t1 = (frame_index + duration_value) * pctx.model.hparams.subsampling_factor;
     token_data.is_word_start = is_word_start_token(pctx.vocab, token_id);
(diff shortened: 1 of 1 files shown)


bhargav191098:
Looking up on the remaining unticked items on the Test Plan: 
>> Build with Parakeet enabled and run existing parakeet decode path on a real clip (no GB10/Parakeet model in this CI agent environment)

Think the gb1 wav file mentioned on the [issue](https://github.com/ggml-org/whisper.cpp/issues/3932) is based off of this - https://github.com/ggml-org/whisper.cpp/blob/080bbbe85230f624f0b52127f1ae1218247989f9/Makefile#L16

Did the following to get the gb1 wav to see the repro route.

```
wget -O samples/gb1.ogg \
  https://upload.wikimedia.org/wikipedia/commons/1/1f/George_W_Bush_Columbia_FINAL.ogg

ffmpeg -i samples/gb1.ogg \
  -ar 16000 -ac 1 -c:a pcm_s16le samples/gb1.wav
```
  
Will try to repro the exact thing later today/tomorrow. Posting the above step to obtain the wav file in case someone wants to do this a bit faster! 

@Solaris-star what do you mean by no parakeet model on the CI agent env? Parakeet model should be available right? Did you mean a test for gb1 using parakeet on CI?

Solaris-star:
You are right — my wording was imprecise. By “no Parakeet model in the CI agent environment” I meant this development environment, not whisper.cpp project CI. The Parakeet model is available, and your `gb1.wav` preparation plus `ggml-org/parakeet-GGUF` model is exactly the runtime validation this change needs.

I updated the PR test plan to say that explicitly. I could not run that validation here because this environment has neither CMake nor a downloaded Parakeet model; I did not mean to claim that project CI lacks the model. If the existing CI job has the model/runtime available, a `parakeet-cli` run on `samples/gb1.wav` with segment output would be the appropriate check.

jhen0409:
I've run this branch locally to cover the unticked runtime-validation items, it fixes the issue.

I hit the same underflow independently while trying to implement `parakeet-tdt_ctc-0.6b-ja` support (follow-up PR coming, maybe). On that model the impact is worse than timestamps: a duration logit often exceeds every token logit, so the full-row `log(softmax)` flushes all token log-probs to `-inf` and the token argmax degenerates too, transcripts come out wrong.

Validation (`parakeet-tdt-0.6b-v2-f16`, Metal, 27:50 flac (samples), `-ps`):

metric | before | this PR
-- | -- | --
joint evals | 14,289 | 9,088 (−36%)
decode time | 6,479 ms | 2,669 ms (−59%)
tokens with dur_idx=0 | 3,179 | 106
zero-width tokens (t0==t1) | 3,778 | 751

Word timestamps vs NeMo reference (`nemo_toolkit[asr]`, `timestamps=True`): word-end mean delta drops from 85 ms to 2 ms (5-min excerpt, 601 matched words); words with end >100 ms off drop from 1,302/3,640 (36%) to 18 (0.5%).

PR #3962, opened 2026-07-30
sync : ggml

ggerganov:


danbev (approved):

ggerganov:
@danbev There are a few failing tests in `ggml_vec_dot_f16`. Likely some FP16 overflow, though it's not clear why it starts happing after this sync.

danbev:
@ggerganov I can take a closer look tomorrow morning :+1:

ggerganov:
Thanks for taking a look.

Btw, we want to merge sync PRs with "rebase" instead of "squash" to preserve the commit history and git blame easier when needed for example.

danbev:
> Btw, we want to merge sync PRs with "rebase" instead of "squash" to preserve the commit history and git blame easier when needed for example.

Ah that make sense, I just did that by habit :disappointed:  Do we need to revert this?

ggerganov:
No need to revert.

PR #3963, opened 2026-07-31
Fix VAD min silence argument parsing

chupakobra6:
## Summary

Fix `--vad-min-silence-duration-ms` (`-vsd`) parsing in `whisper-vad-speech-segments`.

The option was incorrectly assigned to `vad_min_speech_duration_ms` instead of `vad_min_silence_duration_ms`. As a result, the requested silence duration was ignored and the minimum speech duration was overwritten.

## Testing

- Built `whisper-vad-speech-segments`
- Ran `test-vad`
- Verified long and short option forms produce the same result
- Tested with `jfk.wav` and a 1000 ms minimum silence duration


danbev (approved):

danbev:
I've opened https://github.com/ggml-org/whisper.cpp/pull/3966 for the failing `vad` CI job.

PR #3950, opened 2026-07-22
verify hparams loaded from parakeet model bin file

bhargav191098:
### PR details 

Fixes the concerns raised in issue [#3933](https://github.com/ggml-org/whisper.cpp/issues/3933) 

The issue is valid. This PR extends the param check to all of the Parakeet parameters. If these values are changed, they no longer represent the model as is, hence I go for a strict check to avoid any attacks via this route.

### Repro

I was able to modify the bin file's n_fft value and as the issue points, out I changed it to 0. Once I do that, this is the behavior on main: 

```
parakeet_model_load: arch                   = Parakeet TDT
parakeet_model_load: n_vocab                = 8192
parakeet_model_load: n_audio_ctx            = 5000
parakeet_model_load: n_audio_state          = 1024
parakeet_model_load: n_audio_head           = 8
parakeet_model_load: n_audio_layer          = 24
parakeet_model_load: n_mels                 = 128
parakeet_model_load: n_fft                  = 0
parakeet_model_load: eps                    = 0.000010
parakeet_model_load: ftype                  = 0
parakeet_model_load: qntvr                  = 0
parakeet_model_load: subsampling_factor     = 8
parakeet_model_load: n_subsampling_channels = 256
parakeet_model_load: n_conv_kernel          = 9
parakeet_model_load: n_pred_dim             = 640
parakeet_model_load: n_pred_layers          = 2
parakeet_model_load: n_tdt_durations        = 5
parakeet_model_load: n_max_tokens           = 10
parakeet_model_load: loaded window function with 400 samples

system_info: n_threads = 4 / 32 | PARAKEET : CUDA : ARCHS = 1200 | BLACKWELL_NATIVE_FP4 = 1 | CPU : SSE3 = 1 | SSSE3 = 1 | AVX = 1 | AVX2 = 1 | F16C = 1 | FMA = 1 | OPENMP = 1 | REPACK = 1 |

Processing file: .\samples\jfk.wav
read_audio_data: reading audio data from '.\samples\jfk.wav' ...
read_audio_data: trying to decode with miniaudio
EXITCODE=-1073741819
```

On a deployed system, if an attacker modifies the bin file, it can become a denial of service issue for an ASR service running on that system.

### Fixed behavior:

With the changes in this branch for the same bad model: 

```
parakeet_model_load: loading model
parakeet_validate_hparams: invalid Parakeet metadata: n_fft = 0, expected 512
parakeet_init_with_params_no_state: failed to load model
[...]

PR #3972, opened 2026-08-05
whisper : reject n_dims outside 1..4 in model load (#3944)

### LINKED ISSUE #3944: Stack buffer overflow in whisper_model_load: file-controlled n_dims written past ne[4]
Loading a crafted legacy-format ggml `.bin` model overflows a fixed 4-element stack array. In `src/whisper.cpp`, `whisper_model_load` reads a 32-bit `n_dims` from the file and then writes `n_dims` values into `ne[4]`, with no check that `n_dims <= 4`:

```cpp
int32_t n_dims;
read_safe(loader, n_dims);           // from the file, unchecked
...
int32_t ne[4] = { 1, 1, 1, 1 };      // fixed 4-element stack array
for (int i = 0; i < n_dims; ++i) {
    read_safe(loader, ne[i]);        // out-of-bounds stack write for i >= 4
    nelements *= ne[i];
}
```

Both `n_dims` and each `ne[i]` come from the file, so the attacker controls the overflow length and the written values. There is no `n_dims < 1 || n_dims > 4` guard; the shape/`nbytes` sanity checks run only after the overflow. The same pattern is at `whisper_vad_init_with_params`.

Reproduction: built `whisper-cli` from HEAD with AddressSanitizer and ran it on a minimal legacy `.bin` (valid magic, minimal hparams with `n_audio_state=1`, a 1-token vocab, one mel filter) whose first tensor record sets `n_dims`. Control (`n_dims=2`) loads fine; crafted (`n_dims=64`) reports:

```
AddressSanitizer: stack-buffer-overflow WRITE of size 4
  in read_safe<int>  (whisper.cpp:964)
  <- whisper_model_load  (whisper.cpp:1885)
```

Impact: a stack out-of-bounds write with attacker-controlled length and data while loading an untrusted whisper `.bin` (a routine download-a-model action) — a canary abort at least, up to saved-register/return-address overwrite depending on build. CWE-121.

Suggested fix: right after `read_safe(loader, n_dims)`, reject `n_dims < 1 || n_dims > 4`, at both `whisper_model_load` and `whisper_vad_init_with_params`.

Found with a structure-aware model-parser fuzzer.

ousamabenyounes:
## What

Reject an out-of-range `n_dims` when loading a model, so a crafted legacy `.bin`
can no longer overflow the fixed 4-element `ne[]` stack array in
`whisper_model_load` / `whisper_vad_init_with_params`.

## Why

Fixes #3944 (CWE-121).

Both functions read a file-controlled `int32 n_dims` and then read `n_dims`
values into `int32_t ne[4]` with no bounds check:

```cpp
read_safe(loader, n_dims);        // from the file, unchecked
...
int32_t ne[4] = { 1, 1, 1, 1 };
for (int i = 0; i < n_dims; ++i) {
    read_safe(loader, ne[i]);     // out-of-bounds stack write for i >= 4
}
```

Both `n_dims` and each `ne[i]` come from the file, so an attacker controls the
overflow length and the written values while loading an untrusted model — a
routine download-a-model action. The shape/size sanity checks run only *after*
the overflow.

The fix adds `n_dims < 1 || n_dims > 4` rejection right after `n_dims` is read,
at both sites, in the existing error style — exactly as suggested in the issue.

## Verification

Built in `ubuntu:24.04` (`cmake`, `g++`), `-DGGML_NATIVE=OFF`.

New `tests/test-model-load.cpp` loads a minimal crafted fixture
(`models/for-tests-ggml-whisper-bad-ndims.bin`, produced by the added
`models/generate-whisper-test-model.py`) whose first tensor declares
`n_dims=64`, and asserts the load returns `NULL` instead of crashing.

### RED → GREEN

**RED** — guard reverted, AddressSanitizer build, run on the crafted model:

```
==ERROR: AddressSanitizer: stack-buffer-overflow ... WRITE of size 4
    #6 read_safe<int>      src/whisper.cpp:964
    #7 whisper_model_load  src/whisper.cpp:1885
SUMMARY: AddressSanitizer: stack-buffer-overflow ... in memcpy
```

(matches the report in the issue exactly).

**GREEN** — with the fix:

```
whisper_model_load: invalid n_dims 64 in model file (expected 1..4)
whisper_init_with_params_no_state: failed to load model
test-model-load: crafted n_dims > 4 model rejected as expected   (exit 0)
```

The load now fails cleanly; no overflow. Follows the crafted-fixture test
pattern established by #3950.


ousamabenyounes:
Closing as superseded by #3957 (merged `df1547b6`).

That PR landed the identical fix for this same CWE-121 stack-buffer-overflow — an explicit `n_dims` bounds check right after the read, before the `for (i < n_dims)` loop that writes into the fixed `int32_t ne[4]`. It covers both sites this PR touches (`whisper_model_load` and `whisper_vad_init_with_params`) and additionally `parakeet_model_load`, which shares the same copy-pasted loader logic.

Since the guard is already in `master` at both of these locations, this PR now only conflicts by re-adding an already-present check, so there's nothing left for it to fix. The one cosmetic difference (`n_dims < 1` here vs `n_dims < 0` upstream) is not a security gap — `n_dims == 0` runs the read loop zero times and leaves `ne` at `{1,1,1,1}`, no overflow.

Thanks — closing in favor of #3957.

PR #3979, opened 2026-08-08
suggestion: add `--continue` flag to `wget` and `wget2`

OotsutsukiHagoromo:


models/download-ggml-model.sh
```diff
@@ -128,7 +128,7 @@ elif [ -x "$(command -v curl)" ]; then
          ${HF_TOKEN:+--header "Authorization: Bearer $HF_TOKEN"} \
          --output ggml-"$model".bin $src/$pfx-"$model".bin
 elif [ -x "$(command -v wget)" ]; then
-    wget --no-config --quiet --show-progress -O ggml-"$model".bin $src/$pfx-"$model".bin
+    wget --no-config --quiet --show-progress --continue -O ggml-"$model".bin $src/$pfx-"$model".bin
```
danbev, line 131: This script has a check earlier so this option will never actually never take effect: https://github.com/ggml-org/whisper.cpp/blob/233fe1fc9b48a09e361d3594520838ca266537fe/models/download-ggml-model.sh#L115-L118

PR #3981, opened 2026-08-09
whisper : default-initialize whisper_mel to avoid uninitialized read

### LINKED ISSUE #3978: whisper_full with n_samples == 0 runs on uninitialized whisper_mel fields (intermittent NULL deref in encoder)
## Summary

`whisper_full` / `whisper_full_with_state` called with `n_samples == 0` runs on **uninitialized** `whisper_mel` fields of a freshly allocated state. Depending on the heap garbage under those fields, the call either silently returns 0 segments or crashes the process with an access violation (read at NULL) inside `whisper_encode_internal`. The non-determinism made this very painful to diagnose in a downstream app — it presented as a rare, seemingly random `0xC0000005` under WER with a misleading `coreclr`-blamed fault bucket, and took two full minidumps to pin down.

## Mechanism (verified on tag v1.7.6; the same pattern is present on current master)

1. `whisper_full_with_state` computes the mel spectrogram **only when there is audio** — `if (n_samples > 0) { ... whisper_pcm_to_mel_with_state(...) ... }` (v1.7.6: `whisper.cpp:6806`). For `n_samples == 0` the mel is never touched.
2. `struct whisper_mel` carries **no initializers** on its int fields (v1.7.6: `whisper.cpp:447-453`), unlike its neighbours in `whisper_state` which are `= 0`-initialized. After `whisper_state * state = new whisper_state;` (v1.7.6: `whisper.cpp:3391`), `mel.n_len / n_len_org / n_mel` are heap garbage; `mel.data` is a correctly constructed **empty** vector (data pointer NULL).
3. `seek_end` is derived from that garbage: `whisper_n_len_from_state` returns `state->mel.n_len_org` (v1.7.6: `whisper.cpp:4185-4187`, used at `:6842`). If the garbage happens to be below the `delta_min` guard (`:6847-6851`), the call quietly returns 0 — which is why the bug is intermittent. Otherwise decoding proceeds.
4. `whisper_encode_internal` then copies the mel with the garbage dimensions: `dst[...] = mel_inp.data[j*mel_inp.n_len + i]` (v1.7.6: `whisper.cpp:2416`) with `mel_inp.data == nullptr` → read at address 0 → process death. The `assert` at `:2404` is compiled out in Release.

## Evidence

Two full Windows minidumps of a .NET app using whisper.cpp v1.7.6 (via a binding) on CPU. In both, the faulting instruction is the mel copy loop (`whisper.dll+0x29f92`, read at `0x0`), and the recovered `whisper_mel` fields are mutually inconsistent garbage over an empty vector:

- dump 1: `n_len = 400521354`, `n_len_org = 201`, `n_mel = 400521474`, `data = {NULL, NULL, NULL}`
- dump 2: `n_len = 4`, `n_len_org = 56`, `n_mel = 0x1d301098` (low half of a leftover heap pointer), `data = {NULL, NULL, NULL}`

(A real mel would satisfy `n_len == n_len_org + 3000` and `n_mel == 80` for the `base` model; neither dump does.) The state's zero-initialized timing counters were all zero, confirming a fresh state rather than corruption.

The zero-sample input came from a real user scenario: a push-to-talk recording stopped faster than the capture backend's first buffer period produces a WAV with a valid header and zero frames, and the (C#) binding turns an empty sample array into `samples = NULL, n_samples = 0` via `fixed` over an empty span.

## Repro

Call `whisper_full(ctx, params, NULL, 0)` (or any `n_samples == 0`) on a fresh context/state, ideally under MSan/page heap or with the allocator primed with nonzero garbage. Whether it returns 0 or crashes depends on the previous contents of the `whisper_state` allocation.

## Suggested fix (either or both)

- Early-return from `whisper_full_with_state` when `n_samples == 0` **and** no mel was previously set via `whisper_set_mel` (e.g. check `state->mel.data` / a validity flag), instead of trusting `mel.n_len_org`;
- and/or give `whisper_mel`'s fields default initializers (`int n_len = 0; ...`) so a never-computed mel reads as "0 frames" and takes the existing too-short path deterministically.

## Related, but distinct

#3923 / PR #3956 cover a heap over-read in `log_mel_spectrogram` for 1–200 samples — that path requires `n_samples > 0` and is fixed on master. This report is about the `n_samples == 0` branch that skips `log_mel_spectrogram` entirely; it is not covered by #3956.

ousamabenyounes:
## What

`whisper_full()` / `whisper_full_with_state()` called with `n_samples == 0` on a
freshly allocated state can dereference a NULL mel buffer inside the encoder, or
silently return 0 — the outcome depends on heap garbage, so the crash is
intermittent.

Give `struct whisper_mel` default member initializers so a never-computed mel
reads as "0 frames".

## Why

The mel spectrogram is only computed when there is audio:
`if (n_samples > 0) { ... whisper_pcm_to_mel_with_state(...) }`. For
`n_samples == 0` the mel is never touched. But `whisper_mel` had no member
initializers, and the state is allocated with `new whisper_state`, so
`mel.n_len / n_len_org / n_mel` were indeterminate heap garbage while `mel.data`
is a correctly-constructed empty vector (NULL data pointer).

`seek_end` is derived from `mel.n_len_org`. If the garbage lands below the
`delta_min` guard the call quietly returns 0; otherwise decoding proceeds and
`whisper_encode_internal` copies the mel with garbage dimensions from the empty
(NULL) buffer → read at address 0. The zero-sample input is reachable from real
use (a push-to-talk recording stopped before the first capture buffer yields a
valid-header, zero-frame WAV).

With the fields default-initialized to 0, `n_samples == 0` deterministically
takes the existing "input too short" path and returns 0 segments. The two code
paths that legitimately set a mel (`log_mel_spectrogram`,
`whisper_set_mel_with_state`) overwrite all three fields before any read, so the
defaults are observable only on the never-computed mel — the bug case.

## Test verification (RED → GREEN)

Added `tests/test-whisper-zero-samples.cpp`: load `for-tests-ggml-tiny.bin`, call
`whisper_full(ctx, params, nullptr, 0)`, assert `rc == 0` and `n_segments == 0`.

Because the read is of uninitialized memory, the deterministic RED is shown with
valgrind on the unpatched struct (only the production change reverted, the new
test applied):

**RED** (upstream base — `whisper_mel` without initializers):
```
==316== Conditional jump or move depends on uninitialised value(s)
==316==    at 0x4987C4C: whisper_full_with_state (whisper.cpp:6886)
==316==    by 0x498B768: whisper_full (whisper.cpp:7811)
==316==    by 0x109375: main (test-whisper-zero-samples.cpp:30)
==316==  Uninitialised value was created by a heap allocation
==316==    by 0x49785B7: whisper_init_state (whisper.cpp:3388)
...
ERROR SUMMARY: (valgrind exit code 1)
```

**GREEN** (with the fix):
```
test-whisper-zero-samples: OK
==75== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
valgrind exit: 0
```

Normal (non-valgrind) run of the new test plus the existing `whisper-cli` tiny
test both pass.

Fixes #3978

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -416,9 +416,9 @@ static const std::map<whisper_alignment_heads_preset, whisper_aheads> g_aheads {
 static std::vector<uint32_t> get_alignment_heads_by_layer(const whisper_context_params & cparams, int il, int32_t n_text_layer, int32_t n_head);
 
 struct whisper_mel {
-    int n_len;
-    int n_len_org;
-    int n_mel;
+    int n_len     = 0;
+    int n_len_org = 0;
+    int n_mel     = 0;
 
     std::vector<float> data;
 };
--- a/tests/test-whisper-zero-samples.cpp
+++ b/tests/test-whisper-zero-samples.cpp
@@ -1,0 +1,30 @@
+#include "whisper.h"
+
+#include <cstdio>
+
+#ifdef NDEBUG
+#undef NDEBUG
+#endif
+#include <cassert>
+
+int main() {
+    struct whisper_context_params cparams = whisper_context_default_params();
+    cparams.use_gpu = false;
+
+    struct whisper_context * ctx = whisper_init_from_file_with_params(WHISPER_MODEL_PATH, cparams);
+    assert(ctx != nullptr);
+
+    struct whisper_full_params params = whisper_full_default_params(WHISPER_SAMPLING_GREEDY);
+    params.no_timestamps = true;
+    params.print_progress = false;
+    params.print_realtime = false;
+
+    const int rc = whisper_full(ctx, params, nullptr, 0);
+    assert(rc == 0);
+    assert(whisper_full_n_segments(ctx) == 0);
+
+    whisper_free(ctx);
+
+    printf("test-whisper-zero-samples: OK\n");
+    return 0;
+}
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -103,6 +103,16 @@ target_link_libraries(${BUFFER_LOADER_TEST} PRIVATE common)
 add_test(NAME ${BUFFER_LOADER_TEST} COMMAND ${BUFFER_LOADER_TEST})
 set_tests_properties(${BUFFER_LOADER_TEST} PROPERTIES LABELS "unit;gh")
 
+# whisper_full() with n_samples == 0 must not read an uninitialized mel (#3978)
+set(ZERO_SAMPLES_TEST test-whisper-zero-samples)
+add_executable(${ZERO_SAMPLES_TEST} ${ZERO_SAMPLES_TEST}.cpp)
+target_include_directories(${ZERO_SAMPLES_TEST} PRIVATE ../include ../ggml/include ../examples)
+target_link_libraries(${ZERO_SAMPLES_TEST} PRIVATE common)
+target_compile_definitions(${ZERO_SAMPLES_TEST} PRIVATE
+    WHISPER_MODEL_PATH="${PROJECT_SOURCE_DIR}/models/for-tests-ggml-tiny.bin")
+add_test(NAME ${ZERO_SAMPLES_TEST} COMMAND ${ZERO_SAMPLES_TEST})
+set_tests_properties(${ZERO_SAMPLES_TEST} PROPERTIES LABELS "tiny;gh")
+
 # VAD test tests VAD in isolation
 set(VAD_TEST test-vad)
 add_executable(${VAD_TEST} ${VAD_TEST}.cpp)


src/whisper.cpp
```diff
@@ -412,9 +412,12 @@ static const std::map<whisper_alignment_heads_preset, whisper_aheads> g_aheads {
 static std::vector<uint32_t> get_alignment_heads_by_layer(const whisper_context_params & cparams, int il, int32_t n_text_layer, int32_t n_head);
 
 struct whisper_mel {
-    int n_len;
-    int n_len_org;
-    int n_mel;
+    // Default-initialized so a freshly allocated whisper_state whose mel was never
```
danbev: Please remove all these comments.

tests/test-whisper-zero-samples.cpp
```diff
@@ -0,0 +1,38 @@
+// Regression test for issue #3978:
+// whisper_full() called with n_samples == 0 on a fresh state must not read the
```
danbev: It is enough to keep the link to the PR on the line above and remove the additional comments.

ousamabenyounes:
Removed the verbose comments as requested. The zero-samples test still passes. 
thank you for your review

danbev:
@ousamabenyounes Thanks! Can you take a look at the conflict?

ousamabenyounes:
@danbev Rebased onto the latest `master` to clear the conflict — the PR is now mergeable again (`02c45f5b`). The change is unchanged from your review: `whisper_mel`'s three fields default-initialized to `0`, plus the zero-samples regression test.

Locally, `ctest -L gh` is green (4/4, including `test-whisper-zero-samples`) on the rebased base. The GitHub Actions checks are showing `action_required` (fork-PR workflows waiting on a maintainer approval-to-run) rather than a failure.

danbev (approved):

Issue #3879, opened 2026-06-14
Undefined behavior in whisper.cpp model loading: null pointer passed to memcpy + signed integer overflow


professor-moody:
## Summary

Two UBSan-detected undefined-behavior bugs occur in `whisper_model_load()` when parsing a crafted
GGUF model file. The first passes a null pointer to `memcpy` in the buffer loader's `read` callback;
the second is a signed integer overflow when computing the mel filter buffer size from unvalidated
header fields. Both are undefined behavior under the C/C++ standards; the practical impact (whether
either becomes a hard crash) is platform- and build-dependent and is not a confirmed SEGV.

## Affected component

- `src/whisper.cpp` — `whisper_model_load()`
  - Bug 1: null pointer passed to `memcpy` at `src/whisper.cpp:3688`
  - Bug 2: signed integer overflow at `src/whisper.cpp:1583`

## Reproduction

PoCs:
- `poc-051-whisper-null-ptr.bin` (68 bytes) — triggers the null-pointer-to-`memcpy` UB
- `poc-051-whisper-int-overflow.bin` (52 bytes) — triggers the signed integer overflow

Load either file through the buffer-based model loader (e.g.
`whisper_init_from_buffer_with_params`). On a UBSan-instrumented build:

```
src/whisper.cpp:3688:16: runtime error: null pointer passed as argument 1, which is declared to never be null
```

```
src/whisper.cpp:1583:43: runtime error: signed integer overflow: 1920274432 * -1094795586 cannot be represented in type 'int32_t'
```

### Bug 1 — null pointer to memcpy (`whisper.cpp:3688`)

The custom buffer loader's `read` callback passes a null-derived source pointer to `memcpy()` when
the buffer context is exhausted or its base pointer is null. Passing a null source to `memcpy` is
undefined behavior regardless of the copy size:

```cpp
loader.read = [](void * ctx, void * output, size_t read_size) {
    buf_context * buf = reinterpret_cast<buf_context *>(ctx);
    size_t size_to_copy = buf->current_offset + read_size < buf->size
        ? read_size : buf->size - buf->current_offset;
    memcpy(output, buf->buffer + buf->current_offset, size_to_copy);  // null source -> UB
    buf->current_offset += size_to_copy;
    return size_to_copy;
};
```

### Bug 2 — signed integer overflow (`whisper.cpp:1583`)

`filters.n_mel` and `filters.n_fft` are read directly from the model file with no validation, and
[...]

PR #3982, opened 2026-08-09
whisper : guard null source in buffer loader read callback

ousamabenyounes:
## What

The buffer-based model loader (`whisper_init_from_buffer_with_params`) installs a
`read` callback that copies from `buf->buffer + current_offset` with `memcpy`.
When the buffer is exhausted — or the supplied buffer is empty/NULL —
`size_to_copy` is 0 and the source pointer is NULL. Passing a NULL pointer to
`memcpy` is undefined behavior even for a zero-length copy.

Skip the copy when there is nothing to copy.

This addresses **bug 1** of #3879 (the null-pointer-to-`memcpy` UB). Bug 2 (the
`int32` overflow when sizing the mel-filter buffer) is a distinct issue already
covered by the open PR #3780, so it is intentionally left out here.

## Why

Loading a crafted/short model through the buffer loader (e.g. via
`whisper_init_from_buffer`) drives the callback with nothing left to copy.
UBSan flags the `memcpy` directly.

## Test verification (RED → GREEN)

Added `tests/test-whisper-buffer-loader.cpp`: loading from a NULL/empty buffer
and from a truncated non-model buffer must return `NULL` without UB. It is
labelled `gh` so CI's `ctest -L gh` runs it.

The undefined behavior is deterministic under UBSan (built with
`-fsanitize=address,undefined -fno-sanitize-recover=all`):

**RED** (upstream base — `memcpy` unguarded, new test applied):
```
/src/src/whisper.cpp:3698:15: runtime error: null pointer passed as argument 2, which is declared to never be null
    #0 ... in operator() /src/src/whisper.cpp:3698
test exit: 1
```

**GREEN** (with the fix):
```
test-whisper-buffer-loader: OK
test exit: 0
```

Normal-build run (regression guard): the new test plus the existing `whisper-cli`
tiny test and the UTF-8 unit test all pass.

Refs #3879 (bug 1). Bug 2 is tracked by #3780.

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -3766,10 +3766,7 @@ struct whisper_context * whisper_init_from_buffer_with_params_no_state(void * bu
 
         size_t size_to_copy = buf->current_offset + read_size < buf->size ? read_size : buf->size - buf->current_offset;
 
-        // When the buffer is exhausted size_to_copy is 0 and buf->buffer may be
-        // null (empty buffer); passing a null source to memcpy is undefined
-        // behavior even for a zero-length copy, so skip it (issue #3879).
-        if (size_to_copy > 0) {
+        if (size_to_copy > 0 && buf->buffer != nullptr) {
             memcpy(output, buf->buffer + buf->current_offset, size_to_copy);
         }
         buf->current_offset += size_to_copy;
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -3766,7 +3766,12 @@ struct whisper_context * whisper_init_from_buffer_with_params_no_state(void * bu
 
         size_t size_to_copy = buf->current_offset + read_size < buf->size ? read_size : buf->size - buf->current_offset;
 
-        memcpy(output, buf->buffer + buf->current_offset, size_to_copy);
+        // When the buffer is exhausted size_to_copy is 0 and buf->buffer may be
+        // null (empty buffer); passing a null source to memcpy is undefined
+        // behavior even for a zero-length copy, so skip it (issue #3879).
+        if (size_to_copy > 0) {
+            memcpy(output, buf->buffer + buf->current_offset, size_to_copy);
+        }
         buf->current_offset += size_to_copy;
 
         return size_to_copy;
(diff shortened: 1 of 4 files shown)


src/whisper.cpp
```diff
@@ -3695,7 +3695,12 @@ struct whisper_context * whisper_init_from_buffer_with_params_no_state(void * bu
 
         size_t size_to_copy = buf->current_offset + read_size < buf->size ? read_size : buf->size - buf->current_offset;
 
-        memcpy(output, buf->buffer + buf->current_offset, size_to_copy);
+        // When the buffer is exhausted size_to_copy is 0 and buf->buffer may be
+        // null (empty buffer); passing a null source to memcpy is undefined
+        // behavior even for a zero-length copy, so skip it (issue #3879).
+        if (size_to_copy > 0) {
```
danbev: This does not guard `buf->buffer` from being a nullptr which is what the comment and the PR description says. The check is included in the linked issues description.

We also have the same issue in [parakeet.cpp](https://github.com/ggml-org/whisper.cpp/blob/52dec9d889cc8e5f0a1978f2eb37a96978a131be/src/parakeet.cpp#L3112). We  have some duplication between whisper.cpp and parakeet.cpp and we might look into extracting some of the common code. But for this PR it would be enough to just add the same checks to parakeet.cpp (no need for another test though).

And also the comments are a bit verbose and don't really add much value, just adding the checks is fine.

tests/test-whisper-buffer-loader.cpp
```diff
@@ -0,0 +1,33 @@
...
+#undef NDEBUG
+#endif
+#include <cassert>
+
+int main() {
+    struct whisper_context_params cparams = whisper_context_default_params();
+    cparams.use_gpu = false;
+
+    // Empty buffer: the read callback is driven with a NULL source pointer.
+    struct whisper_context * ctx_empty = whisper_init_from_buffer_with_params(nullptr, 0, cparams);
```
danbev:
```suggestion
    struct whisper_context * ctx_empty = whisper_init_from_buffer_with_params(nullptr, 1, cparams);
```
The above will trigger a segfault with the current implementation which might be more useful to capture the buffer being nullptr.

ousamabenyounes:
Thanks for your review :  
Added the null-buffer check to both Whisper and Parakeet, updated the regression test to cover a null buffer with a non-zero size, and removed the verbose comments. 
UBSan reproduces the issue before the fix and passes after it.

danbev (approved):

PR #3997, opened 2026-08-19
whisper : bypass cross-attention scaling for OpenVINO backend

mostafafaheem:
This PR adds a preprocessor check for `GGML_USE_OPENVINO` to bypass attention scaling in the cross-attention subgraph, since this scaling is handled internally by the OpenVINO backend.

### DIFF
--- a/src/whisper.cpp
+++ b/src/whisper.cpp
@@ -2310,7 +2310,7 @@ static struct ggml_cgraph * whisper_build_graph_cross(
                 layer.cross_attn_k_w,
                 cur);
 
-#if !defined(GGML_USE_OPENVINO) // scaling is handled internally in OpenVINO backend
+#ifndef GGML_USE_OPENVINO // scaling is handled internally in OpenVINO backend
         Kcross = ggml_scale(ctx0, Kcross, Kscale);
 #endif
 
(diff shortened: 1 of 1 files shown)


mostafafaheem:
@ggerganov @danbev Can you please have a look? I tried doing this inside the backend with a graph transformation but it was hitting false positives, no matter how specific the matching subgraph was. If this PR is not acceptable, I will try to find another approach.

src/whisper.cpp
```diff
@@ -2310,7 +2310,9 @@ static struct ggml_cgraph * whisper_build_graph_cross(
                 layer.cross_attn_k_w,
                 cur);
 
+#if !defined(GGML_USE_OPENVINO) // scaling is handled internally in OpenVINO backend
```
danbev: For consistency I thing this should perhaps be:
```suggestion
#ifndef WHISPER_USE_OPENVINO // scaling is handled internally in OpenVINO backend
```

mostafafaheem: I changed to use `ifndef` instead of `if !defined` but the variable should stay the same, since this is related to the GGML OpenVINO backend, not the standalone encoder in whisper.

danbev: Ah yes, I was wrong about that and thanks for clarifying!

danbev (approved):

PR #4019, opened 2026-08-27
docs(stream): clarify VAD mode timestamps and CWD model path errors

apollo-2006:
Added documentation for undocumented timestamp behavior in VAD mode and clarified CWD errors to prevent parser breaks," and hit submit

AI use: I found this while building against the library and used an AI assistant to help verify the relevant source (file/line refs above). The report and the documentation wording are mine, and I've checked every claim against the source myself.

### DIFF
--- a/examples/stream/README.md
+++ b/examples/stream/README.md
@@ -27,7 +27,7 @@ a transcription block that is suitable for parsing.
 
 ## Output format
 
-The shape depends on --step, and there is no flag to override it.
+The output format depends on --step, and there is no flag to override it.
 
 Default (--step > 0): one rolling segment, no timestamps, rewritten in
 place with ANSI erase-line escapes. Fine for a terminal, wrong for a pipe.
--- a/examples/stream/README.md
+++ b/examples/stream/README.md
@@ -25,6 +25,36 @@ It's best to tune it to the specific use case, but a value around `0.6` should b
 When silence is detected, it will transcribe the last `--length` milliseconds of audio and output
 a transcription block that is suitable for parsing.
 
+## Output format
+
+The shape depends on --step, and there is no flag to override it.
+
+Default (--step > 0): one rolling segment, no timestamps, rewritten in
+place with ANSI erase-line escapes. Fine for a terminal, wrong for a pipe.
+
+VAD mode (--step 0): timestamped blocks meant for parsing.
+
+    ### Transcription 0 START | t0 = 0 ms | t1 = 4000 ms
+
+    [00:00:00.000 --> 00:00:03.480]   Hello there.
+
+    ### Transcription 0 END
+
+- timestamps are enabled implicitly by --step 0; there is no -nt here
+- strip the [t0 --> t1] prefix; don't discard those lines as log noise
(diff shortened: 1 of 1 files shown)


apollo-2006:
@danbev - you wrote #3065, so you're the right person to ask about the part I couldn't settle while writing this up.

When VAD is on, the timestamps that come back are relative to the filtered audio and get mapped back onto the original timeline. Reading the review thread on #3065, the mapping scales speech back across the original segment length rather than preserving where the silence actually sat. For a caller doing word-level alignment, is that difference meant to be treated as noise, or is preserving true silence positions something you'd want a follow-up for?

I hit this building a local voice assistant with a "Delphi" wake word on top of whisper-stream, so my failure mode was practical rather than theoretical: I was discarding timestamped lines as log noise until I realised the prefix was the data.

I'm a third-year CS student and I'm using this project mostly as an excuse to get better at reading code I didn't write. Thanks for #3065 either way.

danbev:
@apollo-2006 The stream example uses [vad_simple](https://github.com/ggml-org/whisper.cpp/blob/978113305b2ead22249b881deafa131dc8884911/examples/stream/stream.cpp#L304) and not the VAD implementation from #3065.

For your question, please take a look at https://github.com/ggml-org/whisper.cpp/pull/3910 which may be relevant.

apollo-2006:
Thanks, that's the part I had wrong. I was looking in the wrong place.

Checked `stream.cpp`: `use_vad` is just `n_samples_step <= 0`, and the only VAD is `vad_simple` in `examples/common.cpp`, which is an energy gate returning a bool. It never cuts silence or emits segments, and `wparams.vad` is never set, so nothing from #3065 runs here. My question didn't apply.

#3910 does answer what I was actually asking. Snapping a token that lands in removed silence to the nearer boundary, instead of interpolating across the gap, is the behaviour I meant. Merged, so nothing needed there.

One thing I did find: in `--step 0`, the `### Transcription N` t0/t1 are wall-clock since stream start, while the `[hh:mm:ss -->]` prefixes are relative to the chunk passed to `whisper_full`. Two clocks, and a parser that mixes them gets it wrong. Want me to add that to this PR? Seems more useful than what's in there now.

apollo-2006:
@danbev this is out of draft and ready for review. Force-pushing does not notify, so flagging it here.

One thing I noticed rereading the sample output: the block prints `### Transcription 0 START | t0 = 0 ms | t1 = 4000 ms` two lines above `[00:00:00.000 --> 00:00:03.480]`. Those are two different clocks sitting next to each other, and nothing on the page says which is which.

Worth a sentence distinguishing them here, or is that out of scope for a stream README?

examples/stream/README.md
```diff
@@ -25,6 +25,36 @@ It's best to tune it to the specific use case, but a value around `0.6` should b
 When silence is detected, it will transcribe the last `--length` milliseconds of audio and output
 a transcription block that is suitable for parsing.
 
+## Output format
+
+The shape depends on --step, and there is no flag to override it.
```
danbev:
```suggestion
The output format depends on --step, and there is no flag to override it.
```

danbev (approved):

PR #4031, opened 2026-09-03
tests : load backends before init when built with GGML_BACKEND_DL

### LINKED ISSUE #4030: tests : ctest -L gh fails on a GGML_BACKEND_DL=ON build
## What happens

`ctest -L gh` fails on any build configured with `GGML_BACKEND_DL=ON`. Two of
the four `gh`-labelled tests abort:

```
2/4 Test #11: test-whisper-buffer-loader .......   Passed    0.00 sec
3/4 Test #12: test-whisper-zero-samples ........Subprocess aborted***Exception
4/4 Test #15: test-parakeet ....................Subprocess aborted***Exception

50% tests passed, 2 tests failed out of 4
```

Both abort the same way:

```
whisper_init_with_params_no_state: devices    = 0
whisper_init_with_params_no_state: backends   = 0
ggml/src/ggml-backend.cpp:595: GGML_ASSERT(device) failed
```

The same test binaries pass when the build does not set `GGML_BACKEND_DL`.

## Why

With `GGML_BACKEND_DL=ON` the backends are loadable modules, and nothing is
registered until something calls `ggml_backend_load_all()`. The examples do
this — `examples/cli/cli.cpp`, `examples/bench/bench.cpp`,
`examples/quantize/quantize.cpp`, `examples/lsp/lsp.cpp`,
`examples/vad-speech-segments/speech.cpp`. The test programs never do, so
`devices` stays 0, `ggml_backend_dev_backend_reg()` gets a null device, and the
assert fires.

That also explains the split in the results:

| test | outcome | why |
| --- | --- | --- |
| `test-whisper-cli-tiny` | passes | drives `whisper-cli`, which calls `ggml_backend_load_all()` |
| `test-whisper-buffer-loader` | passes | never needs a device |
| `test-whisper-zero-samples` | aborts | no backend registered |
| `test-parakeet` | aborts | no backend registered |

## Why CI does not catch it

No workflow currently pairs the two. The workflows that run `ctest -L gh` —
`build-clang.yml`, `build-gcc.yml`, `build-sanitize.yml` — never set
`GGML_BACKEND_DL`. The workflow that does set it, `release.yml`, runs no
`ctest` at all.

## Reproducing

```bash
cmake -B build -DCMAKE_BUILD_TYPE=Release -DGGML_BACKEND_DL=ON -DGGML_NATIVE=OFF
cmake --build build --config Release -j
ctest --test-dir build -L gh --output-on-failure
```

Seen on macOS 26.6.2 / arm64 with Apple clang 21.0.0, on both a
`GGML_CPU_ALL_VARIANTS=ON` and a plain `GGML_BACKEND_DL=ON` build. Nothing in
it looks platform-specific.

## Fix

Adding `ggml_backend_load_all()` to the affected tests, the way the examples
do, should be all it takes. Happy to send that as a PR if the approach looks
right — or if the preference is for the tests to stay backend-agnostic and for
CI to cover the combination instead, that works too.

Noticed while adding the macOS release jobs in #4029, which builds with
`GGML_BACKEND_DL=ON`; that PR deliberately adds no `ctest` step because of this.

apollo-2006:
Fixes #4030.

`ggml_backend_load_all()` at the top of `main()` in the four tests that
initialise a context. Every example already does this:
`examples/cli/cli.cpp`, `bench`, `quantize`, `lsp`,
`vad-speech-segments`. The tests were the only callers that did not.

With `GGML_BACKEND_DL=ON` no backend is registered until something calls it,
so `whisper_init_*` and `parakeet_init_*` ran with `devices = 0` and aborted
on `GGML_ASSERT(device)` in `ggml_backend_dev_backend_reg`.

No new include: `ggml_backend_load_all()` is already reachable through
`whisper.h` and `parakeet.h`, both of which pull in `ggml-cpu.h`, which
includes `ggml-backend.h`.

### Before / after

Full suite on a `-DGGML_BACKEND_DL=ON -DGGML_CPU_ALL_VARIANTS=ON` build,
macOS 26.6.2 / arm64, Apple clang 21.0.0:

| test | before | after |
| --- | --- | --- |
| `test-whisper-zero-samples` | aborted | passes |
| `test-vad` | aborted | passes |
| `test-parakeet` | aborted | passes |
| `test-vad-full` | aborted | reaches init, see below |

`ctest -L gh` goes from 2 of 4 to 4 of 4. The full suite goes from three
aborts to none.

`test-vad-full` is labelled `base;en` rather than `gh` because it needs a real
`models/ggml-base.en.bin`, which I do not have locally. It had the same fault
and the fix is verified the same way: before, it printed `devices = 0` and hit
`GGML_ASSERT(device)`; after, it prints `devices = 3` and gets through
initialisation. Substituting the stub model it then fails its own
`assert(n_segments == 1)`, identically on a `GGML_BACKEND_DL` and a normal
build, which is the stub having no segments rather than anything to do with
this change.

### Regression check

The same suite on a build without `GGML_BACKEND_DL` is unchanged; 14 of 15
before and after, the one failure being `test-vad-full` for the missing model
described above. Both configurations now produce identical results.

### Why CI did not catch it

None of the workflows pairs the two. `build-clang.yml`, `build-gcc.yml` and
`build-sanitize.yml` run `ctest -L gh` but never set `GGML_BACKEND_DL`;
`release.yml` sets it but runs no tests. Adding a `ctest` step to a
`GGML_BACKEND_DL` job would cover the combination, but that felt like a
separate change from fixing the tests, so I have left it out here.

### DIFF
--- a/tests/test-parakeet.cpp
+++ b/tests/test-parakeet.cpp
@@ -111,6 +111,8 @@ static int test_valid_model() {
 }
 
 int main(){
+    ggml_backend_load_all();
+
     if(test_valid_model() != 0){
         return 1;
     }
--- a/tests/test-vad-full.cpp
+++ b/tests/test-vad-full.cpp
@@ -13,6 +13,8 @@
 #include <cassert>
 
 int main() {
+    ggml_backend_load_all();
+
     std::string whisper_model_path = WHISPER_MODEL_PATH;
     std::string vad_model_path     = VAD_MODEL_PATH;
     std::string sample_path        = SAMPLE_PATH;
--- a/tests/test-vad.cpp
+++ b/tests/test-vad.cpp
@@ -48,6 +48,8 @@ struct whisper_vad_segments * test_detect_timestamps(
 }
 
 int main() {
+    ggml_backend_load_all();
+
     std::string vad_model_path = VAD_MODEL_PATH;
     std::string sample_path    = SAMPLE_PATH;
 
--- a/tests/test-whisper-zero-samples.cpp
+++ b/tests/test-whisper-zero-samples.cpp
@@ -8,6 +8,8 @@
 #include <cassert>
 
 int main() {
+    ggml_backend_load_all();
+
     struct whisper_context_params cparams = whisper_context_default_params();
     cparams.use_gpu = false;
 


apollo-2006:
@danbev one thing I could not settle from the source. `test-vad` is labelled `unit` and `test-vad-full` is `base;en`, so neither runs under `ctest -L gh`, while `test-whisper-buffer-loader` beside it is `unit;gh`. That puts two of the four tests this PR fixes outside the label CI actually runs, so a regression here would go unnoticed the same way this did.

Is that split deliberate, keeping anything that needs a model out of the gh job, or should `test-vad` pick up `gh` the way the buffer-loader test does?

danbev (approved):

danbev:
> Is that split deliberate, keeping anything that needs a model out of the gh job, or should `test-vad` pick up `gh` the way the buffer-loader test does?

There is a separate workflow, [build-vad.yml](https://github.com/ggml-org/whisper.cpp/blob/52a939a2a762224e255d366c1182b2af4dd1a032/.github/workflows/build-vad.yml#L50) which runs test-vad so it does not require the `gh` label. 
And test-vad-full is left as a manual test as it requires a larger whisper model.

PR #4032, opened 2026-09-03
ruby : Add #free method, check MemoryView strictly

KitaitiMakoto:
Hello,

This pull request

* adds `#free` method to unload large model manually,
* makes the check of MemoryView more strictly, and
* fixed a class for test

Thank you.

### DIFF
--- a/bindings/ruby/README.md
+++ b/bindings/ruby/README.md
@@ -202,7 +202,7 @@ whisper.transcribe("path/to/audio.wav", params, n_processors: Etc.nprocessors)
 
 Note that transcription occasionally might be low accuracy when it works in parallel.
 
-If n_processors is greater than 1, you cannot set any callbacks including new_segment_callback, progress_callback, encoder_begin_callback, abort_callback, and log_callback set by Whisper.log_set.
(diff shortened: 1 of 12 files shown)


bindings/ruby/README.md
```diff
@@ -202,7 +202,7 @@ whisper.transcribe("path/to/audio.wav", params, n_processors: Etc.nprocessors)
 
 Note that transcription occasionally might be low accuracy when it works in parallel.
 
-If n_processors is greater than 1, you cannot set any callbacks including new_segment_callback, progress_callback, encoder_begin_callback, abort_callback, and log_callback set by Whisper.log_set.
+If n_processors is greater than 1, you cannot set some callbacks including  encoder_begin_callback, abort_callback. Also, progress_callback is ignored even if it's set.
```
danbev: Nit: There is an additional white space here after `including`:
```suggestion
If n_processors is greater than 1, you cannot set some callbacks including encoder_begin_callback, abort_callback. Also, progress_callback is ignored even if it's set.
```

danbev (approved):

bindings/ruby/README.md
KitaitiMakoto: Thanks!

PR #4038, opened 2026-09-06
devops : reduce Vulkan Docker image to 39.4% of its original size (now 692 MB)

turicas:
The current Vulkan Docker image is 1.75 GB. Making it smaller not only reduces download and storage requirements, but can also make the image faster to build and distribute. This PR applies 4 changes to `main-vulkan.Dockerfile`, reducing the final image size from 1.75 GB to 692 MB (39.4% of the original size). The build time also decreased from 2m41s to 2m16s (84.4% of the original).

These were the main contributors to the current image size:
- Installing recommended APT packages
- Installing development packages in the runtime/final stage
- Copying the entire repository and build artifacts into the runtime/final stage (including the `base.en` model)

I also made a couple of improvements:
- Use `apt` instead of `apt-get`
- Remove `bash -c` as the entrypoint

I'm willing to apply the same changes to the other Dockerfiles in the repository if the maintainers approve this approach. I started with the Vulkan image because it's the one I use.

Result:

```text
$ docker image ls whispercpp-test
REPOSITORY        TAG                                   IMAGE ID       CREATED          SIZE
whispercpp-test   04-remove-entrypoint                  70eb4215ee40   11 minutes ago   692MB
whispercpp-test   03-copy-only-needed-files             57eeec07fb51   19 minutes ago   692MB
whispercpp-test   02-remove-dev-packages-from-runtime   35cde18f6236   22 minutes ago   1.37GB
whispercpp-test   01-do-not-install-recommends          5677b0c472aa   28 minutes ago   1.6GB
whispercpp-test   00-original                           2b93628d52a7   31 minutes ago   1.75GB
```

Notes:

1. Regarding `libsdl2-dev` (removed in the second commit): commit bea43e0c (not mine) introduced this package in the runtime image, but since it's not required for production use, I think it should instead be installed by users who need it in an image derived from this one.
2. I've run all the `docker run` commands from the README using the final Dockerfile in this PR and everything worked as expected (the examples were also fixed/extended).

### DIFF
--- a/.devops/main-vulkan.Dockerfile
+++ b/.devops/main-vulkan.Dockerfile
@@ -1,8 +1,8 @@
 FROM ubuntu:24.04 AS build
 WORKDIR /app
 
-RUN apt-get update && \
-  apt-get install -y build-essential wget cmake git libvulkan-dev spirv-headers glslc \
+RUN apt update && \
+  apt install --no-install-recommends -y build-essential ca-certificates cmake git glslc libvulkan-dev spirv-headers wget \
   && rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
 
 COPY .. .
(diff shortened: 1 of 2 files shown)


danbev (approved):

turicas:
@danbev thanks for the review! Should I create one more PR for the other Dockerfiles?

danbev:
> Should I create one more PR for the other Dockerfiles?

I think it would be best to do them as separate PRs. Thanks!

PR #4040, opened 2026-09-07
Add support for publishing Windows ARM64 binaries

MugundanMCW:
Description:

* The PR adds support for building and publishing Windows ARM64 binaries in releases.
* Added Windows ARM64 job in existing CI workflow. 
* Currently SDL2 binaries are only provided for Windows 32-bit and 64-bit system. So, I have excluded SDL2 option from build configuration.
* The generated binaries are working natively on Windows ARM64.


danbev:
Sorry for possibly duplicating some of your work and effort here :disappointed:  
Could you take a look at https://github.com/ggml-org/whisper.cpp/pull/4048 and see if this covers your use case?

MugundanMCW:
> Sorry for possibly duplicating some of your work and effort here 😞 Could you take a look at #4048 and see if this covers your use case?

sure

PR #4064, opened 2026-09-13
vad : reject n_encoder_layers other than 4 in model load

apollo-2006:
A VAD model whose header declares `n_encoder_layers` other than 4 causes an out-of-bounds heap read in `whisper_vad_init_with_params`, reachable from the ordinary `whisper-cli --vad -vm <file>` path on any user-supplied VAD model.

`n_encoder_layers` is read from the file and used, unbounded, to size three heap arrays, while the encoder graph below is hardcoded to four layers and indexes those arrays at `[0..3]` regardless of what the file declared. A value below 4 reads past the allocation.

The fix rejects it at the point of the read, matching the `n_dims` guard in #3957. The graph only implements four layers, so any other value is out of range rather than something to accommodate. For a well-formed model this is a no-op.

Tested with a crafted 3-layer model: SIGABRT before, rejected with exit 10 after, and heap-buffer-overflow under ASan before, clean after. The real silero v6.2.0 model transcribes unchanged. `test-vad` and `test-vad-full` pass.


src/whisper.cpp
```diff
@@ -4920,6 +4920,14 @@ struct whisper_vad_context * whisper_vad_init_with_params(
     {
         read_safe(loader, hparams.n_encoder_layers);
 
+        // The encoder graph below is built for exactly 4 layers and indexes the
```
danbev: Remove the comments as the check and the error message are clear enough. 

And please avoid using AI for generating your pull requests. They are verbose and it takes a lot of time to read them especially in cases like this when the change i trivial.

apollo-2006:
comments are gone, and ive trimmed the description. thank you for the review. sorry for the inconvenience

danbev (approved):

PR #4069, opened 2026-09-16
cli : load backends after validating input files

erikdw:
I noticed that expensive operations (e.g., compiling shaders) are done prior to checking if the input file even exists.

Solution:
* Move backend loading after the input-file checks so we can fail early.
* Also fail early if the model path is missing.

> [!Note]
> AI assistant used to locate the call and apply the move. I can discuss the change.

Made with [Cursor](https://cursor.com)


examples/cli/cli.cpp
```diff
@@ -1048,6 +1046,14 @@ int main(int argc, char ** argv) {
         whisper_log_set(cb_log_disable, NULL);
     }
 
+    if (!is_file_exist(params.model.c_str())) {
```
danbev: Optional, but perhaps move this up to before or after the `params.fname_inp` check.

danbev (approved):

examples/cli/cli.cpp
erikdw: Thanks I like that suggestion!

danbev (approved):

danbev:
The failing android CI jobs were fixed as part of https://github.com/ggml-org/whisper.cpp/commit/4afa009405863a41b4e86e0bca141ac62f9ea194.

PR #4077, opened 2026-09-20
whisper : run abort_callback on the language auto-detect path

### LINKED ISSUE #3888: Language auto-detect in `whisper_full` ignores `encoder_begin_callback` and `abort_callback`

renvins:
When `whisper_full / whisper_full_with_state` runs the language autodetection, the encode/decode paths ignore the `abort_callback` passed as a parameter. There was the `encoder_begin_callback` that was fixed in #3936.

What I changed:
- Created internal helper `whisper_lang_auto_detect_internal` to not change the public API
- `whisper_full_with_state` now passes `params.abort_callback` down to the new helper I built.
- Added batch prep and kv seq_rm lines in the decoding part because they are present in the public decoding wrapper, that I stopped calling because it can't take the callback.
- Added a test to assert the abort during lang auto detection. 

Fixes #3888


renvins:
@danbev tagging for visibility and for your review :)

danbev (approved):

renvins:
Good, all checks pass. I think we can merge.

